PR #2, opened 2021-08-17
Updated package-lock.json & npm package release

rryter:
As per discussion with @CJ42


rryter:
I think npm 7 is causing problems here.

PR #3, opened 2021-08-17
Use Typechain to provide Typescript types

rryter:
🧙 With Typechain we can automatically create Typescript types
- Web3
- Ethers


Hugoo:
@rryter I see `generate-types` is called in `postinstall`.

Is there an "automatic" way to also re generate-types on build or release? Or is it a manual process?

rryter:
> @rryter I see `generate-types` is called in `postinstall`.
> 
> Is there an "automatic" way to also re generate-types on build or release? Or is it a manual process?

Yes, with `hardhat` we could simply use the `@typechain/hardhat` plugin.

PR #8, opened 2021-08-20
Merge ci-cd with permissions

CJ42:
Downgrade truffle to 5.3.0 (Fix London upgrade bug).
Add workflow for `.jar` android artifacts + publish to Maven.


rryter (approved):
Look good to me.

The skipped test will be fixed in the next PR.

PR #9, opened 2021-08-24
CI failing because of the wrong Swift file name

JeneaVranceanu:
File name should be changed in `make-ios.js` because in the workflow it is correct.


JeneaVranceanu:
@CJ42 Hi, please take a look tomorrow.

PR #19, opened 2021-09-22
WIP Lsp8

mattgstevens:
## Description
Includes an implementation of LSP8, with extensions to support compatability with existing clients and tools as well as a capped supply version.

## Details
From the agreed on spec, added `function getOperatorsOf(bytes32 tokenId)` to return the list of operator addresses for a given tokenId. 

## Tasks
- [ ] depends on implementation for ERC725Y allowing memory values (one solution in this PR https://github.com/ERC725Alliance/ERC725/pull/55)
- [ ] matches agreed on specification https://gist.github.com/frozeman/f601ffae185a19b78d459c20989b9440


contracts/LSP8/extensions/ILSP8CappedSupply.sol
```diff
@@ -0,0 +1,21 @@
...
+interface ILSP8CappedSupply is ILSP8 {
+  /**
+   * @dev Returns the number of tokens that have been minted.
+   */
+  function tokenSupplyCap() external view returns (uint256);
+
+  /**
+   * @dev Returns the number of tokens available to be minted.
+   */
+  function mintableSupply() external view returns (uint256);
```
frozeman, line 20: I think its unnecessary if you can do `totalsupplyCap() - totalSupply()`

PR #18, opened 2021-09-22
Continue on Develop before creating rc-* branch

CJ42:
Merge latest change (see below) onto develop

- [x] `KeyManager` with MultiChannel nonce + tests
- [x] Update README with installations + usage example
- [x] Updates related to `ERC725Y` breaking changes (setData array) 
    - created LSP2 Utils library (for `getDataSingle(…)`)
    - applied to contracts implementation for LSP3 and LSP4.
    - `KeyManager` loop through keys to check permissions
- [x] Package-lock.json version 2 (NPM 7)
- [x] Clean Solidity code + comments

Next to do:

- [ ]  split **KeyManager** ✅ , LSP3Account and UniversalReceiver contract between Core and normal / Init
- [x] change of parameters in signatures (_data) as 3rd param + refactor unit tests with new order
- [x] changed order of signature (flipped abi <> nonce) in `executeRelayCall`

PR #20, opened 2021-09-23
fix: import path for KeyManager

mattgstevens:
## Description
The current `develop` branch fails when building the contracts using `npm run build`.

## Tasks
- [x] fix import path


mattgstevens:
this should be fixed in
https://github.com/lukso-network/lsp-smart-contracts/pull/18

PR #50, opened 2021-11-09
Include MythX in CI

CJ42:

Adding MythX audit to analyse security vulnerabilities in smart contracts 🔐  🛠️
MythX audit is now part of the CI, and will run on every PR to the branches "main" and "develop"

**MythX Settings**

- run in `deep` mode: explore a larger space (x90) of program states and execution paths 👀  + faster computing time 🚀 
- use solc version **0.8.7**
- Run for the following contracts:
    - `UniversalProfile.sol`, `Keymanager.sol`, `UniversalReceiverDelegate.sol`, `LSP7.sol`, `LSP8.sol` x 5
    - their proxy version x 5
    - `Create2Factory.sol`
  
- Run asynchronously with the `async` flag.

More informations about the deep mode can be found in this [MythX blog post: _MythX Pro Security Analysis Explained_](https://blog.mythx.io/features/mythx-full-mode-security-analysis-explained/)


CJ42:
@frozeman @Hugoo 

**Not sure about the following. Would be good to test with async: `false`**
For the `async` mode, I wonder if we should keep it to:

- `true`: this will make the CI to always pass, and just submit the contracts to MythX. The analysis reports can then be examined afterwards.

- `false`: this will make the CI to keep running, until all the analysis are complete. It might take a while for each CI to run, but the CI would fail if High Vulnerabilities are detected.

Hugoo:
> **Not sure about the following. Would be good to test with async: `false`** For the `async` mode, I wonder if we should keep it to:
> 
> * `true`: this will make the CI to always pass, and just submit the contracts to MythX. The analysis reports can then be examined afterwards.
> * `false`: this will make the CI to keep running, until all the analysis are complete. It might take a while for each CI to run, but the CI would fail if High Vulnerabilities are detected.

I'm in favour of the async false so it can really block the PR from being merged in case of issues.

If not, can we add a bot or something that add the security audit report in the PR discussion? (Similar to Codecov?)

@CJ42

CJ42:
@CallumGrindle what is your opinion on this?

CJ42:
> > **Not sure about the following. Would be good to test with async: `false`** For the `async` mode, I wonder if we should keep it to:
> > 
> > * `true`: this will make the CI to always pass, and just submit the contracts to MythX. The analysis reports can then be examined afterwards.
> > * `false`: this will make the CI to keep running, until all the analysis are complete. It might take a while for each CI to run, but the CI would fail if High Vulnerabilities are detected.
> 
> I'm in favour of the async false so it can really block the PR from being merged in case of issues.
> 
> If not, can we add a bot or something that add the security audit report in the PR discussion? (Similar to Codecov?)
> 
> @CJ42

I am in favour of a bot. If that could add that automatically as a comment to the PR, it would be really great.
As mentioned by @CallumGrindle , we could even write a script that call the MythX API and push a comment in the PR.

PR #51, opened 2021-11-10
build: Fix path in `make-ios.js`

YamenMerhi:
- [x] build: Fix path in `make-ios.js`


YamenMerhi:
**Continue to the change in [PR#48](https://github.com/lukso-network/lsp-universalprofile-smart-contracts/pull/48) .**

PR #59, opened 2021-11-16
Fix solhint CI issue + use Node.js v16 in workflows

CJ42:

- [x] Fix failing CI for solhint by using Nodejs version 16 in workflow.

- **moved solhint inside a `devDependency`** + run as a npm script, instead of installing solhint globally in workflow
= **save 2min 30s** in workflow ⏲️ 

- [x] updated release workflows to use Node.js version 16 to avoid same issue when running `npm ci`

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,18 +13,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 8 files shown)


frozeman (approved):

frozeman:
@CJ42 ready?

PR #63, opened 2021-11-19
Organise constants files in unit tests

CJ42:

- include all ERC725Y keys related to LSPs inside `constants.ts` + make all references from test to this file.
- deleted util file `keymanager.ts` + moved its content inside `constants.ts`
- improve layout of some solidity contracts.


Hugoo (changes requested):
Why are there so many old commits in this PR? There are 29 commits from 2months ago.
Is it possible for you to rebase on main/develop and just put the relevant commits in this PR?

It makes it hard to read as I don't know what is new and what is not.
I also spotted lot of format related changes. But they might come from old commits.

Let's make sure the whole project is formatted and the format check is done in the CI so we don't have ever again a format change in a PR.

I also saw a few commits related to submodules but I think this is old as the submodules have been removed a couples of days ago?

CJ42:
@Hugoo 

> Why are there so many old commits in this PR? There are 29 commits from 2months ago.
> Is it possible for you to rebase on main/develop and just put the relevant commits in this PR?

it has been merged already.
I opened this branch long time ago, and left it on the side. So was just looking to finish the job and close the branch.

> I also spotted lot of format related changes. But they might come from old commits.
> Let's make sure the whole project is formatted and the format check is done in the CI so we don't have ever again a format change in a PR.

We should add a prettier config file. We always have file changes between me and @YamenMerhi, as we have different settings for Prettier on our own VS Code

> I also saw a few commits related to submodules but I think this is old as the submodules have been removed a couples of days ago?

Yes correct, we do not have submodules anymore.

Hugoo:
@CJ42 Ok cool thanks.

> We should add a prettier config file. We always have file changes between me and @YamenMerhi, as we have different settings for Prettier on our own VS Code

Yes you should add a `.prettierrc` in the root so formatting stays consistent between all devs. Format changes between devs should not happen.

PR #64, opened 2021-11-23
ci: add prettier config for repo + edit solhint rules

CJ42:

- [x] added prettier rules for Solidity code with devDependency `prettier-config-solidity`
- [x] added prettier rules for JS + TS files (scripts + tests) in `.prettierrc`
- [x] cleaned solhint configurations by removing rules mentioned that are the same as default ones.
- [x] Flag `cal..value` + low level calls in Solidity with warnings (previously `off`)

Done with @YamenMerhi


.prettierrc
```diff
@@ -1,3 +1,12 @@
...
\ No newline at end of file
+  "overrides": [
+    {
+      "files": ["*.js", "*.,ts"],
+      "options": {
+        "tabWidth": 4,
+        "useTabs": true,
+        "printWidth": 100
+      }
+    }
```
Hugoo, line 10: cool !

This won't apply to solidity files then?

I also like to set these settings:

```
"trailingComma": "all",
"singleQuote": true
```

Not sure what is default for `singleQuote`

Hugoo: I see you have `prettier-plugin-solidity` so I assume you want this to apply to `.sol` files too?

Hugoo (commented):
Will it also perform the prettier/format check in the CI ?

Let's make sure it checks for prettier/format before code is merged

frozeman (approved):

.prettierrc
CJ42: @Hugoo 

I have added trailing commas and single quotes.
This will not apply to Solidity file no, but to JS / Typescript files.

Default configs `prettier-plugin-solidity` will be applied to `.sol` files.

CJ42:
> Will it also perform the prettier/format check in the CI ?
> 
> Let's make sure it checks for prettier/format before code is merged

I will test first and add this into a new PR.

PR #77, opened 2021-12-03
perf: use internal function call on deployment

CJ42:

Save gas on deployment with internal function calls.

Internal function calls are efficient in Solidity as they translate in JUMP instructions within the contract bytecode.

Use `_setData(...)` within constructor / initializer for `UniversalProfile` and `LSP9Vault` = saves 1,425 gas ⬇️


frozeman:
will it build in develop?

CJ42:
@frozeman let me merge develop into it first, so we can merge

PR #82, opened 2021-12-10
refactor: use Solidity built-in functions for ERC165 interface IDs

CJ42:

**Contracts**

- use syntax `type(ILSP{N}{StandardName}).interfaceId` to get ERC165 interface identifiers, instead of hardcoded constants.

This ensures that we rely on the core Solidity feature, instead of manually calculated values, that could lead to errors ✅ 

**Tests**

- replace occurence of interfaceIDs literals in test files with reference to `constants.ts` file.

### DIFF
--- a/contracts/LSP1UniversalReceiver/LSP1Constants.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1Constants.sol
@@ -1,9 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
+//
+import "@erc725/smart-contracts/contracts/interfaces/ILSP1_UniversalReceiver.sol";
+import "@erc725/smart-contracts/contracts/interfaces/ILSP1_UniversalReceiverDelegate.sol";
+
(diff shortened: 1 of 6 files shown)


frozeman:
And how do we read the interface IDs? Meaning how can people find them, and know the are the same as in the LSP specs?

CJ42:
@frozeman the interface IDs are currently defined in the contracts on _docs.lukso.tech:_ https://docs.lukso.tech/contracts/interface-ids

All the interface IDs are also mentioned in the file `constants.ts` (currently inside the `tests/` folder.
https://github.com/lukso-network/lsp-universalprofile-smart-contracts/blob/develop/tests/utils/constants.ts

I would suggest that we move this file out of this folder, and include it in the npm package. This way, clients building application in Javascript can get the interface IDs from this file (alongside the contract ABIs).

My motivation to suggest this in this PR is:
- it would be easier to maintain in the future (the more places and files we define interface IDs in, the harder it might be to maintain and update them all if these interface IDs change).
- we ensure that we use Solidity to define them, and get less prone to errors.

PR #89, opened 2021-12-17
feat: LSP8CompatibilityForERC721 sets ERC165 interfaces for ERC721 & …

mattgstevens:
…ERC721Metadata and emits ERC721 events for Transfer and Approval

## Desciption
When testing with some existing tooling for ERC721 the current LSP8 compatibility contract was not allowing indexers to behave as expected. Importantly the ERC165 interface id for ERC721Metadata was missing, and expected events for ERC721 need to be emitted.


contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721.sol
```diff
@@ -23,7 +28,18 @@ contract LSP8CompatibilityForERC721 is
...
+    ) LSP8IdentifiableDigitalAsset(name_, symbol_, newOwner_) {
+        _registerInterface(_INTERFACEID_ERC721);
+        _registerInterface(_INTERFACEID_ERC721METADATA);
+    }
+
+    /*
+     * @dev Compatible with ERC721Metadata tokenURI.
+     */
+    function tokenURI(uint256 tokenId) public view virtual override returns (string memory) {
+        bytes memory data = ERC725Utils.getDataSingle(this, _LSP4_METADATA_KEY);
```
CJ42, line 40: This will return the whole JSONURL value, not just the URI encoded part.
We might need to chunk the JSONURL bytes to the encoded URI part (36th bytes), as per the LSP specs:
https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#example-2

CJ42 (commented):
Added 2 comments :)

frozeman (approved):

frozeman:
@mattgstevens please solve the conflicts, and address the comment. 
I am not sure if we need to add the ERC721 to calculate interface IDs, as we do that for the LSP2 standards, not necessarily for legacy standards. @CJ42 whats your opinion on that?

CJ42:
@frozeman
I think it might be relevant (and be easier) for front/back-end clients to check the interface ID of the token / NFT contract first.
So that they know which ABI to use to create an instance (in web3.js / ethers.js).

Otherwise, it might be cumbersome to compare each function selector (for instance for the `transfer` function between the LSP7/8 version _vs_ the compatibility).

mattgstevens:
@CJ42 if you are also happy, this can be merged.

CJ42 (approved):
All good to me.

PR #91, opened 2021-12-20
Feat/URD Optimizing

YamenMerhi:
## What Changes this PR introduce ?
- [x] Add auto-generated docs to .gitignore at the moment
- [x] Add additional if statement to check if the owner of the UniversalProfile is a KeyManager in the **URD contract** 
- **As we agreed on our last meeting**, the URD should allow the call without trying to perform any action on UP, if the owner is not a KM.
**ERC165Checker** is used to check if the address of the owner of the UP supports `_LSP6_INTERFACEID`


frozeman (approved):
please add auto generated docs again to the repo. we will decide how we do them, but they should be part of the repo.

and merge after yourself

PR #93, opened 2021-12-29
feat: constants generator for multiple languages;

JeneaVranceanu:
Please, see this [README.md](https://github.com/JeneaVranceanu/lsp-universalprofile-smart-contracts/tree/feat/constants_generator/constants) file for a description and a list of pros and cons.

Here is an archive with 3 source code files in Kotlin, TypeScript and Swift that are automatically generated:
[generated_files_example.zip](https://github.com/lukso-network/lsp-universalprofile-smart-contracts/files/7822045/generated_files_example.zip)


constants/README.md
```diff
@@ -0,0 +1,298 @@
...
+- as time goes by languages' versions supported by these scripts are inevitably going to be outdated. Though, most likely the backwards compatibility will be there anyway. It's the new syntax features that we won't get a taste of until we update scripts manually. Language versions incompatibility is **very unlikely** to be happening any time soon.
+
+## How to use?
+
+First of all, when a new constant must be added you should think whether it is an `enum`, `JSONSchema`, `json` object or a single line declaration - `const`.
+
+Supported types:
+- `enum`;
+- `json`;
+- `JSONSchema`;
```
CJ42, line 25: `JSONSchema` is a bit generic.
I would be a bit more specific and specify `ERC725Y` in the type. So `ERC725Y_JSONSchema`, or something like that.

constants/README.md
```diff
@@ -0,0 +1,298 @@
...
+```
+/// binary = .... 1111 1111 1111 (only 1s)
+public val ALL_PERMISSIONS_SET = "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+```
+
+### Types requirements.
+
+<h4>`enum`</h4>
+
+Enumeration requires name, raw type like `String` or any other primitive type, and a list of cases (at least an empty list or scripts will crash).
```
CJ42, line 81:
```suggestion
Enumeration requires name, raw type (`rawValueType`) like `String` or any other primitive type, and a list of cases (at least an empty list or scripts will crash).
```

CJ42 (commented):
@JeneaVranceanu added comments.

I think this would be very useful for developers that want to build interfaces in other languages / frameworks than Javascript.

Let's say if I want to create a dashboard to manage my Universal Profile.

If I want to use a stack like Python + [Web3.py](https://web3py.readthedocs.io/en/stable/) and [Django](https://www.djangoproject.com/), this tool would be quite helpful to get me all the data that I need in the first place, and then get started easily (just need to research the values I need through the LSP specs + docs, and then add them in the JSON file to generate the constants). 

So quite useful in removing the need of coding the constants manually (which could be error prone).

constants/README.md
JeneaVranceanu: Will do. Thanks for the suggestion!

constants/README.md
JeneaVranceanu: Nice catch!

constants/README.md
JeneaVranceanu: Updated.

frozeman:
Might be better as its own standalone tool `tools-lsp2-constant-generator` ??

JeneaVranceanu:
> Might be better as its own standalone tool `tools-lsp2-constant-generator` ??

Absolutely better as a standalone tool since it can and will be used by different platforms (web, mobile, desktop; any frontend/backend) while the current repository is used by the smart contract and backend teams only as I understand.

JeneaVranceanu:
Must add support for error types: https://github.com/lukso-network/lsp-smart-contracts/commit/8e3213c298b3f0f1fbcefa5036e78edb4c5780ff

JeneaVranceanu:
Must add support for length and index keys in array types:
```
// use this key to get the number of elements in the array
length: '0x114bd03b3a46d48759680d81ebb2b414fda7d030a7105a851867accf1c2352e7',
// use this key + bytes16(index) to access an index in the array
index: '0x114bd03b3a46d48759680d81ebb2b414',
```

https://github.com/lukso-network/lsp-smart-contracts/commit/b10beec3314edebbb5bf0f39957cb59f876e6ab0#diff-e7ad5f13058c7d8fe78b1910deadef9c6f714595824f73d12f572e8b7e9d258dR91

PR #98, opened 2022-01-11
Lsp8compat proxy

mattgstevens:
## Description
Adds a contract LSP8CompatabilityForERC721Init for the proxy version. This was left out after the change from an abstract contract to a constructor version, and will remain so now that this compatability version should register the ERC165 interfaces for ERC721 and ERC721Metadata.


CJ42 (approved):

contracts/LSP8IdentifiableDigitalAsset/extensions/ILSP8CompatibilityForERC721.sol
```diff
@@ -90,14 +90,14 @@ interface ILSP8CompatibilityForERC721 is ILSP8IdentifiableDigitalAsset {
      * @param tokenId The tokenId to query
      * @return The address of the operator for `tokenId`
      */
-    function getApproved(uint256 tokenId) external returns (address);
+    function getApproved(uint256 tokenId) external view returns (address);
 
     /*
      * @dev Compatible with ERC721 tokenURI.
      * @param tokenId The tokenId to query.
      * @return True if all tokenId are approved, false otherwise.
```
CJ42, line 98:
```suggestion
     * @return True if all tokenId are approved for this `operator`, false otherwise.
```

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721Core.sol
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * @dev LSP8 extension, for compatibility for clients / tools that expect ERC721.
+ */
+abstract contract LSP8CompatibilityForERC721Core is
+    LSP8IdentifiableDigitalAssetCore,
+    LSP4Compatibility,
+    ILSP8CompatibilityForERC721
+{
+    using EnumerableSet for EnumerableSet.AddressSet;
```
CJ42, line 27: Add ERC725Utils in the list of library being used.

```suggestion
    using ERC725Utils for IERC725Y;
    using EnumerableSet for EnumerableSet.AddressSet;
```

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721Core.sol
```diff
@@ -0,0 +1,243 @@
...
+        public
+        view
+        virtual
+        override
+        returns (string memory)
+    {
+        // silence compiler warning about unused variable
+        tokenId;
+
+        bytes memory data = ERC725Utils.getDataSingle(this, _LSP4_METADATA_KEY);
```
CJ42, line 42: I would suggest this change for readability.

```suggestion
        bytes memory data = IERC725Y(this).getDataSingle(_LSP4_METADATA_KEY);
        bytes memory data = ERC725Utils.getDataSingle(this, _LSP4_METADATA_KEY);
```

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721Core.sol
```diff
@@ -0,0 +1,243 @@
...
+    /*
+     * @inheritdoc ILSP8CompatibilityForERC721
+     */
+    function isApprovedForAll(address tokenOwner, address operator)
+        public
+        view
+        virtual
+        override
+        returns (bool)
+    {
```
CJ42, line 123: Why does it just return `false`?

CJ42 (commented):
See comments below

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721Core.sol
mattgstevens: ok can do

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721Core.sol
mattgstevens: ok can do

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibilityForERC721Core.sol
mattgstevens: returning false here as during our discussions for LSP8 we decided not to have the "operator for all" concept.

this means that an operator address must be approved for each tokenId. an inheriting contract could override and implement something else if they want, but this seemed like the best option for a default implementation

contracts/LSP8IdentifiableDigitalAsset/extensions/ILSP8CompatibilityForERC721.sol
mattgstevens: good catch. changed to use a similar statement as the original ERC721 & fixed the @param entries

CJ42 (approved):
Good to go for me

PR #103, opened 2022-01-14
Lsp7compat proxy and erc20 events

mattgstevens:
## Description
Adds a contract LSP7CompatabilityForERC20Init for the proxy version. This was left out after the change from an abstract contract to a constructor version, and is being implemented this way to support the pattern to have a constructor and proxy version.

### DIFF
--- a/contracts/LSP7DigitalAsset/LSP7DigitalAssetInitAbstract.sol
+++ b/contracts/LSP7DigitalAsset/LSP7DigitalAssetInitAbstract.sol
@@ -15,9 +15,9 @@ import "../LSP4DigitalAssetMetadata/LSP4DigitalAssetMetadataInitAbstract.sol";
  * @dev Proxy Implementation of a LSP7 compliant contract.
  */
 abstract contract LSP7DigitalAssetInitAbstract is
+    LSP7DigitalAssetCore,
     Initializable,
-    LSP4DigitalAssetMetadataInitAbstract,
-    LSP7DigitalAssetCore
+    LSP4DigitalAssetMetadataInitAbstract
 {
     /**
      * @notice Sets the token-Metadata and register LSP7InterfaceId
--- a/contracts/LSP7DigitalAsset/extensions/ILSP7CompatibilityForERC20.sol
+++ b/contracts/LSP7DigitalAsset/extensions/ILSP7CompatibilityForERC20.sol
@@ -10,7 +10,8 @@ import "../ILSP7DigitalAsset.sol";
  */
 interface ILSP7CompatibilityForERC20 is ILSP7DigitalAsset {
     /**
-     * @dev To provide compatibility with indexing ERC20 events.
+     * @notice To provide compatibility with indexing ERC20 events.
+     * @dev Emitted when `amount` tokens is transferred from `from` to `to`.
      * @param from The sending address
      * @param to The receiving address
      * @param value The amount of tokens transfered.
@@ -18,7 +19,8 @@ interface ILSP7CompatibilityForERC20 is ILSP7DigitalAsset {
     event Transfer(address indexed from, address indexed to, uint256 value);
 
     /**
-     * @dev To provide compatibility with indexing ERC20 events.
+     * @notice To provide compatibility with indexing ERC20 events.
+     * @dev Emitted when `owner` enables `spender` for `value` tokens.
      * @param owner The account giving approval
      * @param spender The account receiving approval
      * @param value The amount of tokens `spender` has access to from `owner`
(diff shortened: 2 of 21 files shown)


YamenMerhi (approved):

contracts/LSP7DigitalAsset/LSP7DigitalAssetInitAbstract.sol
```diff
@@ -15,9 +15,9 @@ import "../LSP4DigitalAssetMetadata/LSP4DigitalAssetMetadataInitAbstract.sol";
  * @dev Proxy Implementation of a LSP7 compliant contract.
  */
 abstract contract LSP7DigitalAssetInitAbstract is
+    LSP7DigitalAssetCore,
```
CJ42, line 18: We should arrange the inheritance order. So to put most base contract on the left, most derived on the right.
Here the order should be:
1. `Initializable`
2. `LSP7DigitalAssetCore`
3. `LSP4DigitalAssetMetadataInitAbstract`

contracts/LSP7DigitalAsset/extensions/ILSP7CompatibilityForERC20.sol
```diff
@@ -9,6 +9,22 @@ import "../ILSP7DigitalAsset.sol";
  * @dev LSP8 extension, for compatibility for clients / tools that expect ERC20.
  */
 interface ILSP7CompatibilityForERC20 is ILSP7DigitalAsset {
+    /**
+     * @dev To provide compatibility with indexing ERC20 events.
+     * @param from The sending address
+     * @param to The receiving address
+     * @param value The amount of tokens transfered.
+     */
+    event Transfer(address indexed from, address indexed to, uint256 value);
```
CJ42, line 18: Maybe adding the tag 

```
* @inheritdoc ILSP7CompatibilityForERC20
```

So that it inherits the `@dev` docs comment below?
```
* @dev Emitted when `amount` tokens is transferred from `from` to `to`.
```

contracts/LSP7DigitalAsset/extensions/LSP7CompatibilityForERC20Core.sol
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    function _transfer(
+        address from,
+        address to,
+        uint256 amount,
+        bool force,
+        bytes memory data
+    ) internal virtual override {
+        super._transfer(from, to, amount, force, data);
```
CJ42, line 89: I am not sure how it is possible to solve this.
But here two `Transfer` events will be emitted:

- 1st: the `Transfer` event from `LSP7DigitalAssetCore`: https://github.com/lukso-network/lsp-smart-contracts/blob/c5a7a5f51060605a06388f9fb3cc2d04c462702e/contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol#L310

- 2nd: the backward-compatible `Transfer` event defined below

Is there a way to not emit the event from the parent `_transfer()` function? This would save some gas, as the parent `Transfer` event has several parameters.

CJ42 (commented):
Added comments.
We can fix the inheritance order in a separate PR.

contracts/LSP7DigitalAsset/extensions/ILSP7CompatibilityForERC20.sol
mattgstevens: will see what the generated docs look like for that, have not tried.

contracts/LSP7DigitalAsset/extensions/LSP7CompatibilityForERC20Core.sol
mattgstevens: im not sure that this is desired, nor that it is possible. 

the parent `_transfer()` has the core logic for updating balances & will emit the LSP7 Transfer event. the purpose of this contract to additionally emit an event that can be picked up by indexers that expect a ERC20 Transfer event.

my understanding of the intention here was not to have only one Transfer event, but to support ERC20 indexing.

CJ42: Ok make sense then. Let's mark this as resolved

contracts/LSP7DigitalAsset/extensions/ILSP7CompatibilityForERC20.sol
mattgstevens: decided to add a @notice for why we have this event

CJ42 (approved):
Good to go for me

PR #109, opened 2022-02-21
fix: security check in UniversalReceiverDelegateUP contract

YamenMerhi:
- Add a security check to confirm that the profile(msg.sender) is the same profile linked to the extracted Key Manager.

### DIFF
--- a/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/Handling/TokenAndVaultHandling.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/Handling/TokenAndVaultHandling.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 // modules
 import "@erc725/smart-contracts/contracts/ERC725Y.sol";
+import "../../../LSP6KeyManager/LSP6KeyManager.sol";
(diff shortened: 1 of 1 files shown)


contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/Handling/TokenAndVaultHandling.sol
```diff
@@ -186,4 +189,11 @@ abstract contract TokenAndVaultHandlingContract {
             amount = 1;
         }
     }
+
+    function _profileChecker(address keyManagerAddress) private {
+        address profileAddress = address(
+            LSP6KeyManager(keyManagerAddress).account()
+        );
+        require(profileAddress == msg.sender, "Incorrect KeyManager");
```
CJ42, line 197: Maybe this error message should be a bit more specific? So explain precisely what is wrong in the interaction.
(The fact that the profile linked to the KeyManager is not the actual profile making the call).

If this error comes on Blockscout, it might be hard to debug without having to look at the Solidity source code.

**NB:** maybe it would be nice in the revert reason string to mention actually that it reverted for **Security** 😉

CJ42 (commented):
added comment

CJ42 (approved):

PR #111, opened 2022-02-23
test: remove duplicates tests for UniversalProfile by creating a `UniversalProfile.behaviour.ts` file

CJ42:

- [x] organise unit tests for **Universal Profile** contracts by creating a `UniversalProfile.behaviour.ts` file that include them all.
- [x] merge tests related to `LSP3IssuedAssets[]` in `Helpers/LSP0ERC725Account.spec.ts` inside `UniversalProfile.behaviour.ts`

- keep only one test in `UniversalProfile.behaviour.ts` about emitting `UniversalReceiver` event.

- remove all the deprecated tests related to `UniversalReceiverAddressStore`, `LSP4DigitalCertificate` and `ERC777` (These have been replaced with test about `LSP7` and `LSP8` in `LSP1UniversalReceiverDelegateUP`.

### DIFF
--- a/tests/UniversalProfile.behaviour.ts
+++ b/tests/UniversalProfile.behaviour.ts
@@ -146,7 +146,7 @@ export const shouldBehaveLikeLSP3 = (
         );
 
         lsp3IssuedAssetsValues.push(
-          new ethers.Wallet.createRandom().address.toLowerCase()
+          ethers.Wallet.createRandom().address.toLowerCase()
         );
 
         let keys = [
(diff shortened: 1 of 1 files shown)


YamenMerhi (approved):

tests/utils/helpers.ts
```diff
@@ -93,11 +93,11 @@ export const TOKEN_ID = {
...
 
 export function getRandomAddresses(count) {
-  let base = "0xa56039d89BD9451A1ac94a680a20302da2dE92";
-
   let addresses = [];
-  for (let ii = 10; ii < count + 10; ii++) {
-    let randomAddress = "0x" + parseInt(base + ii).toString(16);
+  for (let ii = 0; ii < count; ii++) {
+    // addresses stored under ERC725Y storage have always lowercases character.
+    // therefore, disable the checksum by converting to lowercase to avoid failing tests
```
CallumGrindle, line 99: Do we test that saving both checksum and non checksum addresses works okay? The contract should be able to handle both without failing (im not totally sure of the context of this function so maybe it fails because of something else)

CJ42: Basically whether you pass a checksummed or non-checksummed address, the contract will convert everything to lower case (because the bytes "letters" (`a`, `b`, `c`, `d`, `e`, `f`) are always lower case.

so all the upper case characters will be converted to lower case when setting.
And so when retrieving, you will receive back as lower cases automatically.

frozeman (approved):

PR #113, opened 2022-02-24
ci: checkout main before pushing iOS artifacts

CallumGrindle:
Improve iOS artefacts release. Checks out main before publishing swift SDK


JeneaVranceanu:
Looks good to me!

CJ42 (approved):

PR #114, opened 2022-02-25
feat: add LSP11BasicSocialRecovery implementation

YamenMerhi:
# Description
- [x] Add **LSP11BasicSocialRecovery** contract
- [x] Add LSP11BasicSocialRecovery tests 
- [x] Add `setupPermission(..)`  function in **LSP6Utils**
- [x] Add **fixtures.ts** to avoid code duplication across tests


contracts/LSP11SocialRecovery/LSP11SocialRecovery.sol
```diff
@@ -0,0 +1,35 @@
...
+     * @notice Initiate the contract with the address of the ERC725 contract, set the owner, the recover threshold
+     * and register LSP11SocialRecovery InterfaceId
+     * @param _account The address of the ER725 contract to recover
+     * @param newOwner The address to control the social recovery contract
+     * @param newThreshold The minimum number of votes needed to add an address as an owner
+     * controller of the account through the KeyManager
+     */
+    constructor(
+        address newOwner,
+        address _account,
```
CJ42, line 23: Wouldn't it be better to use the interface type, instead of an `address` as argument?
```suggestion
        ERC725 _account,
```

contracts/LSP11SocialRecovery/LSP11SocialRecovery.sol
```diff
@@ -0,0 +1,35 @@
...
+        uint256 newThreshold
+    ) {
+        account = ERC725(_account);
+
+        if (newOwner != owner()) {
+            OwnableUnset.initOwner(newOwner);
+        }
+
+        recoverThreshold = newThreshold;
+        _registerInterface(_INTERFACEID_LSP11);
```
CJ42, line 33: Since the contract implements only one interface, it might be easier and cheaper that it inherit from a plain `ERC165` contract. See the one from OZ here -> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/introspection/ERC165.sol

So we could write a `supportsInterface(...)` function inside the Core contract:

```solidity
    /**
     * @dev See {IERC165-supportsInterface}.
     */
    function supportsInterface(bytes4 _interfaceId) public view virtual override returns (bool) {
        return _interfaceId == _INTERFACEID_LSP11 ||
            super.supportsInterface(_interfaceId);
    }
```

contracts/LSP11SocialRecovery/LSP11SocialRecoveryCore.sol
```diff
@@ -0,0 +1,180 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+// modules
+import "@erc725/smart-contracts/contracts/utils/OwnableUnset.sol";
+import "@openzeppelin/contracts/utils/introspection/ERC165Storage.sol";
```
CJ42, line 6: Maybe putting `@erc725` import first, and then `@openzeppelin` second?

contracts/LSP11SocialRecovery/LSP11Constants.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: CC0-1.0
+pragma solidity ^0.8.0;
+
+// --- ERC165 interface ids
+
+bytes4 constant _INTERFACEID_LSP11 = 0x65c8d0e5;
```
CJ42, line 6: It might be a bit overkill to have one file for just one constant.
Putting this constant directly above the Core contract in the `LSP11SocialRecoveryCore` should be enough in my opinion.

contracts/LSP11SocialRecovery/LSP11SocialRecoveryCore.sol
```diff
@@ -0,0 +1,180 @@
...
+ */
+abstract contract LSP11SocialRecoveryCore is
+    ILSP11SocialRecovery,
+    OwnableUnset
+{
+    // The account to recover
+    ERC725 public account;
+
+    // The hash of the secret word set by the owner
+    bytes32 private _secretHash;
```
CJ42, line 35: Was not sure about the order of variables based on the visibility.
So did a quick research.
Based on:

- what is said in the Solidity docs (although about **functions**): https://docs.soliditylang.org/en/v0.8.12/style-guide.html#order-of-functions
- these two contracts from `@openzeppeling/contracts`:
    - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/abdb20a6bdb1700d58ea9e01b7471dafdef52a68/contracts/governance/TimelockController.sol#L24-L27
    - https://github.com/OpenZeppelin/openzeppelin-contracts/blob/abdb20a6bdb1700d58ea9e01b7471dafdef52a68/contracts/governance/Governor.sol#L32-L50
    
`public` state variables should be first.
`private` and `internal` state variables should go next.

contracts/LSP11SocialRecovery/LSP11SocialRecoveryCore.sol
```diff
@@ -0,0 +1,180 @@
...
+
+    // The minimum number of votes that an address should get
+    // to be granted the owner permissions in the ERC725 contract
+    uint256 public recoverThreshold;
```
CJ42, line 39: Would it be relevant to have some of these `public` getter defined in the interface too?
Maybe not all, but some that might be relevant across implementations.

So defined as follow `ILSP11SocialRecovery`:

```solidity
contract ILSP11SocialRecovery {

    // The number of recovery attempt
    function recoverCounter() external view returns (uint256);

    // The minimum number of votes that an address should get
    // to be granted the owner permissions in the ERC725 contract
    function recoverThreshold() external view returns (uint256);

     // The current number of the guardians
    function guardiansNumber() external view returns (uint256);

    ...

}

```

contracts/LSP11SocialRecovery/LSP11SocialRecoveryCore.sol
```diff
@@ -0,0 +1,180 @@
...
+
+    // The number of recovery attempt
+    uint256 public recoverCounter;
```
CJ42, line 42: Is it safe to make the counter `public`?

CJ42 (commented):
added comments

contracts/LSP11SocialRecovery/LSP11SocialRecovery.sol
YamenMerhi: Absolutely, let's do it also in LSP6.

contracts/LSP11SocialRecovery/LSP11SocialRecoveryCore.sol
YamenMerhi: Good idea, to discuss

contracts/LSP11SocialRecovery/LSP11SocialRecoveryCore.sol
YamenMerhi: To discuss

contracts/LSP11SocialRecovery/LSP11SocialRecovery.sol
CJ42: Only thing about this is that like LSP6, it should be possible to attach this contract to an LSP8 NFT (as mentioned by @frozeman). But an LSP8 uses on ERC725Y (not the full ERC725 with both X + Y).

The contract type parameter might not then be possible in achieving this goal. So instead, maybe we should check that the `address` passed in the constructor implements **either ERC725X or ERC725Y**.

```solidity
constructor(
        address newOwner,
        address _account,
        uint256 newThreshold
    ) {
        require(
            ERC165Lib.supportsInterface(_account, _INTERFACEID_ERC725X) ||
            ERC165Lib.supportsInterface(_account, _INTERFACEID, ERC725Y),
            "linked account is not an ERC725 X or Y contract"
        );
    
        // not sure then what this assignment should look like?
        account = ERC725(_account);

        if (newOwner != owner()) {
            OwnableUnset.initOwner(newOwner);
        }

        recoverThreshold = newThreshold;
        _registerInterface(_INTERFACEID_LSP11);
    }
```

@frozeman @YamenMerhi let me know your thoughts :)

YamenMerhi: Ah yeah, or we could put the type as ERC725Y instead of checking the interface, in this way we allow profiles/tokens

contracts/LSP11SocialRecovery/LSP11Constants.sol
YamenMerhi: To discuss

contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -54,4 +55,57 @@ library LSP6Utils {
...
+        }
+    }
+
+    function setDataViaKeyManager(
+        address keyManagerAddress,
+        bytes32[] memory keys,
+        bytes[] memory values
+    ) internal returns (bytes memory result) {
+        bytes memory payload = abi.encodeWithSelector(
+            hex"14a6e293",
```
CJ42, line 105: Maybe it would be better for readability to use `setData.selector`?  (or in case the functions signatures / selectors in the `IERC725Y` interface change).

```suggestion
           IERC725Y.setData.selector
```

Or since we are potentially thinking to use function overloading for `setData(...)` in ERC725Y, we could use:

```solidity
bytes memory payload = abi.encodeWithSignature(
    "setData(bytes32[],bytes[])",
    keys,
    values
);
```

CJ42 (commented):
added comments

CJ42 (approved):

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,242 @@
...
+                    _guardiansVotes[recoverCounter][recoverProcessId][_guardians.at(i)] ==
+                    msg.sender
+                ) {
+                    senderVotes++;
+                }
+            }
+        }
+        require(senderVotes >= _guardiansThreshold, "You didnt reach the threshold");
+
+        require(keccak256(abi.encodePacked(plainSecret)) == _secretHash, "Wrong secret");
```
CJ42, line 237: why not putting this `require` statement at the beginning as the first line to run in the function?

contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -104,4 +104,31 @@ library LSP6Utils {
         bytes memory payload = abi.encodeWithSelector(hex"14a6e293", keys, values);
         result = ILSP6KeyManager(keyManagerAddress).execute(payload);
     }
+
+    function setupPermissions(
```
CJ42, line 108: We should also rename this function. `setupPermissions` is very generic, and does not describe what is happening.
Also IMO the word `setup` implies an interaction is happening, while it's not the case.

The function only create key-value pairs and return them:
- create the key-value pair for the permission array length with value as previous length +1
- create the key-value pair of the address of the new controller for the permission array index
- create the key-value pair of the permissions of the new controller address

I really like this function tho. It's super useful and powerful. It does a good part of the heavy lifting job and devs will definitely re-use this kind of library function in the future 😃 💪 

Any ideas of the naming? A long name is not too much of a problem, but should read and be understandable when reading in an implementation.

CJ42 (changes requested):
@YamenMerhi added some comments.

Let's not forget to also add the tests for LSP11 to run in the CI!
Both in `package.json` + in the CI workflow.

contracts/LSP6KeyManager/LSP6Utils.sol
CJ42: @YamenMerhi @skimaharvey here are my suggestion of function renaming:

```solidity
function createPermissionsKeysForController(
        IERC725Y account,
        address controller,
        bytes memory permissions
    ) internal view returns (bytes32[] memory permissionsKeys, bytes[] memory permissionsValues) {
```

skimaharvey: Agreed, since the length of function has no effect on gas, I think it makes more sense calling it this way even though it is long

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,242 @@
...
+abstract contract LSP11BasicSocialRecoveryCore is OwnableUnset, ERC165, ILSP11BasicSocialRecovery {
+    using EnumerableSet for EnumerableSet.AddressSet;
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+
+    // ---- state variables
+
+    /**
+     * @inheritdoc ILSP11BasicSocialRecovery
+     */
+    address public override account;
```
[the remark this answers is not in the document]
skimaharvey: I think we should update to 0.8.10 and above. Prior 0.8.10 version, compiler was adding extra codes (including `EXTCODESIZE` which is expensive). Since then, it is not adding these codes when return data is expected. We make quite a lot of external calls, may be a good idea to think about it? Wdut? 
https://github.com/ethereum/solidity/releases/tag/v0.8.10

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,240 @@
...
+        unchecked {
+            for (uint256 i = 0; i < guardiansLength; i++) {
+                if (
+                    _guardiansVotes[recoverCounter][recoverProcessId][_guardians.at(i)] ==
+                    msg.sender
+                ) {
+                    senderVotes++;
+                }
+            }
+        }
```
skimaharvey, line 235: Nice I think if we want to optimise even more we could do:
```
        unchecked {
            for (uint256 i = 0; i < guardiansLength; ) {
                if (
                    _guardiansVotes[recoverCounter][recoverProcessId][_guardians.at(i)] ==
                    msg.sender
                ) {
                    ++senderVotes;
                }
                ++i 
            }
        }
```

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,242 @@
...
+        return _guardiansVotes[_recoveryCounter][recoverProcessId][guardian];
+    }
+
+    // ---- public functions
+
+    /**
+     * @inheritdoc ILSP11BasicSocialRecovery
+     */
+    function addGuardian(address newGuardian) public virtual override onlyOwner {
+        require(!_guardians.contains(newGuardian), "Provided address is already a guardian");
```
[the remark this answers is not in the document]
skimaharvey: if we dont we probably would need to reduce the length of the error's message as it is > 32 bytes

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: I think it will be too much, IMO readability > gas opt @CJ42 opinion ?

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,242 @@
...
+    ) public virtual override NotZeroBytes32(newHash) {
+        _checkRequirements(recoverProcessId, plainSecret);
+
+        _secretHash = newHash;
+        // Starting new recover counter
+        unchecked {
+            _recoveryCounter++;
+        }
+
+        address keyManager = ERC725(account).owner();
```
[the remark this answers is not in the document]
YamenMerhi: Actually, I have a different opinion. 

First, deploying and setting up the social recovery contract could be done before having the key manager as owner, this will reduce the cost of mapping every transaction through the key Manager. 

Second, If we made this check in the constructor and removed the check in the `recoverOwnership(...)`, we will have a problem, because let's say we made the check in the constructor and at that time the erc725 contract was owned by LSP6KeyManager but then after a while, the owner changed to an EOA, then while executing the `recoverOwnership(...)` and because we don't have a check for the owner that it supports LSP6, then the transaction will be executed on an EOA successfully and we will not have a revert error and this is wrong ❌ .

That's why IMO it doesn't matter who is the owner of the linked account on the deployment, it matters when we are recovering and the check should be there to ensure 100% that the owner of the linked account support LSP6.

Let's discuss @CJ42 @skimaharvey 🚀

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: The place of the `require` statement was intended to be last because it's the last thing that should be checked.

Although it's a fragile security improvement, you should not reveal that the `plainSecret` is right and the error is coming from the threshold, otherwise, malicious guardians will be knowing your plainSecret and now can vote to recover on another processId and take over your account since they will know the plainSecret.

Example: Let's say you are recovering and you provided the right secret word and you didn't reach the threshold yet.

Using current code, the call will revert and say that you didn't reach the threshold, and the guardians will not know that the plainSecret is right or not because the check of plainSecret is after the check of the threshold. (Even if they can check that the plainSecret is right or not in a manual way)

Using your suggestion, the call will revert and say that you didn't reach the threshold, but at this stage since it's reverting at this line of code, that means that it passed the check of the `plainSecret`, which could be a little bit more dangerous.

Let's discuss @CJ42 @skimaharvey 🚀

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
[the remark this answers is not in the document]
YamenMerhi: Yep let's discuss this later. Added a note in Notion

contracts/LSP11BasicSocialRecovery/ILSP11BasicSocialRecovery.sol
```diff
@@ -0,0 +1,113 @@
...
+     * @dev Returns the address voted for by the provided guardian in the
+     * provided recoverProcessId
+     */
+    function getGuardianVote(bytes32 recoverProcessId, address guardian)
+        external
+        view
+        returns (address);
+
+    /**
+     * @dev Adds a guardian of the owner
```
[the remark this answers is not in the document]
YamenMerhi: Guardian of the owner, the owner is the linked account.

contracts/LSP11BasicSocialRecovery/ILSP11BasicSocialRecovery.sol
```diff
@@ -0,0 +1,113 @@
...
+     * @param _guardiansThreshold The threshold to set
+     *
+     * Requirements:
+     *
+     * - `_guardiansThreshold` cannot be 0
+     * - `_guardiansThreshold` cannot be more than the guardiansCount
+     */
+    function setThreshold(uint256 _guardiansThreshold) external;
+
+    /**
```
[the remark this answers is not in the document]
YamenMerhi: To discuss with frozeman

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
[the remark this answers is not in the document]
YamenMerhi: We already discussed the use of custom errors in our code, let's first find the solution for it then we can introduce them. I prefer custom errors over string errors :)
For now, let's limit the error message to < 32.

YamenMerhi:
@CJ42 @skimaharvey Topics to discuss when frozeman is back:
- Introducing events
- Name of functions and variables 
- Custom errors, etc.

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
CJ42: Ok make totally sense I agree @YamenMerhi 👍

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
CJ42: I think it's ok for readability for to keep it as it is for now.

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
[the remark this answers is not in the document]
CJ42: Same more in favour of a custom error like a `NotGuardian(address _address)`. Will be easier to debug.

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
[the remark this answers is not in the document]
CJ42: ok make sense for initial setup

CJ42 (approved):

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,241 @@
...
+        require(
+            ERC165Checker.supportsERC165Interface(keyManager, _INTERFACEID_LSP6),
+            "Owner of account doesn't support LSP6 InterfaceId"
+        );
+
+        // Setting permissions for `msg.sender`
+        (bytes32[] memory keys, bytes[] memory values) = LSP6Utils
+            .createPermissionsKeysForController(
+                ERC725(account),
```
deepface19, line 198: got error when i try to compile
``` TypeError: Invalid type for argument in function call. Invalid implicit conversion from contract ERC725 
to contract IERC725Y requested.
   --> contracts/LSP11BasicSocialRecoveryCore.sol:198:17:
    |
198 |                 ERC725(account),
```

deepface19 (commented):
i got error when i try to compile

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,239 @@
...
+            for (uint256 i; i < guardians.length; i++) {
+                if (_guardiansChoice[currentRecoveryCounter][guardians[i]] == msg.sender)
+                    callerSelections++;
+            }
+        }
+
+        if (callerSelections < _guardiansThreshold) revert ThresholdNotReachedForCaller(msg.sender);
+        if (newHash == bytes32(0)) revert SecretHashCannotBeZero();
+        if (keccak256(abi.encodePacked(plainSecret)) != _recoverySecretHash)
+            revert WrongPlainSecret();
```
CJ42, line 221: suggested renaming:

```suggestion
            revert InvalidSecret();
```

b00ste (approved):

CJ42 (approved):

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,239 @@
...
+        uint256 currentRecoveryCounter,
+        string memory plainSecret,
+        bytes32 newHash,
+        address[] memory guardians
+    ) internal view {
+        if (msg.sender == address(0)) revert AddressZeroNotAllowed();
+        uint256 callerSelections;
+
+        unchecked {
+            for (uint256 i; i < guardians.length; i++) {
```
skimaharvey, line 212: cant we have a DoS situation with a big array of guardians? If so may be we should document it?

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
skimaharvey, line 221: why don't we check for that before looping on the guardians?

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,239 @@
...
+        _guardians.remove(existingGuardian);
+        emit RemovedGuardian(existingGuardian);
+    }
+
+    /**
+     * @inheritdoc ILSP11BasicSocialRecovery
+     */
+    function setGuardiansThreshold(uint256 newThreshold) public virtual onlyOwner {
+        if (newThreshold > _guardians.length())
+            revert ThresholdCannotBeHigherThanGuardiansNumber(newThreshold, _guardians.length());
```
skimaharvey, line 132: maybe checking that `newThreshold != 0`?

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
```diff
@@ -0,0 +1,239 @@
...
+                ALL_REGULAR_PERMISSIONS
+            );
+
+        LSP6Utils.setDataViaKeyManager(keyManager, keys, values);
+
+        emit RecoverProcessSuccessful(
+            currentRecoveryCounter,
+            msg.sender,
+            newSecretHash,
+            _guardians.values()
```
skimaharvey, line 190: do we really need to emits all the guardians?

skimaharvey (changes requested):
See thread on discord

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: It was pointed by Fabian, and i think it might make sense to know who was the guardians at this specific recovery process. But let's discuss

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: We allow threshold to be 0, so the contract can act like a password recovery

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: Good point, will document

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: @skimaharvey https://github.com/lukso-network/lsp-smart-contracts/pull/114#discussion_r929754074

skimaharvey: Hmm yes make it less likely to happen but if you know the `_recoverySecretHash` you can easily check if `keccak256(plainSecret)` is correct or not but I guess yes we can leave it there. No big deals.

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
skimaharvey: This guardians array could be quite big and costly when emitting. 
Probably would be in favor of not emitting them. @CJ42 @b00ste ?

skimaharvey (approved):

contracts/LSP11BasicSocialRecovery/LSP11BasicSocialRecoveryCore.sol
YamenMerhi: I wouldn't go with invalid as secretHash use the word invalid, what about `IncorrectSecret()` ?

b00ste (approved):

skimaharvey (approved):

CJ42 (approved):

PR #120, opened 2022-03-16
fix: `isValidSignature` function check

YamenMerhi:
- [x] Fix `isValidSignature(...)` function to check if the called contract supports ERC1271. If yes call, otherwise return FailValue.

### Current Behaviour
- `isValidSignature(...)` function is calling the owner (contract) even if it doesn't support the ERC1271.
This can lead to unwanted behaviour like reverting if the called contract doesn't implement the `isValidSignature(...)` function.


CJ42 (commented):
> This can lead to unwanted behaviour like reverting if the called contract doesn't implement the isValidSignature(...) function.

Would be good to add this extra test in `UniversalProfile.behaviour.ts`. So to ensure it does not happen anymore.

PR #121, opened 2022-03-18
fix: apply checks-effects-interactions pattern

YamenMerhi:
- [x] Moving the place of the hooks to be called after the state changing, following checks-effects-interactions pattern.
- [x] Modify the logic of the UP-URD and the Vault-URD to suit the hook changing.


CJ42 (commented):
The linter reported that the 3rd argument `data` is not used anymore.
https://github.com/lukso-network/lsp-smart-contracts/blob/d1ed5fc2ac087f9020b5b40f46c1aab9a577fe06/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/Handling/TokenAndVaultHandling.sol#L31

CJ42 (approved):

PR #128, opened 2022-03-31
fix: admin caller with `ALL_PERMISSIONS` can call any functions part of the Universal Profile's ABI

CJ42:

Since PR #123, the function selector in the payload sent to `execute(...)` or `executeRelayCall(...)` is not checked
for callers with `ALL_PERMISSIONS`.

This introduces a bug as callers with ALL_PERMISSIONS can now call functions like `revokeOwnership` on the Universal Profile.

## What is the expected behaviour?

Caller should be allowed to interact with only the 3 functions below on the Universal Profile, regardless of their permissions (admin or not).

- `setData(...)`
- `execute(...)`
- `transferOwnership(...)`

## Fix

Introduces an internal function to validate the function being called on the linked Universal Profile.

### DIFF
--- a/tests/LSP6KeyManager/tests/Security.test.ts
+++ b/tests/LSP6KeyManager/tests/Security.test.ts
@@ -101,6 +101,17 @@ export const testSecurityScenarios = (
     }
   });
 
+  it("should revert when admin with ALL PERMISSIONS try to call `renounceOwnership(...)`", async () => {
+    let payload =
+      context.universalProfile.interface.encodeFunctionData(
+        "renounceOwnership"
+      );
+
+    await expect(
+      context.keyManager.connect(context.owner).execute(payload)
+    ).toBeRevertedWith("_validateERC725Selector: invalid ERC725 selector");
+  });
+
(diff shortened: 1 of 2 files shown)


YamenMerhi (approved):
Great!

tests/LSP6KeyManager/tests/Security.test.ts
```diff
@@ -101,6 +101,17 @@ export const testSecurityScenarios = (
...
   });
 
+  it("should revert when admin with ALL PERMISSIONS try to call `renounceOwnership(...)`", async () => {
+    let payload =
+      context.universalProfile.interface.encodeFunctionData(
+        "renounceOwnership"
+      );
+
+    await expect(
+      context.keyManager.connect(context.owner).execute(payload)
```
JeneaVranceanu, line 111: A bit out of my competence but shouldn't `executeRelayCall` also be tested for the same behaviour?

CJ42: Yes it should!

PR #131, opened 2022-04-05
feat(lsp6)!: add `account()` function as part of the LSP6 interface

CJ42:

This PR introduces a breaking in the LSP6 interface. changing LSP6 interface ID.

- **LSP6 interface ID (before):** `0x6f4df48b `
- **LSP6 interface ID (after):**  `0x32e6d0ab `

## Features

- [x] add the standard `account()` function to return the `address` of the ERC725 account contract linked with this KeyManager.

## Refactoring

- [x] wrap contract calls to get permissions from storage within `ERC725Y(account)`.
- [x] use import statements with curly braces to import specific symbols from files.
- [x] improve readability of `isValidSignature(...)` function by using `LSP6Utils` library.

### DIFF
--- a/contracts/Helpers/KeyManager/KeyManagerHelper.sol
+++ b/contracts/Helpers/KeyManager/KeyManagerHelper.sol
@@ -13,7 +13,7 @@ contract KeyManagerHelper is LSP6KeyManager {
     constructor(address _account) LSP6KeyManager(_account) {}
 
     function getPermissionsFor(address _address) public view returns (bytes32) {
-        return account.getPermissionsFor(_address);
+        return ERC725Y(account).getPermissionsFor(_address);
     }
 
     function getAllowedAddressesFor(address _address)
@@ -21,7 +21,7 @@ contract KeyManagerHelper is LSP6KeyManager {
         view
         returns (bytes memory)
     {
-        return account.getAllowedAddressesFor(_address);
+        return ERC725Y(account).getAllowedAddressesFor(_address);
     }
 
     function getAllowedFunctionsFor(address _address)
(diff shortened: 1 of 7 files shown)


contracts/Helpers/KeyManager/KeyManagerHelper.sol
```diff
@@ -13,23 +13,23 @@ contract KeyManagerHelper is LSP6KeyManager {
     constructor(address _account) LSP6KeyManager(_account) {}
 
     function getPermissionsFor(address _address) public view returns (bytes32) {
-        return account.getPermissionsFor(_address);
+        return ERC725Y(account).getPermissionsFor(_address);
```
JeneaVranceanu, line 16: Is the `account` used here the same that is returned by the new `account()` function?

JeneaVranceanu: Looks like it.
I'm asking because it's said that new `account()` function can return, as one of the options, `ERC725X only` account.
While here it's always considered to be `ERC725Y`.

CJ42: Yes it is the same.
The `account()` function is intended to be used more by external contracts, to fetch the address of the linked account.

JeneaVranceanu: Won't it break if account will refer to a contract that doesn't support ERC725Y?

YamenMerhi (approved):

contracts/Helpers/KeyManager/KeyManagerHelper.sol
CJ42: As far as I am aware, the casting in the code snippet above is used to be able to call the `getData(...)` function.

The KeyManager:
- receives a payload in raw byte
- send this payload via a **low-level** call. So since low-level is used, the KeyManager does not check that these functions are present in the linked account.

So I would say what matters is actually not about if the linked account supports the standard. But rather if it has one the functions `setData(...)`, `execute(...)` and `transferOwnership(...)`.
If you try to send payload via

frozeman (approved):

PR #134, opened 2022-04-11
fix(lsp6): fix bugs for `AllowedERC725YKeys` when input is multiple keys that include allowed + not allowed keys

CJ42:

🐛  🩹  **Bug Fixes**

The `try { ... } catch { ... }` blocks in the test suite are misleading and do not run the assertion if the contract call goes successfully (while it is expected to revert).

- [x] re-write tests to check for proper revert handling.

After re-writing the test, it appears that the `AllowedERC725YKeys` feature does not work properly depending on the keys given as inputs.

- [x] refactor the nested loops when checking for `AllowedERC725YKeys`
- [x] refactor `AllowedERC725YKeys` to use a **mask** and ANDing `&` for comparing key slices = save around `2,000+ gas` per key comparison.
- [x] create file `LSP6Errors.sol` to contain all LSP6 custom errors definitions.

### DIFF
--- a/contracts/Helpers/KeyManager/KeyManagerHelper.sol
+++ b/contracts/Helpers/KeyManager/KeyManagerHelper.sol
@@ -32,6 +32,14 @@ contract KeyManagerHelper is LSP6KeyManager {
         return ERC725Y(account).getAllowedFunctionsFor(_address);
     }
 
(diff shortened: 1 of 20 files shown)


YamenMerhi (approved):
Great!

PR #136, opened 2022-04-11
build: increase number of runs for Solidity optimizer

CJ42:

Increase number of runs to **`10,000`** in the Solidity optimizer, so to optimize for contract interactions rather than deployment.

**gas costs LYX transfer:**
- everything allowed: -147 gas
- one allowed address: -240 gas
- one allowed address + one allowed standard: -348 gas


YamenMerhi (approved):
Cool

CJ42:
> Verified if increasing the number of runs in the optimizer increases the computation (and time) required to verify a contract on Blockscout:

 No, it actually reduces the verification time by 13 seconds.

CJ42:
Close as need more research on the implications of increasing the number of runs.

PR #143, opened 2022-04-19
refactor(lsp6): use bytes4 selector in `Executed` event

CJ42:

Similar to https://github.com/ERC725Alliance/ERC725/pull/98

Save gas by only use the first 4 bytes of the `_data` payload (= function selector) in the `Executed` event in LSP6.


YamenMerhi (approved):
Cool

PR #153, opened 2022-04-28
refactor: use `!= 0` instead of `> 0`

YamenMerhi:
- Changed all occurrences of `> 0` to `!= 0` because `!= 0` is cheaper than `> 0`.
These two statements are equivalent for uint values since they cannot be less than zero.


CJ42:
@YamenMerhi how much gas is saved?

YamenMerhi:
@CJ42 Not that much, for a token transfer it's 18 gas.

frozeman (approved):

Issue #155, opened 2022-05-04
Multiple Inheritance and Linearization


YamenMerhi:
## What is this issue about?
After upgrading the version of the **erc725-smart-contracts** to **3.0.1**, an error occurred related to the inheritance hierarchy.
The error provided was: `TypeError: Linearization of inheritance graph impossible`. The error was mainly in the **token contracts** that have a very complex inheritance hierarchy by extending from the **LSP4** and **ERC725Y** contracts.

The same problem exists in **erc725-smart-contracts** pointed in the issue: [#105](https://github.com/ERC725Alliance/ERC725/issues/105)

According to the solidity docs:

> Languages that allow multiple inheritances have to deal with several problems. One is the [Diamond Problem](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem). Solidity is similar to Python in that it uses “[C3 Linearization](https://en.wikipedia.org/wiki/C3_linearization)” to force a specific order in the directed acyclic graph (DAG) of base classes. This results in the desirable property of monotonicity but disallows some inheritance graphs. Especially, the order in which the base classes are given in the is directive is important: **You have to list the direct base contracts in the order from “most base-like” to “most derived”.**

My understanding of most base to most derive is opposed to the implementation we currently have, for example:
- Current version at [a402b28](https://github.com/lukso-network/lsp-smart-contracts/blob/a402b28dcf23995c07ea448ccf09d49f34d60a5b/contracts/LSP7DigitalAsset/extensions/LSP7CompatibilityForERC20.sol#L10): `contract LSP7CompatibilityForERC20 is LSP7CompatibilityForERC20Core, LSP7DigitalAsset`
- My Understanding of **Most Base to Most Derive**: `contract LSP7CompatibilityForERC20 is LSP7DigitalAsset, LSP7CompatibilityForERC20Core`
Because in my opinion, `LSP7DigitalAsset` is an existing module and the extension that is being built on top of is `LSP7CompatibilityForERC20Core`.

Looking at **OZ contracts**, they use the same approach [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/6bd6b76d1156e20e45d1016f355d154141c7e5b9/contracts/token/ERC721/extensions/ERC721Pausable.sol#L16): `contract ERC721Pausable is ERC721, Pausable`.

## Suggest Solutions
To fix those issues, we need to revert the changes of PR #108 and apply the same pattern (most base to most derive) on all other contracts.

PR #157, opened 2022-05-09
fix: inherit from most base to most derived

YamenMerhi:
- Inherit from most base to most derived: resolve issue #155 
- Fix inheritance problem in erc725 by upgrading erc725-smart-contracts to 3.0.2
- Override ERC165 `supportsInterface(...)` function In LSP0/LSP9/LSP7/LSP8 contracts.
- Resolve solhint warnings

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -25,10 +25,10 @@ import {_INTERFACEID_LSP1_DELEGATE, _LSP1_UNIVERSAL_RECEIVER_DELEGATE_KEY} from 
  * @dev Bundles ERC725X and ERC725Y, ERC1271 and LSP1UniversalReceiver and allows receiving native tokens
  */
 abstract contract LSP0ERC725AccountCore is
-    IERC1271,
-    ILSP1UniversalReceiver,
     ERC725XCore,
-    ERC725YCore
+    ERC725YCore,
+    IERC1271,
+    ILSP1UniversalReceiver
 {
     event ValueReceived(address indexed sender, uint256 indexed value);
 
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -16,8 +16,8 @@ import {ERC725YCore} from "@erc725/smart-contracts/contracts/ERC725YCore.sol";
 import {ERC725XCore} from "@erc725/smart-contracts/contracts/ERC725XCore.sol";
 
 // constants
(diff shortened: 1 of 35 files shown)


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -109,4 +109,21 @@ abstract contract LSP0ERC725AccountCore is
         }
         emit UniversalReceiver(_msgSender(), _typeId, returnValue, _data);
     }
+
+    /**
+     * @dev See {IERC165-supportsInterface}.
+     */
+    function supportsInterface(bytes4 interfaceId)
```
CJ42, line 116: Why the `supportInterface(...)` function has to be in both standard and Core version of the contract?

contracts/LSP9Vault/LSP9Vault.sol
```diff
@@ -76,7 +76,13 @@ contract LSP9Vault is LSP9VaultCore, ERC725 {
     /**
      * @dev See {IERC165-supportsInterface}.
      */
-    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+    function supportsInterface(bytes4 interfaceId)
```
CJ42, line 79: same question as for LSP0

contracts/LSP4DigitalAssetMetadata/LSP4DigitalAssetMetadataInitAbstract.sol
```diff
@@ -13,7 +13,7 @@ import "./LSP4Constants.sol";
  * @author Matthew Stevens
  * @dev Inheritable Proxy Implementation of a LSP8 compliant contract.
  */
-abstract contract LSP4DigitalAssetMetadataInitAbstract is Initializable, ERC725YInitAbstract {
+abstract contract LSP4DigitalAssetMetadataInitAbstract is ERC725YInitAbstract {
```
CJ42, line 16: Why `Initializable` had to be removed here?

CJ42 (commented):
added comments

contracts/LSP4DigitalAssetMetadata/LSP4DigitalAssetMetadataInitAbstract.sol
YamenMerhi: `Initializable` is already inherited in `ERC725YInitAbstract`

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: the `supportsInterface(...)` function should be overridden in LSP0Core because its inherited from XCore and YCore.
Same in the standard version, it should be overridden because it's inherited from ERC725 and LSP0Core.

contracts/LSP9Vault/LSP9Vault.sol
YamenMerhi: Same answer applies for LSP9

PR #160, opened 2022-05-11
feat: add ErrorHandlerLib in LSP6

YamenMerhi:
- Introducing `ErrorHandlerLib` in `execute(...)` and `executeRelayCall(..)` to support custom errors.

### DIFF
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -15,6 +15,7 @@ import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {ErrorHandlerLib} from "@erc725/smart-contracts/contracts/utils/ErrorHandlerLib.sol";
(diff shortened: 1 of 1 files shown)


contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -86,24 +86,23 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         _verifyPermissions(msg.sender, _calldata);
 
         // solhint-disable avoid-low-level-calls
-        (bool success, bytes memory result_) = address(target).call{
-            value: msg.value,
-            gas: gasleft()
-        }(_calldata);
+        (bool success, bytes memory result) = target.call{value: msg.value, gas: gasleft()}(
```
CJ42, line 90: Great for `result_` -> `result`

For the formatting, I would suggest that we put it back to how it was and stick to the line length = 100 in the prettier config for `.sol` files.

Otherwise, anyone working on the `LSP6KeyManagerCore` and making a PR will change the style back to how it is in red and it will show in the diff of any future PR.

CJ42 (commented):
We need to add tests for CALL, STATICCALL and DELEGATECALL that revert with custom errors.

YamenMerhi:
Didn't add more tests for now as I receive: `VM Exception while processing transaction: reverted with an unrecognized custom error`;
This is an issue with Hardhat where it doesn't recognize the external custom errors.
More details in this issue: https://github.com/NomicFoundation/hardhat/issues/1618
Will look at this later.

frozeman (approved):

PR #161, opened 2022-05-12
feat: introduce Claim Ownership pattern in LSP0

CJ42:

The Ownable pattern introduces some risk with `transferOwnership(address)`, where control over an Ownable contract can be lost if ownership is ownership is transferred to an `address` that cannot be controlled, or transferred to the wrong `address` by accident.

## New Feature

Introduce a new pattern called **Claim Ownership**, that requires a 2 step process for full contract ownership to be transferred:

step 1) transfer ownership to a new owner `address` (EOA or contract) via the standard function `transferOwnership(address)`
step 2) require this new owner to claim ownership over the contract, by calling the `claimOwnership()` function. 

 - [x] create `interface` and `abstract` contract for ClaimOwnership

**BREAKING CHANGES**
- Introduce `ClaimOwnership` in LSP0
 - [x] return true on `LSP0.supportsInterface(interfaceId)`, when passing the ClaimOwnership interfaceId as parameter.
 - [x] ⚠️ edit LSP0 interface ID to include ClaimOwnership function selector in XOR calculation.
 
 ## Refactor
 
 - [x] edit LSP0 `transferOwnership(address)` behaviour to use ClaimOwnership logic, instead of `OwnableUnset`
 - [x] add support for ClaimOwnership in LSP6, required caller to have permission CHANGEOWNER when calling function selector `claimOwnership()` on an LSP0 via the KeyManager.
 
 ## Tests
 
 - [x] add tests for `transferOwnership(address)` + `claimOwnership()` when transferring ownership from EOA to EOA
 - [x] update LSP6 tests for CHANGEOWNER when calling `claimOwnership()` via KeyManager.

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725Account.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725Account.sol
@@ -2,8 +2,12 @@
 pragma solidity ^0.8.0;
 
 // modules
-import "./LSP0ERC725AccountCore.sol";
-import "@erc725/smart-contracts/contracts/ERC725.sol";
+import {LSP0ERC725AccountCore} from "./LSP0ERC725AccountCore.sol";
+import {ERC725} from "@erc725/smart-contracts/contracts/ERC725.sol";
+
+// constants
+import {_INTERFACEID_LSP0, _INTERFACEID_ERC1271} from "./LSP0Constants.sol";
+import {_INTERFACEID_LSP1} from "../LSP1UniversalReceiver/LSP1Constants.sol";
 
 /**
  * @title Implementation of ERC725Account
--- a/contracts/LSP0ERC725Account/LSP0ERC725Account.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725Account.sol
@@ -15,7 +15,7 @@ import {_INTERFACEID_LSP1} from "../LSP1UniversalReceiver/LSP1Constants.sol";
  * @author Fabian Vogelsteller <fabian@lukso.network>, Jean Cavallera (CJ42), Yamen Merhi (YamenMerhi)
  * @dev Bundles ERC725X and ERC725Y, ERC1271 and LSP1UniversalReceiver and allows receiving native tokens
  */
-contract LSP0ERC725Account is LSP0ERC725AccountCore, ERC725 {
+contract LSP0ERC725Account is ERC725, LSP0ERC725AccountCore {
     /**
      * @notice Sets the owner of the contract
      * @param _newOwner the owner of the contract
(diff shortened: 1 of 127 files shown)


contracts/Utils/ClaimOwnership.sol
```diff
@@ -0,0 +1,23 @@
...
+    address public override pendingOwner;
+
+    function claimOwnership() public virtual override {
+        require(msg.sender == pendingOwner, "OwnableClaim: caller is not the pendingOwner");
+
+        _setOwner(pendingOwner);
+        pendingOwner = address(0);
+    }
+
+    function transferOwnership(address _newOwner) public virtual override(OwnableUnset) onlyOwner {
```
YamenMerhi, line 20: When having this pattern:
```solidity
    function transferOwnership(address _newOwner)
        public
        virtual
        override(ClaimOwnership, OwnableUnset)
        onlyOwner
    {
        ClaimOwnership.transferOwnership(_newOwner);
    }
```
The require of `onlyOwner` will run twice, not that much of gas difference but it's more of a structural thing.
So we can do this:
```solidity
    function transferOwnership(address _newOwner) public virtual override onlyOwner {
        _transferOwnership(_newOwner);
    }

    function _transferOwnership(address _newOwner) internal virtual {
        pendingOwner = _newOwner;
    }
```
And when having an overrided function:
```solidity
    function transferOwnership(address _newOwner)
        public
        virtual
        override(LSP0ERC725AccountCore, OwnableUnset)
        onlyOwner
    {
        _transferOwnership(_newOwner);
    }
```

contracts/LSP0ERC725Account/LSP0ERC725Account.sol
```diff
@@ -37,4 +38,13 @@ contract LSP0ERC725Account is ERC725, LSP0ERC725AccountCore {
...
             super.supportsInterface(interfaceId);
     }
+
+    function transferOwnership(address _newOwner)
+        public
+        virtual
+        override(LSP0ERC725AccountCore, OwnableUnset)
+        onlyOwner
+    {
+        super.transferOwnership(_newOwner);
```
YamenMerhi, line 48: Could be done like this as mentioned in the other note:
```solidity

    function transferOwnership(address _newOwner)
        public
        virtual
        override(LSP0ERC725AccountCore, OwnableUnset)
        onlyOwner
    {
        _transferOwnership(_newOwner);
    }

File: contracts/LSP0ERC725Account/LSP0ERC725AccountInitAbstract.sol [hunk 3]
```diff
@@ -35,4 +36,13 @@ abstract contract LSP0ERC725AccountInitAbstract is ERC725InitAbstract, LSP0ERC72
...
             super.supportsInterface(interfaceId);
     }
+
+    function transferOwnership(address _newOwner)
+        public
+        virtual
+        override(LSP0ERC725AccountCore, OwnableUnset)
+        onlyOwner
+    {
+        super.transferOwnership(_newOwner);
```
YamenMerhi, line 46: Same here

contracts/Utils/ClaimOwnership.sol
YamenMerhi, line 20: No need to explicitly list `override(OwnableUnset)` it can be just `override`

YamenMerhi (commented):
Great feature!

PR #163, opened 2022-05-13
fix!: prevent editing `LSP4TokenName` and `LSP4TokenSymbol` after deployment of Digital Assets

CJ42:

The Specification of [LSP4 - Digital Asset Metadata](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-4-DigitalAsset-Metadata.md) defines the following specification for the ERC725Y data keys `LSP4TokenName` and `LSP4TokenSymbol`

> **This MUST NOT be changeable, and set only during initialization of the token.**

Enforce this behaviour by creating an `LSP4DigitalAssetMetadataCore` contract that prevent from overriding these data keys after the digital asset contract has been deployed.

Screenshot 2022-05-13 at 08 21 56


contracts/LSP4DigitalAssetMetadata/LSP4DigitalAssetMetadata.sol
```diff
@@ -12,7 +13,7 @@ import "./LSP4Constants.sol";
  * @author Matthew Stevens
  * @dev Implementation of a LSP8 compliant contract.
  */
-abstract contract LSP4DigitalAssetMetadata is ERC725Y {
+abstract contract LSP4DigitalAssetMetadata is LSP4DigitalAssetMetadataCore, ERC725Y {
```
CJ42, line 16: @YamenMerhi what do you think should be the inheritance here?

```js
abstract contract LSP4DigitalAssetMetadata is LSP4DigitalAssetMetadataCore, ERC725Y {
    // ...
}
```

or 

```js
abstract contract LSP4DigitalAssetMetadata is ERC725Y, LSP4DigitalAssetMetadataCore {
```

CJ42 (commented):
@YamenMerhi added a question for the inheritance order

contracts/LSP4DigitalAssetMetadata/LSP4DigitalAssetMetadata.sol
YamenMerhi: I think it should be the last one in terms of the most base to most derived.
But I'm wondering if there is a need for the core version of the contract? Solidity makes overriding look complex, so what do you feel about removing the core version and overriding setData in `LSP4DigitalAssetMetadata `  and `LSP4DigitalAssetMetadataInitAbstract` with the logic inside `LSP4DigitalAssetMetadataCore `?

CJ42: Yes I am also wondering the same.
As I did it this way without the Core in the first place. See this commit: https://github.com/lukso-network/lsp-smart-contracts/pull/163/commits/d69b01f3aa83014ce90ab6b2c402fa8101e6b252

PR #168, opened 2022-05-21
refactor: simplify inheritance in LSP0 & LSP9

YamenMerhi:
- Simplifying inheritance of LSP0 and LSP9 by removing the inheritance of ERC725 and setting the owner manually in the constructor/initializer.
- Style functions in ERC order with adding few nastpec comments.

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -35,14 +35,14 @@ abstract contract LSP0ERC725AccountCore is
     ILSP1UniversalReceiver
 {
     /**
-     * @notice Emitted when a native token is received
+     * @notice Emitted when receiving native tokens
      * @param sender The address of the sender
      * @param value The amount of value sent
      */
     event ValueReceived(address indexed sender, uint256 indexed value);
 
     /**
-     * @dev Emits an event when a native token is received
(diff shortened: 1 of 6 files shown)


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -34,8 +34,16 @@ abstract contract LSP0ERC725AccountCore is
     IERC1271,
     ILSP1UniversalReceiver
 {
+    /**
+     * @notice Emitted when a native token is received
```
CJ42, line 38: Good to comment this event 🙂 

Maybe we could make it plural?
`Emitted when a native token is received` -> `Emitted when receiving native tokens`

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -34,8 +34,16 @@ abstract contract LSP0ERC725AccountCore is
...
+     * @param sender The address of the sender
+     * @param value The amount of value sent
+     */
     event ValueReceived(address indexed sender, uint256 indexed value);
 
+    /**
+     * @dev Emits an event when a native token is received
```
CJ42, line 45: same suggestion as above

CJ42 (approved):
Looks good to me

frozeman (approved):

PR #169, opened 2022-05-21
chore: add custom `Initializable` contract

YamenMerhi:
- Adding `Initializable` contract to **Custom** folder where this version has the booleans checks initialized and initializing public for making the implementation as generic as possible. One example would be to check these booleans by external contracts for security reasons when performing delegatecall.


YamenMerhi:
Will be imported from ERC725.
Introduced here: https://github.com/ERC725Alliance/ERC725/pull/111

PR #170, opened 2022-05-24
build: use JS for constants file in npm package

CJ42:

## Build

- [x] replace `constants.ts` > `constants.js`
- [x] replace interface IDs names in `constants.js` to use the ones from the [docs.lukso.tech > Standards > Contract Implementations > Interface IDs](https://docs.lukso.tech/standards/smart-contracts/interface-ids)

## Docs

add section in README to explain how to import the constants values from `constants.js`

- [x] add code snippet for import examples
- [x] add table that lists the available constants values.


package.json
```diff
@@ -27,7 +27,7 @@
     "test:lsp8": "NODE_NO_WARNINGS=1 jest LSP8",
     "test:lsp9": "NODE_NO_WARNINGS=1 jest LSP9",
     "test:universalfactory": "NODE_NO_WARNINGS=1 jest UniversalFactory",
-    "build": "npx hardhat compile",
+    "build": "npx hardhat compile && npx tsc constants.ts",
```
Hugoo, line 30: I think you can remove `npx` for hardhat and tsc as they are in the devDep

Hugoo:
Usually, the tsc output goes into a build / dist folder which is in the gitignore so these build files are not added in the repo.

Here the generated `constants.js` file is at the root and it is not ignored.

image



I would suggest to add it in the gitignore so it is not pushed by accident if someone makes a local build.

package.json
CJ42: @Hugoo removed, as I converted it to a plain JS file.

package.json
```diff
@@ -27,7 +27,7 @@
     "test:lsp8": "NODE_NO_WARNINGS=1 jest LSP8",
     "test:lsp9": "NODE_NO_WARNINGS=1 jest LSP9",
     "test:universalfactory": "NODE_NO_WARNINGS=1 jest UniversalFactory",
-    "build": "npx hardhat compile && npx tsc constants.ts",
+    "build": "npx hardhat compile",
     "package": "npx hardhat prepare-package",
```
Hugoo, line 31: here too

frozeman (approved):

PR #175, opened 2022-05-26
refactor: move custom implementation to Custom folder

YamenMerhi:
- Instead of adding custom keyword to the name of the contract to mark it as a custom implementation, move the custom implementations to the `Custom` folder.
- Adding comments for the purpose of the custom implementation. (Natspec to be added later)
- Move ERC165Checker and ClaimOwnership to `Custom` folder


contracts/Custom/ClaimOwnership.sol
```diff
@@ -1,6 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
+// This contract is a modified version of the OwnableUnset implementation, where we transfer Ownership as a 2 step
+// process, this allows to prevent for mistakes during ownership transfer, and so prevent a contract from potentially
```
CJ42, line 5: Suggested changes 

`and so prevent a contract from potentially being lost forever` -> `and so prevent control of a contract from potentially being lost forever`

contracts/Custom/ERC165Checker.sol
```diff
@@ -3,6 +3,10 @@
 
 pragma solidity ^0.8.0;
 
+// This contract is a modified version of OpenZeppelin implementation, where we modify the visibility of
```
CJ42, line 6: I would suggest adding the name of the contract from OZ in the comments: 


[...] is a modified version of ERC165Checker from OpenZeppelin

Maybe the comment could explain that the 3 x interface IDs are not checked to save gas?

CJ42 (commented):
added review comments

contracts/Custom/ERC165Checker.sol
YamenMerhi: The name of the contract is the same, but mentioned gas savings

CJ42 (approved):

PR #179, opened 2022-05-31
LSP8Enumerable Extension

lykhonis:
- Adds mappings from/to token-index
- Adds public function tokenAt(index) that returns a tokenId if exists
- totalSupply() returns total number of tokens. When used with tokenAt(index) can enable enumeration over all tokens that exist


contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
```diff
@@ -0,0 +1,51 @@
...
+    mapping(uint256 => bytes32) private _indexToken;
+
+    // Mapping from token id to index
+    mapping(bytes32 => uint256) private _tokenIndex;
+
+    /**
+     * @inheritdoc ILSP8Enumerable
+     */
+    function tokenAt(uint256 index) external view override returns (bytes32) {
+        require(index < totalSupply(), "Index is out of bounds");
```
DanielAbalde, line 25: I think it is convenient to remove this require because it will be very common (if not most cases) that this function is called from a loop iterating the whole collection. Developers can check if the result is null anyway.

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
```diff
@@ -0,0 +1,51 @@
...
+    // Mapping from token index to token id
+    mapping(uint256 => bytes32) private _indexToken;
+
+    // Mapping from token id to index
+    mapping(bytes32 => uint256) private _tokenIndex;
+
+    /**
+     * @inheritdoc ILSP8Enumerable
+     */
+    function tokenAt(uint256 index) external view override returns (bytes32) {
```
CJ42, line 24: Suggested changes

`external` -> `public`

Maybe `public` would fit better here? Since it is an extension and so custom LSP8Enumerable implementation can reuse the function inside other functions bodies?

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
```diff
@@ -0,0 +1,51 @@
...
+        return _indexToken[index];
+    }
+
+    function _beforeTokenTransfer(
+        address from,
+        address to,
+        bytes32 tokenId
+    ) internal virtual override(LSP8IdentifiableDigitalAssetCore) {
+        if (from == address(0)) {
+            uint256 index = totalSupply();
```
CJ42, line 35: You can use the state variable `_existingTokens` here, instead of the getter function `totalSupply()`.
This variable is marked as [`internal` in `LSP8IdentifiableDigitalAssetCore.sol`](https://github.com/lukso-network/lsp-smart-contracts/blob/99ea7a65c9a74e40b89158115dc33d446aee5633/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol#L37), and so can be reused in the child contract.

It is slightly cheaper (-20 gas).
```suggestion
            uint256 index = _existingTokens;
```

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
CJ42: I agree @DanielAbalde 👍

tests/LSP8IdentifiableDigitalAsset/extensions/LSP8Enumerable.behavior.ts
```diff
@@ -0,0 +1,80 @@
...
+  let context: LSP8EnumerableTestContext;
+
+  beforeAll(async () => {
+    context = await buildContext();
+  });
+
+  describe('when no minted tokens', () => {
+    it('should not get token', async () => {
+      const tokenSupply = await context.lsp8Enumerable.totalSupply();
+      await expect(context.lsp8Enumerable.tokenAt(tokenSupply)).toBeRevertedWith("Index is out of bounds")
```
CJ42, line 39: I agree with @DanielAbalde here. It should not revert but simply return `0` since no tokens have been minted.

tests/LSP8IdentifiableDigitalAsset/extensions/LSP8Enumerable.behavior.ts
```diff
@@ -0,0 +1,80 @@
...
+  lsp8Enumerable: LSP8EnumerableTester;
+  deployParams: LSP8EnumerableDeployParams;
+};
+
+export const shouldBehaveLikeLSP8Enumerable = (
+  buildContext: () => Promise<LSP8EnumerableTestContext>
+) => {
+  let context: LSP8EnumerableTestContext;
+
+  beforeAll(async () => {
```
CJ42, line 32: I would suggest replacing this `beforeAll()`, by a `beforeEach()`.

```suggestion
  beforeEach(async () => {
```

To avoid having to write at the end of each tests.

```ts
await context.lsp8Enumerable.burn(tokenId);
```

Unless there is a specific reason that I am not aware of? (like the test suite running faster, but since it is running in the CI, it should not be a big issue).

@YamenMerhi @jakeprins @Hugoo  what do you think about this? :)

CJ42 (commented):
Thanks a lot for this contribution @lykhonis 🙂  🎉 
I have added some review comments.

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
YamenMerhi: From a standard point of view, I think we should keep it.
We should not assume that this function will be used only off-chain when iterating on the whole collection, as a smart contract could be given an index and wants to check for the `tokenId` at this index.
If we remove the check it will return `bytes32(0)` for non-existing tokens which is still a valid `tokenId` if the type of `tokenId` was incremental. 

_Incremental:_
`0x0000000000000000000000000000000000000000000000000000000000000000`
`0x0000000000000000000000000000000000000000000000000000000000000001`
`0x0000000000000000000000000000000000000000000000000000000000000002`

What do you think ? @lykhonis @DanielAbalde @CJ42

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableInit.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity ^0.8.0;
+
+// modules
+import {LSP8EnumerableInitAbstract} from "./LSP8EnumerableInitAbstract.sol";
+
+// solhint-disable no-empty-blocks
+/**
+ * @dev LSP8 extension, adds access to a token id by an index.
+ */
+abstract contract LSP8EnumerableInit is LSP8EnumerableInitAbstract {
```
YamenMerhi, line 12: After discussion with @CJ42 , you can delete this contract

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
DanielAbalde: I don't see very clear in which use cases the cherry-picking would be by index rather than by tokenId directly, since you need to know what do you want to select.

Also, I assume looping this function will be used on-chain as well. For example, a contract that registers other NFT contracts, so that it can traverse the entire collection of tokens and owners from the index, to give permissions or rewards or logs or whatever.

But good point about the bytes32(0). I just searched the docs and I haven't seen anything about null keys. I think it should be part of the standard that bytes32(0) is a forbidden key, representing a null/empty key. Same for each valueType, its default value must be a forbidden key, to improve error resistance.

Incremental tokens can start with 1... `0x0000000000000000000000000000000000000000000000000000000000000001`, right? Starting from 0 only have true sense when it's understood as memory/processing jumps or some kind of offset, and here the thing is to represent an address or key, not to represent how far it is from the first possible value. <sup><sub>_Unless we want to live in a world where nothing leads to something. 🤯_</sup></sub>

lykhonis: Removed requirement as we already stated in docs it can return 0x0 if no token

lykhonis:
> Whats the gas cost increase with this extra index storing?

I used deployment cost of test with and without _indexToken, _tokenIndex mappings. Gas used is 2148857-2114573, the cost of having 2 mappings should be around 34284.

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8EnumerableCore.sol
CJ42: Let's keep it as it is, but I will suggest more tests in the future to compare calling from EOA (off-chain) _vs_ from a contract, and see what is the difference.

My guess is that:
- calling off-chain returns `0x`
- calling from a contract on-chain will return `0x0000000000000000000000000000000000000000000000000000000000000000`

This can be clarified in another PR. This is already a good start :)

CJ42 (approved):

CJ42:
@frozeman @YamenMerhi happy to merge this on my end 👍

lykhonis:
Will expand on unit tests before merge. Thx

lykhonis:
Updated doc to indicate return `0x00` instead of `0x` (no data). Return type is always bytes32, thus no data cannot be result

lykhonis:
Any update?

frozeman:
> > Whats the gas cost increase with this extra index storing?
> 
> I used deployment cost of test with and without _indexToken, _tokenIndex mappings. Gas used is [2148857-2114573](tel:2148857-2114573), the cost of having 2 mappings should be around 34284.

My question was more about single transfers  not deployment cost :)

CJ42:
Great job @lykhonis and super thanks too @DanielAbalde and @jakeprins for the extra pair of eyes on the PR.
Awesome community! 👌

PR #181, opened 2022-06-01
refactor: remove address typecast in URD-UP

YamenMerhi:
- Remove address typecast in URD-UP


CJ42 (approved):
Good catch!

PR #185, opened 2022-06-03
refactor: gas optimize LSP1UniversalReceiverDelegate

YamenMerhi:
- Refactor on `LSP5Utils.sol` lib used in both URD - UP and Vault:
     - Optimize typecasting variables
     - Remove un-needed **SLOAD**
     
Gas optimized: **~2500**

### DIFF
--- a/contracts/LSP5ReceivedAssets/LSP5Utils.sol
+++ b/contracts/LSP5ReceivedAssets/LSP5Utils.sol
@@ -59,14 +59,14 @@ library LSP5Utils {
         IERC725Y _account,
         bytes32 _arrayKey,
         bytes12 mapPrefix,
-        bytes32 _mapKeyToRemove
+        bytes32 _mapKeyToRemove,
+        bytes memory mapValue
     ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
-        uint64 index = extractIndexFromMap(_account, _mapKeyToRemove);
+        uint64 index = extractIndexFromMap(mapValue);
         bytes32 arrayKeyToRemove = LSP2Utils.generateArrayKeyAtIndex(_arrayKey, index);
 
-        bytes memory rawArrayLength = _account.getData(_arrayKey);
+        uint256 arrayLength = uint256(bytes32(_account.getData(_arrayKey)));
 
-        uint256 arrayLength = abi.decode(rawArrayLength, (uint256));
         uint256 newLength = arrayLength - 1;
 
         if (index == (arrayLength - 1)) {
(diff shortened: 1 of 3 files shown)


contracts/LSP5ReceivedAssets/LSP5Utils.sol
```diff
@@ -59,14 +59,14 @@ library LSP5Utils {
...
+        bytes memory mapValue
     ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
-        uint64 index = extractIndexFromMap(_account, _mapKeyToRemove);
+        uint64 index = extractIndexFromMap(mapValue);
         bytes32 arrayKeyToRemove = LSP2Utils.generateArrayKeyAtIndex(_arrayKey, index);
 
-        bytes memory rawArrayLength = _account.getData(_arrayKey);
+        uint256 arrayLength = uint256(bytes32(_account.getData(_arrayKey)));
 
-        uint256 arrayLength = abi.decode(rawArrayLength, (uint256));
```
CJ42, line 69: is it where the main gas cost occurs?

contracts/LSP5ReceivedAssets/LSP5Utils.sol
```diff
@@ -59,14 +59,14 @@ library LSP5Utils {
...
-        bytes32 _mapKeyToRemove
+        bytes32 _mapKeyToRemove,
+        bytes memory mapValue
     ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
-        uint64 index = extractIndexFromMap(_account, _mapKeyToRemove);
+        uint64 index = extractIndexFromMap(mapValue);
         bytes32 arrayKeyToRemove = LSP2Utils.generateArrayKeyAtIndex(_arrayKey, index);
 
-        bytes memory rawArrayLength = _account.getData(_arrayKey);
+        uint256 arrayLength = uint256(bytes32(_account.getData(_arrayKey)));
```
CJ42, line 68: Be aware that the explicit casting here result in zero padding on the right side.

So if in storage we have a 31 bytes -> `0x00000000000000000000000000000000000000000000000000000000000005`
It will add 2 x 0 bytes in the end to make a `bytes32` -> `0x0000000000000000000000000000000000000000000000000000000000000500`

which `0x0500` in hex = `1280` in decimal

Is it what we want?

@frozeman @YamenMerhi

CJ42 (commented):
Looks good. Added some discussion comments

contracts/LSP5ReceivedAssets/LSP5Utils.sol
YamenMerhi: Hmm, I see.
I will revert the changes, and I think the old behavior is better as if the arrayLength in bytes was wrongly set to != 32 bytes it will revert, so that's good.

contracts/LSP5ReceivedAssets/LSP5Utils.sol
YamenMerhi: No the gas optimization comes from the changes in `extractIndexFromMap` function

frozeman (approved):

PR #210, opened 2022-06-20
chore: add L16 endpoints to `hardhat.config.ts`

CJ42:
added to the config, so to be able to deploy + verify contracts directly from the CLI.


YamenMerhi (approved):
Cool

PR #215, opened 2022-06-23
test: resolve UniversalFactory failing tests

YamenMerhi:
- Resolve UniversalFactory failing tests.

**Todo**: Refactor UniversalFactory Tests (**Later**)

### DIFF
--- a/contracts/Helpers/FactoryTester.sol
+++ b/contracts/Helpers/FactoryTester.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+
+contract FactoryTester is Initializable {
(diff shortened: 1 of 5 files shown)


contracts/Helpers/FactoryTester.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+
+contract FactoryTester is Initializable {
```
CJ42, line 6: I would suggest to rename this contract to something like `ImplementationTester` or something along these lines.

`FactoryTester` is a bit misleading here, as the name of the contract suggest that the contract is an actual Factory contract that deploy other contracts while it's actually not.

CJ42 (commented):
added review comments

CJ42 (approved):

PR #218, opened 2022-06-28
feat: rename LSP8CompatibilityForERC721 to LSP8CompatibleERC721 + add mintable preset contracts

YamenMerhi:
- [X] Add a mintable version of `LSP8CompatibleERC721` called LSP7CompatibleERC721Mintable``
- [x] Rename contracts from `LSP8CompatibilityForERC721` to `LSP8CompatibleERC721`
> NB: ⚠️ only the extension part of this contract has been tested.


contracts/LSP8IdentifiableDigitalAsset/presets/LSP8CompatibleERC721Mintable.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {LSP8CompatibleERC721} from "../extensions/LSP8CompatibleERC721.sol";
+
+contract LSP8CompatibleER7210Mintable is LSP8CompatibleERC721 {
```
CJ42, line 6:
```suggestion
contract LSP8CompatibleER721Mintable is LSP8CompatibleERC721 {
```

hardhat.config.ts
```diff
@@ -113,8 +113,8 @@ const config: HardhatUserConfig = {
       "LSP4Compatibility",
       "LSP7CompatibilityForERC20",
       "LSP7CompatibilityForERC20Init",
-      "LSP8CompatibilityForERC721",
-      "LSP8CompatibilityForERC721Init",
+      "LSP8CompatibleERC721",
```
CJ42, line 116: We also need to add `LSP8CompatibleERC721Mintable` and `LSP8CompatibleERC721MintableInit` in the artifact list, so that a Dapp interface can have the `mint` function exposed to mint from the interface.

CJ42 (changes requested):

PR #219, opened 2022-07-01
feat: add payable on up  construct

Hugoo:
Todo:

- [ ] ajouter sur le init

Done with @CJ42


YamenMerhi (approved):
Could you mark the `initialize` function of `UniversalProfileInit` as payable also?

CJ42 (changes requested):
We also need to allow this feature in th eproxy version.
This require some tests + probably the `_initialize` function of `LSP0ERC725AccountInitAbstract` and `UniversalPofileInitAbstract` `payable` too, so that the `initialize(...) payable` function can pass the value down into the inheritance.

Below are the actual functions in the proxy that setup the UP and ERC725Account on deployment: 

https://github.com/lukso-network/lsp-smart-contracts/blob/53e33cc5359244cbc6f9f1db4ffc7c5b1348ad67/contracts/LSP0ERC725Account/LSP0ERC725AccountInitAbstract.sol#L15-L17

https://github.com/lukso-network/lsp-smart-contracts/blob/53e33cc5359244cbc6f9f1db4ffc7c5b1348ad67/contracts/UniversalProfileInitAbstract.sol#L20-L25

YamenMerhi:
@CJ42  `internal` and `private` functions cannot be payable.
It's not about passing the value to the internal functions (logic), it's about making the contract receive value by initializing with sending value along the way.

CJ42:
Ah thanks @YamenMerhi for the clarification. I did not know actually that `internal` and `private` cannot be `payable`.
So just need to add the tests for the contracts that are deployed behind proxy.

CJ42 (approved):

CJ42:
Thanks @Hugoo for this contribution! 🚀 😃 
@YamenMerhi let me know when it's good to go and be merged.

YamenMerhi (approved):
Super ! @Hugoo @CJ42

PR #222, opened 2022-07-04
refactor: remove ErrorHandler Lib

YamenMerhi:
- Remove ErrorHandler Lib and replace it with Address lib from OZ


CJ42 (commented):
Nice cleanup!

PR #239, opened 2022-07-27
fix: add `setApprovalForAll(...)` in `LSP8CompatibleERC721`

### LINKED ISSUE #225: LSP8 setApproveForAll
Currently LSP8 only allows approval of operators **per token** using:

```solidity
function authorizeOperator(address operator, bytes32 tokenId) public
```
https://docs.lukso.tech/standards/smart-contracts/lsp8-identifiable-digital-asset/#authorizeoperator

However, an option to **approve an operator for all tokens** is missing. A "setApproveForAll" equivalent should be taken into consideration.

In fact, [ILSP8CompatibleERC721](https://github.com/lukso-network/lsp-smart-contracts/blob/develop/contracts/LSP8IdentifiableDigitalAsset/extensions/ILSP8CompatibleERC721.sol) lacks the "setApprovalForAll" method that is present in the EIP721 (and EIP1155 too).
https://eips.ethereum.org/EIPS/eip-721
https://eips.ethereum.org/EIPS/eip-1155

Working on an NFT exchanger, where two people can exchange multiple NFTs using the contract as the intermediary, I wanted to support LSP8 but found this limitation. The UX of this contract is quite bad, the user has to approve the Swapper contract as an operator on every NFT contract they wants to trade, then register a swap and get its id, then both parties must deposit their NFTs into the Swapper contract and then one of the two can claim the other's NFTs. That's quite a few signatures needed for a fairly simple process, so the user experience (while maintaining the properties offered by this implementation) is quite tedious. So is even worse when the user does not have to approve the Swapper contract on each NFT contract as an operator, but has to do it for each token (even if these are put together in the same transaction, on the next swap they will have to repeat the approval for other (probably new) tokens ids). So a "setApprovalForAll" is quite convenient, in addition to complete compatibility with ERC721, as it means that the user only has to approve on the NFT contract once.

The [OpenZeppelin implementation of ERC721](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol) includes two mappings for approvals:
```solidity
  // Mapping from token ID to approved address
    mapping(uint256 => address) private _tokenApprovals;

    // Mapping from owner to operator approvals
    mapping(address => mapping(address => bool)) private _operatorApprovals;
  ```
However [LSP8](https://github.com/lukso-network/lsp-smart-contracts/blob/develop/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol) just include one:
```solidity

    // Mapping a `tokenId` to its authorized operator addresses.
    mapping(bytes32 => EnumerableSet.AddressSet) internal _operators;
```

So to expose an equivalent to "setApprovalForAll" seems to me to be too little cost for the positive impact of Lukso's NFTs being more compatible with the NFT platforms in the Ethereum ecosystem.

CJ42:

The backward compatible version of LSP8 with ERC721 is missing a function `setApprovalForAll(...)` that is part of the ERC721 interface.

This PR introduces this function in `LSP8CompatibleERC721` alongside a mapping `_operatorsApprovals` storing the operators address that are approved for all tokenIds owned by a `tokenOwner`.

> **NB:** gas cost of `transferFrom(...)` with `LSP8CompatibleERC721` is around 188k gas.

### DIFF
--- a/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721Core.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721Core.sol
@@ -27,6 +27,12 @@ abstract contract LSP8CompatibleERC721Core is
 {
     using EnumerableSet for EnumerableSet.AddressSet;
 
+    /**
+     * Mapping from owner to operator approvals
+     * @dev for backward compatibility with ERC721
+     */
+    mapping(address => mapping(address => bool)) private _operatorApprovals;
+
     /*
      * @inheritdoc ILSP8CompatibleERC721
      */
@@ -57,6 +63,10 @@ abstract contract LSP8CompatibleERC721Core is
         authorizeOperator(operator, bytes32(tokenId));
 
         emit Approval(tokenOwnerOf(bytes32(tokenId)), operator, tokenId);
+    }
+
+    function setApprovalForAll(address operator, bool approved) public virtual override {
+        _setApprovalForAll(msg.sender, operator, approved);
     }
 
     /**
@@ -92,11 +102,7 @@ abstract contract LSP8CompatibleERC721Core is
         override
         returns (bool)
     {
-        // silence compiler warning about unused variable
-        tokenOwner;
-        operator;
-
-        return false;
+        return _operatorApprovals[tokenOwner][operator];
     }
 
     /**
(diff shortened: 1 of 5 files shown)


skimaharvey (approved):

OnahProsperity (approved):

YamenMerhi (approved):

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721Core.sol
```diff
@@ -161,9 +170,15 @@ abstract contract LSP8CompatibleERC721Core is
...
     ) internal virtual override {
+        address operator = _msgSender();
+
+        if (!isApprovedForAll(from, operator) && !_isOperatorOrOwner(operator, tokenId)) {
+            revert LSP8NotTokenOperator(tokenId, operator);
+        }
+
         super._transfer(from, to, tokenId, force, data);
 
-        emit Transfer(from, to, abi.decode(abi.encodePacked(tokenId), (uint256)));
```
skimaharvey, line 166: I think we missed the one on line 153 ?

YamenMerhi: Can you fix it? It's showing in `authorizeOperator(..)`

YamenMerhi (approved):

frozeman (approved):

PR #240, opened 2022-07-28
Initiate Mocha and Migrate UP, LSP2, LSP4 Tests from Jest to Mocha

CJ42:

As discussed previously, we want to start moving all the tests from Jest to Chai, so to benefit from:

- better Waffle Chai matches (`@ethereum-waffle/jest` being deprecated ⚠️ https://www.npmjs.com/package/@ethereum-)waffle/jest
- include `hardhat-gas-reporter`
- include `solidity-coverage`
- better support for testing BigNumber (currently, we have to convert everything with Jest)
- ensure Typescript types errors are watched properly when running tests (discovered that Jest does not catch them at all).

This PR introduces this migration, by starting to re-factor the `expect` test assertions from Jest to Mocha/Chai for the following  LSPs

- [x] Universal Profile
- [x] LSP2
- [x] LSP4


YamenMerhi (approved):
Cool!

PR #250, opened 2022-08-12
fix: Hide linter warnings

b00ste:
Hiding the linter error using solhint-disable.

### DIFF
--- a/contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
+++ b/contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
@@ -26,5 +26,6 @@ contract UniversalReceiverTester {
         require(success, "low-level call to `universalReceiver(...)` function failed");
     }
 
+    // solhint-disable no-empty-blocks
     receive() external payable {}
 }
--- a/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
@@ -45,7 +45,7 @@ contract LSP1UniversalReceiverDelegateUP is
      */
     function universalReceiverDelegate(
         address sender,
-        uint256 value,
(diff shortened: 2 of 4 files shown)


YamenMerhi (approved):

contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
```diff
@@ -26,5 +26,6 @@ contract UniversalReceiverTester {
         require(success, "low-level call to `universalReceiver(...)` function failed");
     }
 
+    // solhint-disable no-empty-blocks
```
CJ42, line 29: @YamenMerhi @b00ste 

The `solc` compiler flags these variables. See screenshot below:

image

What should we do? Should we push them on the stack to silence the compiler warning?

This makes me think that we push these variables in the stack in `LSP8IdentifiableDigitalAssetCore` and `LSP8CompatibleERC721Core`. See these references in the code:

- https://github.com/lukso-network/lsp-smart-contracts/blob/712a6d98f23c4825e96be35310063e5a103e5fcb/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721Core.sol#L43-L44
- https://github.com/lukso-network/lsp-smart-contracts/blob/945c776276144e8afa0e386d023ba300bc1ebcef/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol#L380-L381

Not sure exactly about the implication, but to me it looks a bit "hacky". My idea would be to not push them on the stack (just save 3 gas), and just use the `solhint-disable` comments instead.

@skimaharvey also let me know what you think :)

CJ42 (commented):
added review comments

contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
YamenMerhi: As shown in the screenshot, we are using `// solhint disable ..` and these warnings are still showing, totally makes sense but IMO let's ignore and not push to the stack. Not because of the additional gas units but it doesn't make sense because we don't need them and it will look strange having variables sitting in functions without using them.

CJ42 (approved):

Issue #242, opened 2022-08-03
Gas Optimisation for LSP2


skimaharvey:
https://github.com/lukso-network/lsp-smart-contracts/blob/756e906546a7826fbe4c97f6721d0754a5c2e3c0/contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol#L26-L30

1 - Using multiple `require` cost less gas than using `&&` 
2 - Error message length is over 32 bytes

@YamenMerhi @CJ42 @bOOste let me know what you guys think

YamenMerhi:
Good points! You can create a PR for them 🚀 No need to shorten string errors as we'll switch to custom errors soon. :)

PR #263, opened 2022-08-18
refactor: replace >0 by !=0

skimaharvey:

`!=0` is a cheaper operation than `>0` and safe to use on `uint` with solidity `ˆ0.8.0` 

#242

### DIFF
--- a/contracts/Factories/UniversalFactory.sol
+++ b/contracts/Factories/UniversalFactory.sol
@@ -90,7 +90,7 @@ contract UniversalFactory {
         bytes32 generatedSalt = _generateSalt(initializeCallData, salt);
         contractCreated = Create2.deploy(msg.value, generatedSalt, byteCode);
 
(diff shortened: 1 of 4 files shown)


CJ42 (approved):
Nice thanks! Do you know the gas difference @skimaharvey ?

YamenMerhi (approved):

skimaharvey:
> Nice thanks! Do you know the gas difference @skimaharvey ?

@CJ42 I would guess 3 gas but I will double check when I have a min.
I think that when using `!=` you are using the `ISZERO` opcode which only needs 1 stack input vs less/greater than needs 2 stack inputs (the value and 0 in our cases) so I guess the difference is the cost of pushing to stack (3 gas).

CJ42:
Amazing thanks for this great explanation @skimaharvey ! 👌🏻

PR #264, opened 2022-08-18
fix: [QSP-6] fix incorrect index check for `_countTrailingZeroBytes(...)` in `LSP6KeyManagerCore.sol`

CJ42:

The former tests for **AllowedERC725YKeys** were not checking the right and expected behaviour when one of the allowed ERC725Y key in the list for the caller is `bytes32(0)`.

## 🐛 Fix

In `LSP6KeyManagerCore`, the function `_countTrailingZeroBytes(...)` does not check the byte at the index properly.

Refactored this function to ensure the function works properly when `key == `bytes32(0)`, while optimising for gas usage.

## 🧪 Tests

Add additional test with an allowed data key that contains one single byte (e.g: `0xaa00000000...00000000`), to check that looking for a specific `nByte` at a byte index inside the allowed data key works properly.


b00ste (approved):

YamenMerhi (approved):
Great refactoring!

PR #266, opened 2022-08-18
fix: [QSP-3] Operator Could Clear Operator List in LSP8

CJ42:

In LSP8, an operator can only be revoked by the token owner via the `revokeOperator(...)` function.

However, any operator can revoke another operator, and all the operators related to the `tokenId` by calling the function `transfer(...)` function with the same address for `from` and `to` (`from == to`).

If the `transfer(...)` function above is called with these arguments, the operator list is cleared and the owner will still have the NFT. The effect would be equivalent to that the operator successfully revokes all operators.

## What is the expected (new) behaviour?

Disallow this behaviour and unexpected edge case from occurring by disallowing specifying the same address for `from` and `to` in the `_transfer(...)` internal function parameter.


b00ste (approved):

skimaharvey (approved):

YamenMerhi (approved):
Cool

PR #269, opened 2022-08-21
refactor: [QSP-7] Use `bytes10` instead of `bytes12` for Mapping.. KeyTypes

YamenMerhi:
Using `bytes10` instead of `bytes12` in LSP2Utils functions and all function that uses LSP2 and making sure to have bytes2(0) as hardcoded function logic.

According to the documentation, data keys of type "Mapping" and "MappingWithGrouping" are formatted such that there are two empty bytes between the first ten bytes and the last twenty bytes. However, the versions of the functions generateMappingKey and generateMappingWithGroupingKey that use byte inputs take bytes12 for the first part of the key without ensuring that there are two empty bytes as per the specification.


CJ42 (approved):
@YamenMerhi This is an excellent fix. It now makes the library `LSP2Utils` more usable by anyone 👌

b00ste (approved):

PR #270, opened 2022-08-21
fix: [QSP-10] Revert when authorizing existing operators in LSP8

YamenMerhi:
- Adding a check for already authorized operators, to revert when users try to add the same address as operator.

Description: Users can call the function authorizeOperator(), multiple times with the same arguments. That will also cause the same event to be emitted multiple times. 

Existing tests testing the possibility to add the same operator were switched from allow behavior to revert behavior + removing useless lines of code (adding the same address as operator)

### DIFF
--- a/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
@@ -102,7 +102,8 @@ abstract contract LSP8IdentifiableDigitalAssetCore is ILSP8IdentifiableDigitalAs
             return;
         }
 
-        _operators[tokenId].add(operator);
+        bool isAdded = _operators[tokenId].add(operator);
+        if (!isAdded) revert LSP8OperatorAlreadyAuthorized(operator, tokenId);
 
         emit AuthorizedOperator(operator, tokenOwner, tokenId);
     }
--- a/tests/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.behaviour.ts
+++ b/tests/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.behaviour.ts
@@ -516,10 +516,6 @@ export const shouldBehaveLikeLSP8CompatibleERC721 = (
 
       describe("when many context.accounts have been approved for the tokenId", () => {
         it("should return the last new authorized operator", async () => {
-          // We approve the same account in the first and third approve call, with a different
-          // account in the second call as the last "new" approval.
-          // This is to highlight its not 100% the same behavior as ERC721 since that implementation
-          // has one active approval at a time, and LSP8 has a list of authorized operator addresses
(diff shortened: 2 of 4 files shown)


tests/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.behaviour.ts
```diff
@@ -531,10 +527,6 @@ export const shouldBehaveLikeLSP8CompatibleERC721 = (
             operatorSecondCall,
             tokenIdAsBytes32(mintedTokenId)
           );
-          await context.lsp8CompatibleERC721.approve(
```
CJ42, line 534: I think this line should not be removed from the test, as we are testing if the backward compatible for ERC721. So the operator address for the mintedTokenId should `operatorFirstAndThirdCall`.

If this test was failing in the past, then there was probably an error in the implementation.

contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
```diff
@@ -102,7 +102,8 @@ abstract contract LSP8IdentifiableDigitalAssetCore is ILSP8IdentifiableDigitalAs
             return;
         }
 
-        _operators[tokenId].add(operator);
+        bool isAdded = _operators[tokenId].add(operator);
+        if (!isAdded) revert LSP8OperatorAlreadyAuthorized(operator, tokenId);
```
CJ42, line 106: I am not sure if we should actually revert or not.
Maybe we could allow re-adding it, but not emit an event if the operator is already in the list.

The underlying `EnumerableSet._add(...)` will not add it a second time (duplicate) in here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/98c3a79b5765d58ef27856b8211c70a4907c63be/contracts/utils/structs/EnumerableSet.sol#L62-L72

To be discussed.

YamenMerhi: Sure, it will not add it a second time but its misleading for interfaces to check the event twice, that's why we want to revert.

tests/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.behaviour.ts
YamenMerhi: I don't think there is an error, as it's showing the last address approved, even when we did allow to authorize the same operator, it was showing the last different operator added (not the address duplicated). So now since we don't allow to add of the same operator, the backward compatible version will be working fine.

b00ste (approved):

contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
CJ42: make sense, agreed from yesterday discussion.

CJ42 (approved):

PR #276, opened 2022-08-23
fix: [QSP - Best Practices] refactor `deployCreate2Proxy(..)` to revert instead of refund caller

YamenMerhi:
- Revert when `(msg.value > 0) && (initiliazeCallData.length == 0)` instead of refunding the amount sent to the caller.

Check added at the end of the function to save gas when doing successful contract deployment instead of adding the check at the top.

- Edit tests to support the new behavior.

### DIFF
--- a/contracts/Factories/UniversalFactory.sol
+++ b/contracts/Factories/UniversalFactory.sol
@@ -134,7 +134,7 @@ contract UniversalFactory {
         } else {
             require(
                 msg.value == 0,
-                "UniversalFactory: Value cannot be sent to Proxies on deployment"
+                "UniversalFactory: value cannot be sent to the factory if initializeCallData is empty"
             );
         }
     }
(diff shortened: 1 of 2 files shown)


contracts/Factories/UniversalFactory.sol
```diff
@@ -132,15 +132,10 @@ contract UniversalFactory {
...
-            if (msg.value != 0) {
-                // solhint-disable avoid-low-level-calls
-                (bool success, bytes memory returnData) = payable(msg.sender).call{
-                    value: msg.value
-                }("");
-                Address.verifyCallResult(success, returnData, "UniversalFactory: Unknow Error");
-            }
+            require(
+                msg.value == 0,
+                "UniversalFactory: Value cannot be sent to Proxies on deployment"
```
CJ42, line 137:
```suggestion
                "UniversalFactory: value cannot be sent to the factory if initializeCallData is empty"
```

CJ42 (changes requested):

CJ42 (approved):

PR #277, opened 2022-08-23
docs [QSP-9]: Add comments about Allowance Double-Spend Exploit in LSP7 Natspec

YamenMerhi:
- Adding comments about Allowance Double-Spend Exploit in LSP7 Natspec in `authorizeOperator(..)` function.

This is just for warning, later actions to be discussed if we want to include `increaseAuthorizedAmount` function and `decreaseAuthorizedAmount` in implementation or/and specs.

### DIFF
--- a/contracts/LSP7DigitalAsset/ILSP7DigitalAsset.sol
+++ b/contracts/LSP7DigitalAsset/ILSP7DigitalAsset.sol
@@ -84,11 +84,15 @@ interface ILSP7DigitalAsset is IERC165, IERC725Y {
      * @param amount The amount of tokens operator has access to.
      * @dev Sets `amount` as the amount of tokens `operator` address has access to from callers tokens.
      *
-     * To increase or decrease the authorized amount of an operator, it's advised to call
-     * {revokeOperator} first, and then re-call {authorizeOperator} with the new amount
-     * to avoid front-running and Allowance Double-Spend Exploit
-     * Check more information here:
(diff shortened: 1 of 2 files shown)


contracts/LSP7DigitalAsset/ILSP7DigitalAsset.sol
```diff
@@ -84,6 +84,12 @@ interface ILSP7DigitalAsset is IERC165, IERC725Y {
      * @param amount The amount of tokens operator has access to.
      * @dev Sets `amount` as the amount of tokens `operator` address has access to from callers tokens.
      *
+     * To increase or decrease the authorized amount of an operator, it's advised to call
```
CJ42, line 87: I would suggest to:
- put the sentence about front-running and Allowance Double-Spent Exploit first at the top
(to make it more stand out) 
- + break with bullet points (if you think it reads better).

I have suggested the changes below ⬇️ 

```
     * To avoid front-running and Allowance Double-Spend Exploit when
     * increasing or decreasing the authorized amount of an operator, 
     * it is advised to:
     *     1. call {revokeOperator} first, and
     *     2. then re-call {authorizeOperator} with the new amount
     * 
     * for more information, see:
     * https://docs.google.com/document/d/1YLPtQxZu1UAvO9cZ1O2RPXBbT0mooh4DYKjA_jp-RLM/
     *
```

CJ42 (commented):
all good, just suggested one change

CJ42 (approved):

PR #280, opened 2022-08-24
fix: [QSP - Best Practices] Add ReentrancyGuard for Mintable token presets

YamenMerhi:
- Adding ReentrancyGuard for Mintable token presets.

Custom ReentrancyGuard was implemented as OZ implementation will cause conflicts, when upgrading to the new erc725-smart-contracts version the conflicts will be gone and then we can implement OZ implementation of ReentrancyGuard.

The `nonReentrant` modifier was added to the public `mint(..)` function and not the internal one because re-entrancy could serve as a feature, not a bug and it's the choice of the developer who will implement `_mint(..)` function in the extended contract to decide if he wants to allow re-entrancy or not.


contracts/LSP7DigitalAsset/presets/LSP7CompatibleERC20MintableInitAbstract.sol
```diff
@@ -21,7 +25,7 @@ contract LSP7CompatibleERC20MintableInitAbstract is LSP7CompatibleERC20InitAbstr
         uint256 amount,
         bool force,
         bytes memory data
-    ) public onlyOwner {
+    ) public onlyOwner nonReentrant {
```
CJ42, line 28: I think we can actually remove the `mint(...)` function from LSP7Mintable, since it is also present in the `LSP7MintableCore` contract.

So the nonRentrant modifier should go at the Core contract level.

https://github.com/lukso-network/lsp-smart-contracts/blob/fc130ce9cd8dd3f49a72a02b993a0d53ee8e8a2d/contracts/LSP7DigitalAsset/presets/LSP7MintableCore.sol#L18-L25

CJ42 (approved):

contracts/LSP7DigitalAsset/presets/LSP7CompatibleERC20MintableInitAbstract.sol
YamenMerhi: Nice suggestion 👍 ! As we discussed, will do the other way around by removing the Core contract instead of `mint(..)` in the 2 versions of the contracts. Same applies for LSP8

b00ste (approved):

PR #282, opened 2022-08-26
feat: [QSP-12] Add two step process for `renounceOwnership(...)` in `ClaimOwnership` contract

b00ste:

This PR introduces a 2-step process for renouncing ownership of an LSP0ERC725UniversalProfile and LSP9Vault.
The method `renounceOwnership()` was overridden to save the the `block.number` to `_lastBlock` and confirming `renounceOwnership()` only if `_lastBlock + 100` <= `block.number` < `_lastBlock + 200`.
A new constant was added, `_delayBlocks` which defines the number of blocks to be delayed.

Tests were added to check if the desired behaviour was achieved.

## Link to the [issue](https://github.com/lukso-network/LIPs/issues/115)

### DIFF
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -22,6 +22,8 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
      */
     address public override pendingOwner;
 
+    error RenounceOwnershipAvailableAtBlockNumber(uint256 blockNumber);
+
     function claimOwnership() public virtual override {
         _claimOwnership();
     }
@@ -55,10 +57,7 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
         uint256 _delayBlocks = 100;
 
         if (_lastBlock <= block.number && (_lastBlock + _delayBlocks) > block.number) {
-            revert RenounceOwnershipPending(
-                "OwnableClaim: Renounce ownership can be confirmed at block",
-                _lastBlock + _delayBlocks
-            );
+            revert RenounceOwnershipAvailableAtBlockNumber(_lastBlock + _delayBlocks);
         } else if (
             (_lastBlock + _delayBlocks) <= block.number &&
             (_lastBlock + _delayBlocks * 2) > block.number
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -11,6 +11,8 @@ import {IClaimOwnership} from "./IClaimOwnership.sol";
 // modules
 import {OwnableUnset} from "@erc725/smart-contracts/contracts/custom/OwnableUnset.sol";
 
+    error RenounceOwnershipAvailableAtBlockNumber(uint256 blockNumber);
+
 abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
     /**
      * @dev The block number saved in the first step for
@@ -28,8 +30,6 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
      * @dev The address that may use `claimOwnership()`
      */
     address public override pendingOwner;
-
-    error RenounceOwnershipAvailableAtBlockNumber(uint256 blockNumber);
 
     function claimOwnership() public virtual override {
         _claimOwnership();
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -17,6 +17,13 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
      * renouncing ownership of the contract
      */
     uint256 private _lastBlock;
+
+    /**
+     * @dev The number of blocks needed to pass for successfully
+     * confirming `renounceOwnership()`
+     */
+    uint256 constant _delayBlocks = 100;
+
     /**
      * @dev The address that may use `claimOwnership()`
      */
@@ -51,11 +58,8 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
      * is more than 200 block back.
      * Execute `renounceOwnership` if the `_lastRenounceOwnershipBlock`
      * is less than 200 blocks back and more than 100 blocks.
-     *
      */
     function _renounceOwnership() internal virtual {
-        uint256 _delayBlocks = 100;
-
         if (_lastBlock <= block.number && (_lastBlock + _delayBlocks) > block.number) {
             revert RenounceOwnershipAvailableAtBlockNumber(_lastBlock + _delayBlocks);
         } else if (
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -16,7 +16,7 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
      * @dev The block number saved in the first step for
      * renouncing ownership of the contract
      */
-    uint256 private _lastRenounceOwnershipBlock;
+    uint256 private _lastBlock;
     /**
      * @dev The address that may use `claimOwnership()`
      */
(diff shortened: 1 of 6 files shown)


CJ42 (changes requested):
I thought we said that we would put these functions in the `ClaimOwnership` contract?

Also, do we introduce an extra function `confirmRenounceOwnership(...)` as well? If this is the case and it is in `ClaimOwnership` contract, doesn't that change the interface IDs in some ways?

@YamenMerhi @frozeman

contracts/Custom/IClaimOwnership.sol
```diff
@@ -4,6 +4,10 @@ pragma solidity ^0.8.0;
 bytes4 constant _INTERFACEID_CLAIM_OWNERSHIP = 0xd225f160;
 
 interface IClaimOwnership {
+    event RenounceOwnershipInitiated();
+
+    error RenounceOwnershipPending(string errorText, uint256 blockNumber);
```
YamenMerhi, line 9: I don't think this should go in the interface. If agreed please let's remove

tests/UniversalProfile.behaviour.ts
```diff
@@ -245,6 +245,83 @@ export const shouldBehaveLikeLSP3 = (
...
+
+      const renounceOwnershipSecond = context.universalProfile
+        .connect(context.accounts[0])
+        .renounceOwnership();
+
+      await expect(renounceOwnershipSecond)
+        .to.emit(context.universalProfile, "OwnershipTransferred")
+        .withArgs(
+          context.accounts[0].address,
+          ethers.utils.hexZeroPad("0x", 20)
```
YamenMerhi, line 317: You can replace with `ethers.constants.AddressZero`

tests/UniversalProfile.behaviour.ts
```diff
@@ -245,6 +245,83 @@ export const shouldBehaveLikeLSP3 = (
...
+        );
+
+      expect(await context.universalProfile.owner()).to.equal(
+        ethers.utils.hexZeroPad("0x", 20)
```
YamenMerhi, line 321: same here

tests/LSP9Vault/LSP9Vault.behaviour.ts
```diff
@@ -221,6 +221,83 @@ export const shouldBehaveLikeLSP9 = (
...
+
+        await expect(renounceOwnershipSecond)
+          .to.emit(context.lsp9Vault, "OwnershipTransferred")
+          .withArgs(
+            context.accounts.owner.address,
+            ethers.utils.hexZeroPad("0x", 20)
+          );
+
+        expect(await context.lsp9Vault.owner()).to.equal(
+          ethers.utils.hexZeroPad("0x", 20)
```
YamenMerhi, line 297: same here

tests/LSP9Vault/LSP9Vault.behaviour.ts
```diff
@@ -221,6 +221,83 @@ export const shouldBehaveLikeLSP9 = (
...
+
+        const renounceOwnershipSecond = context.lsp9Vault
+          .connect(context.accounts.owner)
+          .renounceOwnership();
+
+        await expect(renounceOwnershipSecond)
+          .to.emit(context.lsp9Vault, "OwnershipTransferred")
+          .withArgs(
+            context.accounts.owner.address,
+            ethers.utils.hexZeroPad("0x", 20)
```
YamenMerhi, line 293: and here

YamenMerhi (changes requested):

contracts/Custom/ClaimOwnership.sol
```diff
@@ -12,8 +12,18 @@ import {IClaimOwnership} from "./IClaimOwnership.sol";
 import {OwnableUnset} from "@erc725/smart-contracts/contracts/custom/OwnableUnset.sol";
 
 abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
+    /**
+     * @dev The block number saved in the first step for
+     * renouncing ownership of the contract
+     */
+    uint256 private _lastBlock;
```
CJ42, line 19: could you add an extra space between the variable and the natspec comment.

contracts/Custom/ClaimOwnership.sol
```diff
@@ -12,8 +12,18 @@ import {IClaimOwnership} from "./IClaimOwnership.sol";
...
+    /**
+     * @dev The address that may use `claimOwnership()`
+     */
     address public override pendingOwner;
 
+    error RenounceOwnershipAvailableAtBlockNumber(uint256 blockNumber);
```
CJ42, line 25: It might be better to put the custom error above the contract IMO.

Like this.

```solidity
pragma solidity ^0.8.5;

import {OwnableUnset} from "@erc725/smart-contracts/contracts/custom/OwnableUnset.sol";

error RenounceOwnershipAvailableAtBlockNumber(uint256 blockNumber);

abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
    // ...
}
```

contracts/Custom/ClaimOwnership.sol
```diff
@@ -31,4 +45,28 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
...
+
+    /**
+     * @dev Save the block number for the first step if `_lastRenounceOwnershipBlock`
+     * is more than 200 block back.
+     * Execute `renounceOwnership` if the `_lastRenounceOwnershipBlock`
+     * is less than 200 blocks back and more than 100 blocks.
+     *
+     */
+    function _renounceOwnership() internal virtual {
+        uint256 _delayBlocks = 100;
```
CJ42, line 57: Maybe we could put that as a `constant` inside the contract. So that it is stored in the contract bytecode, to make it a bit cheaper.

tests/LSP9Vault/LSP9Vault.behaviour.ts
```diff
@@ -221,6 +221,80 @@ export const shouldBehaveLikeLSP9 = (
...
+      before(async () => {
+        await network.provider.send("hardhat_mine", ["0xffff"]);
+      });
+
+      beforeEach(async () => {
+        context = await buildContext();
+      });
+
+      it("should fail to confirm if delay didn't expire", async () => {
+        const renounceOwnershipOnce = context.lsp9Vault
```
CJ42, line 235: You have to `await` here the transaction in order to wait for the transaction receipt.

```suggestion
        const renounceOwnershipOnce = await context.lsp9Vault
```

tests/LSP9Vault/LSP9Vault.behaviour.ts
```diff
@@ -221,6 +221,80 @@ export const shouldBehaveLikeLSP9 = (
...
+          .connect(context.accounts.owner)
+          .renounceOwnership();
+
+        await expect(renounceOwnershipOnce).to.emit(
```
CJ42, line 239: and here remove the `await`

tests/LSP9Vault/LSP9Vault.behaviour.ts
```diff
@@ -221,6 +221,80 @@ export const shouldBehaveLikeLSP9 = (
...
+          "RenounceOwnershipInitiated"
+        );
+
+        expect(await context.lsp9Vault.owner()).to.equal(
+          context.accounts.owner.address
+        );
+
+        await network.provider.send("hardhat_mine", ["0x62"]); // skip 98 blocks
+
+        const renounceOwnershipSecond = context.lsp9Vault
```
CJ42, line 250: same comment as above.

tests/LSP9Vault/LSP9Vault.behaviour.ts
```diff
@@ -221,6 +221,80 @@ export const shouldBehaveLikeLSP9 = (
...
+          .connect(context.accounts.owner)
+          .renounceOwnership();
+
+        await expect(renounceOwnershipSecond)
+          .to.be.revertedWithCustomError(
+            context.lsp9Vault,
+            "RenounceOwnershipAvailableAtBlockNumber"
+          )
+          .withArgs((await renounceOwnershipOnce).blockNumber + 100);
```
CJ42, line 259: so here in the end you can remove the `await` in this line :)

CJ42 (changes requested):
I think it's ok for now, but should also tests the more cases, where for instance the block delay timed out (for the future).

tests/LSP9Vault/LSP9Vault.behaviour.ts
b00ste: if i make the suggested changes the test passes with wrong event name

CJ42 (approved):

YamenMerhi (approved):

PR #288, opened 2022-08-30
fix: Add Checks to ensure contract cannot be self owned with ClaimOwnership 

YamenMerhi:
- Introduce a check in ClaimOwnership contract, in the `transferOwnership(..)` to avoid setting the `address(this)` as _pendingOwner_.
- Adjust the contract name in other revert string

### DIFF
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -49,7 +49,7 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
     }
 
     function _claimOwnership() internal virtual {
-        require(msg.sender == pendingOwner, "OwnableClaim: caller is not the pendingOwner");
+        require(msg.sender == pendingOwner, "ClaimOwnership: caller is not the pendingOwner");
         _setOwner(pendingOwner);
         pendingOwner = address(0);
     }
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -13,6 +13,11 @@ import {OwnableUnset} from "@erc725/smart-contracts/contracts/custom/OwnableUnse
 
 error RenounceOwnershipAvailableAtBlockNumber(uint256 blockNumber);
 
+/**
+ * @dev reverts when trying to transfer ownership to the address(this)
+ */
+error CannotSelfTransferOwnership();
+
 abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
     /**
      * @dev The block number saved in the first step for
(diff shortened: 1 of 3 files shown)


b00ste (approved):

contracts/Custom/ClaimOwnership.sol
```diff
@@ -44,12 +49,13 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
     }
 
     function _claimOwnership() internal virtual {
-        require(msg.sender == pendingOwner, "OwnableClaim: caller is not the pendingOwner");
+        require(msg.sender == pendingOwner, "ClaimOwnership: caller is not the pendingOwner");
```
CJ42, line 52: nice adjustment 👌

contracts/Custom/ClaimOwnership.sol
```diff
@@ -44,12 +49,13 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
...
         _setOwner(pendingOwner);
         pendingOwner = address(0);
     }
 
     function _transferOwnership(address newOwner) internal virtual {
+        if (newOwner == address(this)) revert CannotSelfTransferOwnership();
```
CJ42, line 58: _Suggested changes_
`CannotSelfTransferOwnership` --> `CannotTransferOwnershipToSelf`

tests/LSP6KeyManager/tests/PermissionChangeOwner.test.ts
```diff
@@ -285,7 +285,7 @@ export const shouldBehaveLikePermissionChangeOwner = (
 
       await expect(
         notPendingKeyManager.connect(context.owner).execute(payload)
-      ).to.be.revertedWith("OwnableClaim: caller is not the pendingOwner");
+      ).to.be.revertedWith("ClaimOwnership: caller is not the pendingOwner");
```
CJ42, line 288: Maybe it would be good to add an extra test in the Key Manager, to test that if the payload `transferOwnership(address(this))` is passed on the Key Manager, it will be matched at the `ClaimOwnership` contract level?

CJ42 (changes requested):

CJ42 (approved):

CJ42:
@YamenMerhi 

Could you also add the custom error definitions for `ClaimOwnership ` inside `constants.js`, so that it can be used by the browser extension, please :)

Either in this PR or a separate one.

YamenMerhi:
@CJ42 Let's do it in a seperate PR, so I can work on introducing other custom errors that we introduced in the past, as well for event signature.

PR #294, opened 2022-08-31
build: add code coverage 

CJ42:


.github/workflows/build-lint-test.yml
```diff
@@ -87,3 +87,12 @@ jobs:
 
       - name: run tests
         run: npm run test:${{ matrix.lsp }}
+
+      # Trigger Coverage
+      - name: Trigger Code Coverage Workflow
+        if: github.event_name == 'push'
```
CallumGrindle, line 93: I think we don't need this conditional, this workflow will only run on pull requests so this wont ever get triggered

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      # This will also generate the Typechain types used by the Chai tests
+      - name: Install dependencies
+        if: steps.build-cache.outputs.cache-hit != 'true'
+        run: npm ci
+
+      - name: Run tests with coverage
+        run: npm run test:coverage
+
+      - name: Upload to Coveralls
```
CallumGrindle, line 39: lgtm, lets see how it works when testing it

.github/workflows/coverage.yml
```diff
@@ -1,10 +1,13 @@
...
 
-name: Test + Generate coverage report
+name: Solidity Code coverage
 
 on:
-  workflow_dispatch:
-    types: [trigger-coverage]
+  push:
+    branches:
+      - "develop"
```
Hugoo, line 10: can you use checkout v3 ?

Hugoo (approved):

.github/workflows/coverage.yml
CJ42: done

Hugoo (approved):
nice!

PR #303, opened 2022-09-05
refactor: disallow DELEGATECALL operation in LSP9Vault

YamenMerhi:
- Overriding execute function in LSP9Vault and disallowing the use of delegatecall, since it's a dangerous operation and could lead to the destruction of the vault.

The use of delegatecall in LSP9 is totally optional. 
Referring to the [LSP9 specs](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-9-Vault.md), it's mention that an LSP9Vault must include the following operation: CALL, CREATE, CREATE2, STATICCALL and CAN include the operation DELEGATECALL.

### DIFF
--- a/contracts/LSP9Vault/LSP9VaultCore.sol
+++ b/contracts/LSP9Vault/LSP9VaultCore.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 // interfaces
 import {IERC725Y} from "@erc725/smart-contracts/contracts/interfaces/IERC725Y.sol";
+import {IERC725X} from "@erc725/smart-contracts/contracts/interfaces/IERC725X.sol";
 import {ILSP1UniversalReceiver} from "../LSP1UniversalReceiver/ILSP1UniversalReceiver.sol";
 import {
     ILSP1UniversalReceiverDelegate
@@ -21,6 +22,8 @@ import {ClaimOwnership} from "../Custom/ClaimOwnership.sol";
 
 // constants
 import {_INTERFACEID_CLAIM_OWNERSHIP} from "../Custom/IClaimOwnership.sol";
+
+import "@erc725/smart-contracts/contracts/constants.sol";
 import {
     _INTERFACEID_LSP1,
     _INTERFACEID_LSP1_DELEGATE,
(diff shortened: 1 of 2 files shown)


b00ste (approved):

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 // interfaces
 import {IERC725Y} from "@erc725/smart-contracts/contracts/interfaces/IERC725Y.sol";
```
CJ42, line 5: You can remove these two lines from the imports (` import {IERC725X} ...` and `import {IERC725Y} ...`). They are not used in the contract anywhere (neither in the inheritance, nor inside the contract code.

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -21,6 +22,8 @@ import {ClaimOwnership} from "../Custom/ClaimOwnership.sol";
 
 // constants
 import {_INTERFACEID_CLAIM_OWNERSHIP} from "../Custom/IClaimOwnership.sol";
+
+import "@erc725/smart-contracts/contracts/constants.sol";
```
CJ42, line 26: Why not adding the exact constants imported here, so they can be easily found?

```solidity
import {
    OPERATION_CALL, 
    OPERATION_STATICCALL, 
    OPERATION_CREATE, 
    OPERATION_CREATE2
} from "@erc725/smart-contracts/contracts/constants.sol";
```

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -74,61 +77,40 @@ contract LSP9VaultCore is ERC725XCore, ERC725YCore, ClaimOwnership, ILSP1Univers
...
     }
 
-    // ERC165
+    // ERC725
 
     /**
-     * @dev See {IERC165-supportsInterface}.
+     * @inheritdoc IERC725X
+     * @param operation The operation to execute: CALL = 0 CREATE = 1 CREATE2 = 2 STATICCALL = 3
+     * @dev Executes any other smart contract.
```
CJ42, line 85: would it be worth mentioning in the Natspec docs as well that DELEGATECALL is disallowed?

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -208,7 +243,7 @@ contract LSP9VaultCore is ERC725XCore, ERC725YCore, ClaimOwnership, ILSP1Univers
     }
 
     /**
-     * @dev Calls the universalReceiver function of the sender if supports LSP1 InterfaceId
+     * @dev Calls the universalReceiver function of the vault sender if supports LSP1 InterfaceId
```
CJ42, line 246: nice extra specs 👍

contracts/LSP9Vault/LSP9VaultCore.sol
CJ42, line 26: Why not adding the exact constants imported here, so they can be easily found?

```solidity
import {
    OPERATION_CALL, 
    OPERATION_STATICCALL, 
    OPERATION_CREATE, 
    OPERATION_CREATE2
} from "@erc725/smart-contracts/contracts/constants.sol";
```

CJ42 (changes requested):

contracts/LSP9Vault/LSP9VaultCore.sol
YamenMerhi: They are used within the natspec

contracts/LSP9Vault/LSP9VaultCore.sol
YamenMerhi: Good pt

contracts/LSP9Vault/LSP9VaultCore.sol
YamenMerhi: I overrided the @param tag to point out that these are the operation allowed. I guess this is a good base

PR #308, opened 2022-09-12
test: Add tests for a LSP6 controlled Mintable LSP7

b00ste:

In this pr we create a new test file that tests the behavior of a LSP6 controlled Mintable LSP7.

Methods:

- using execute(..) in LSP7 through LSP6 --> reverts
- using mint(..) in LSP7 through LSP6 --> reverts
- using renounceOwnership(..) in LSP7 through LSP6 --> reverts
- using transferOwnership(..) in LSP7 through LSP6 --> works
- using setData(..) in lSP7 through LSP6 --> works

Permissions:

- CHANGEOWNER --> works
- CHANGEPERMISSIONS --> works 
- ADDPERMISSIONS --> works
- SETDATA --> works
- CALL --> not working
- STATICCALL --> not working
- DELEGATECALL --> not working
- DEPLOY --> not working
- TRANSFERVALUE --> not working
- SIGN --> works
- ENCRYPT --> not tested


tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+  );
+
+  const keyManager = await new LSP6KeyManager__factory(accounts[0]).deploy(
+    lsp7.address
+  );
+
+  const keys = [
+    ERC725YKeys.LSP6["AddressPermissions[]"].length,
+    ERC725YKeys.LSP6["AddressPermissions[]"].index + ARRAY_INDEX.ZERO,
+    ERC725YKeys.LSP6["AddressPermissions:Permissions"] +
```
CJ42, line 48: maybe you can store the `address[0]` in a `owner` variable (for the address that has ALL PERMISSIONS)?

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+
+describe("When deploying LSP7 with LSP6 as owner", () => {
+  let context: LSP6ControlledToken;
+
+  before(async () => {
+    context = await buildContext();
+  });
+
+  it("should have lsp6 as owner of the lsp7", async () => {
+    expect(await context.token.owner()).to.be.equal(context.keyManager.address);
```
CJ42, line 96:
```suggestion
    expect(await context.token.owner()).to.equal(context.keyManager.address);
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        context.accounts[0].address.substring(2),
+    ];
+
+    const values = [
+      ARRAY_LENGTH.ONE,
+      context.accounts[0].address,
+      ALL_PERMISSIONS,
+    ];
+
+    expect(await context.token["getData(bytes32[])"](keys)).to.be.deep.equal(
```
CJ42, line 113:
```suggestion
    expect(await context.token["getData(bytes32[])"](keys)).to.deep.equal(
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+      values
+    );
+  });
+
+  describe("using execute(..) in LSP7 through LSP6", () => {
```
CJ42, line 118:
```suggestion
  describe("when trying to call execute(..) function on LSP7 through LSP6", () => {
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+    it("should revert", async () => {
```
CJ42, line 119: Maybe we could add a bit more descriptions in the tests, so that we know why it should revert?

```suggestion
    it("should revert because function does not exist on LSP7", async () => {
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+      const newTokenContract = await new LSP7Tester__factory(
+        context.accounts[0]
+      ).deploy("NewTokenName", "NewTokenSymbol", context.accounts[0].address);
+      const mintPayload = newTokenContract.interface.encodeFunctionData(
```
CJ42, line 123: How about adding a comment above, to explain that we try to call the `mint(...)` function in another contract? (as we were discussing)

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        .connect(context.accounts[0])
+        .execute(payload);
+
+      await expect(executePayload).to.be.revertedWith(
+        "LSP6: Unknown Error occured when calling the linked target contract"
+      );
+    });
+  });
+
+  describe("using mint(..) in LSP7 through LSP6", () => {
```
CJ42, line 144: same comment as above.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        1,
+        true,
+        "0x",
+      ]);
+
+      const executeMintPayload = context.keyManager
+        .connect(context.accounts[0])
+        .execute(mintPayload);
+
+      await expect(executeMintPayload)
```
CJ42, line 158: I would suggest to put the whole call inside the `expect(...)`, as the tests sometimes are buggy when testing events like that.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        .to.be.revertedWithCustomError(
+          context.keyManager,
+          "InvalidERC725Function"
+        )
+        .withArgs(renounceOwnershipPayload);
+    });
+  });
+
+  describe("using transferOwnership(..) in LSP7 through LSP6", () => {
+    it("should change the owner of LSP7 contract", async () => {
```
CJ42, line 186: We could add some extra tests here, to ensure that afterwards, to ensure that:

- the functions marked as `onlyOwner` on the LSP7 token contract cannot be called by the KeyManager anymore afterwards.
- only the new owner (the EOA) can call these functions.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+      const transferOwnershipPayload =
+        context.token.interface.encodeFunctionData("transferOwnership", [
+          context.accounts[0].address,
```
CJ42, line 189: It is a bit misleading here to use the same address as the EOA that had all the permissions previously (`accounts[0]`).

Maybe we could use an other account to make it more explicit? Like:

```js
const newOwner = context.accounts[1];

const transferOwnershipPayload =
        context.token.interface.encodeFunctionData("transferOwnership", [
          newOwner.address,
        ]);

```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        ]);
+
+      await context.keyManager
+        .connect(context.accounts[0])
+        .execute(transferOwnershipPayload);
+
+      expect(await context.token.owner()).to.equal(context.accounts[0].address);
+    });
+
+    it("should revert when calling setData(..)", async () => {
```
CJ42, line 199: ah ok this is done here. Maybe we can improve the description.

Or use a `describe` block?

```js
describe("when previous owner try to interact via the Key Manager", () => {
    it("`setData(...)` -> should revert with 'caller is not the owner' error.", async () => {
        // ....
    }) 
})
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          ARRAY_LENGTH.TWO,
+          context.accounts[0].address,
+          context.accounts[1].address,
+          ALL_PERMISSIONS,
+          PERMISSIONS.CHANGEOWNER,
+        ];
+
+        expect(
+          await context.token["getData(bytes32[])"](keys)
+        ).to.be.deep.equal(values);
```
CJ42, line 310:
```suggestion
        ).to.deep.equal(values);
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        await context.keyManager
+          .connect(context.accounts[1])
+          .execute(transferOwnershipPayload);
+
+        expect(await context.token.owner()).to.equal(
+          context.accounts[1].address
+        );
+      });
+
+      after(async () => {
```
CJ42, line 359: very nice 👌

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          ALL_PERMISSIONS,
+          PERMISSIONS.CHANGEOWNER,
+        ];
+
+        expect(
+          await context.token["getData(bytes32[])"](keys)
+        ).to.be.deep.equal(values);
+      });
+
+      it("should revert when trying to use `renounceOwnership()`", async () => {
```
CJ42, line 313: maybe we should not put this test here, as it is not related to the permission CHANGEOWNER.
Because renounceOwnership cannot be called via the KeyManager (the KM does not provide this path).

So maybe in a separate describe block for edge cases?

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          context.accounts[1].address,
+          context.accounts[2].address,
+          ALL_PERMISSIONS,
+          PERMISSIONS.CHANGEOWNER,
+          PERMISSIONS.CHANGEPERMISSIONS,
+        ];
+
+        expect(
+          await context.token["getData(bytes32[])"](keys)
+        ).to.be.deep.equal(values);
```
CJ42, line 403:
```suggestion
        ).to.deep.equal(values);
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          ERC725YKeys.LSP6["AddressPermissions:Permissions"] +
+          context.accounts[0].address.substring(2);
+        const value = PERMISSIONS.CALL;
+        const payload = context.token.interface.encodeFunctionData(
+          "setData(bytes32,bytes)",
+          [key, value]
+        );
+
+        const changePermission = context.keyManager
+          .connect(context.accounts[1])
```
CJ42, line 417: maybe we can define this signer as `addressCanChangePermission`?
Otherwise, it is difficult to infer which permissions the caller has by reading the test.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          "setData(bytes32,bytes)",
+          [key, value]
+        );
+
+        await context.keyManager.connect(context.accounts[2]).execute(payload);
+
+        expect(await context.token["getData(bytes32)"](key)).to.equal(value);
+      });
+
+      it("should add back ALL_PERMISSIONS of the address", async () => {
```
CJ42, line 440: This is actually the wrong behaviour that we will have to fix.
It should be possible if you have CHANGEPERMISSIONS permission to add more permissions that it is initially.

See this comment: https://github.com/lukso-network/LIPs/pull/102#issuecomment-1205535756

As a note. Will have to update in the future.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          PERMISSIONS.CHANGEPERMISSIONS,
+          PERMISSIONS.ADDPERMISSIONS,
+        ];
+
+        expect(
+          await context.token["getData(bytes32[])"](keys)
+        ).to.be.deep.equal(values);
+      });
+
+      it("should be allowed to add permissions", async () => {
```
CJ42, line 501: I suggest this wording, to make it more clear.

```suggestion
      it("should be allowed to add a new controller address with some permissions", async () => {
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          "setData(bytes32,bytes)",
+          [key, value]
+        );
+
+        await context.keyManager.connect(context.accounts[3]).execute(payload);
+
+        expect(await context.token["getData(bytes32)"](key)).to.equal(value);
+      });
+
+      it("should revert when trying to remove permissions", async () => {
```
CJ42, line 516: same here.

```suggestion
      it("should not be authorized to edit permissions of an existing controller (e.g: removing permissions)", async () => {
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        const payload = context.token.interface.encodeFunctionData(
+          "setData(bytes32,bytes)",
+          [key, value]
+        );
+
+        const executePayload = context.keyManager
+          .connect(context.accounts[3])
+          .execute(payload);
+
+        await expect(executePayload)
```
CJ42, line 530: Maybe we can put the contract call direct inside the `expect` here?
Since we don't use the `const` defined.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+          ALL_PERMISSIONS,
+          PERMISSIONS.CHANGEOWNER,
+          PERMISSIONS.CHANGEPERMISSIONS,
+          PERMISSIONS.ADDPERMISSIONS,
+          PERMISSIONS.SETDATA,
+        ];
+
+        expect(
+          await context.token["getData(bytes32[])"](keys)
+        ).to.be.deep.equal(values);
```
CJ42, line 596: Can you check that the other tests don't have the `be` like here, but just `to.deep.equal(...)`

```suggestion
        ).to.deep.equal(values);
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+      });
+
+      it("should allow second controller to use setData", async () => {
```
CJ42, line 599: nice test 👌

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+      it("should restrict second controller with AllowedERC725YKeys", async () => {
+        const key =
+          ERC725YKeys.LSP6["AddressPermissions:AllowedERC725YKeys"] +
+          context.accounts[4].address.substring(2);
+        const value = abiCoder.encode(
+          ["bytes32[]"],
+          [
+            [
+              ethers.utils
+                .keccak256(ethers.utils.toUtf8Bytes("FirstRandomString"))
```
CJ42, line 625: You can define this variable in a constant at the beginning of the `describe` block and reuse it across tests (define the data key prefix in a constant for instance).

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+        await context.keyManager.connect(context.accounts[4]).execute(payload);
+
+        expect(
+          await context.token["getData(bytes32[])"](keys)
+        ).to.be.deep.equal(values);
+      });
+
+      it("should revert when trying to change a key that is not allowed", async () => {
+        const keys = [
+          ethers.utils.keccak256(ethers.utils.toUtf8Bytes("FirstRandomString")),
```
CJ42, line 685: not sure to see here which data key is not allowed.

I guess the first one because it does not have 16 x `00`s? Maybe you could hash another word to make this more explicit.

```suggestion
          ethers.utils.keccak256(ethers.utils.toUtf8Bytes("Not Allowed ERC725Y data key")),
```

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
```diff
@@ -0,0 +1,840 @@
...
+    /**
+     * Testing the following permissions is skipped because
+     * execute(..) is not available in a token contract.
+     *
+     * CALL
+     * STATICCALL
+     * DELEGATECALL
+     * DEPLOY
+     * TRANSFERVALUE
+     */
```
CJ42, line 756: How about moving the describe block when testing `execute(...)` here?
We could also add the other tests for CALL, DELEGATECALL, STATICCALL, DEPLOY and TRANSFERVALUE in the future. But this is overkill for now.

CJ42 (changes requested):
Nice testing for some potential future use cases! I have added some review comments.

tests/LSP6KeyManager/tests/LSP6ControlledToken.test.ts
b00ste: you’re right and the test already exists on line 170

YamenMerhi (approved):

CJ42 (approved):

PR #309, opened 2022-09-12
ci: add workflow to deploy + verify contracts on L16 Blockscout explorer

CJ42:

Add a workflow that on each new release automatically:

1. deploy lsp-smart-contracts on L16 network
2. verify the contracts, so that the methods fields on L16 Blockscout display.

### DIFF
--- a/.github/workflows/deploy-verify.yml
+++ b/.github/workflows/deploy-verify.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Deploy + Verify LSP7Mintable on L16
         run: |
           npx hardhat deploy --network luksoL16 --tags LSP7Mintable >> LSP7_deployment.txt
-          LSP7_Address=$(grep -o -E '0x(\w|\s){40}' URD_deployment.txt | tail -n 1)
+          LSP7_Address=$(grep -o -E '0x(\w|\s){40}' LSP7_deployment.txt | tail -n 1)
           echo "module.exports = [ 'LSP7 Mintable', 'LSP7M', '$DEPLOYER_ADDRESS', false ];" >> lsp7arguments.js
           npx hardhat verify $LSP7_Address --constructor-args lsp7arguments.js --network luksoL16
 
--- a/.github/workflows/deploy-verify.yml
+++ b/.github/workflows/deploy-verify.yml
@@ -52,6 +52,7 @@ jobs:
             "UniversalProfile",
             "LSP6KeyManager",
             "LSP1UniversalReceiverDelegateUP",
+            "LSP1UniversalReceiverDelegateVault",
             "LSP7Mintable",
             "LSP8Mintable",
             "LSP9Vault",
--- a/.github/workflows/deploy-verify.yml
+++ b/.github/workflows/deploy-verify.yml
@@ -12,6 +12,9 @@ on:
       - "ci/verify-contracts"
 
 jobs:
+  build:
+    runs-on: ubuntu-latest
+
   deploy-verify:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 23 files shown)


.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
+# This workflow deploys and verify the lsp-smart-contracts and verify them.
+
+name: Deploy + Verify Contracts
+
+env:
+  CONTRACT_VERIFICATION_PK: ${{ secrets.CONTRACT_VERIFICATION_PK }}
+  DEPLOYER_ADDRESS: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
```
CallumGrindle, line 7: Just a thought but maybe we will need to replicate this in future across different networks, we can use a matrix or something with different keys perhaps

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+          path: |
+            artifacts
+            node_modules 
+            types
+          key: ${{ github.run_id }}
+
+      # Save the output of the deployment in a text file.
+      # For the UP, we need to specify the contract as UniversalProfile and LSP0ERC725Account have the same runtime code.
+      - name: Deploy + Verify UniversalProfile + LSP6KeyManager on L16
+        if: always()
```
CallumGrindle, line 46: is this necessary?

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+          npx hardhat deploy --network luksoL16 --tags LSP6KeyManager --reset >> UP_KM_deployment.txt
+          UP_ADDRESS=$(grep -o -E '0x(\w|\s){40}' UP_KM_deployment.txt | head -n 2 | tail -n 1)
+          npx hardhat verify $UP_ADDRESS $DEPLOYER_ADDRESS --network luksoL16 --contract contracts/UniversalProfile.sol:UniversalProfile
+          KM_ADDRESS=$(grep -o -E '0x(\w|\s){40}' UP_KM_deployment.txt | tail -n 1)
+          npx hardhat verify $KM_ADDRESS $UP_ADDRESS --network luksoL16
+
+      - name: Deploy + Verify LSP1UniversalReceiverDelegateUP on L16
+        if: always()
+        run: |
+          npx hardhat deploy --network luksoL16 --tags LSP1UniversalReceiverDelegateUP --reset >> URD_deployment.txt
```
CallumGrindle, line 57: We might be able to chain these commands directly from the output of the previous command rather than creating the txt file

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+        run: |
+          npx hardhat deploy --network luksoL16 --tags LSP7Mintable --reset >> LSP7_deployment.txt
+          LSP7_Address=$(grep -o -E '0x(\w|\s){40}' LSP7_deployment.txt | tail -n 1)
+          echo "module.exports = [ 'LSP7 Mintable', 'LSP7M', '$DEPLOYER_ADDRESS', false ];" >> lsp7arguments.js
+          npx hardhat verify $LSP7_Address --constructor-args lsp7arguments.js --network luksoL16
+
+      - name: Deploy + Verify LSP8Mintable on L16
+        if: always()
+        run: |
+          npx hardhat deploy --network luksoL16 --tags LSP8Mintable --reset >> LSP8_deployment.txt
```
CallumGrindle, line 72: Maybe there is a way we can reuse this code as I think it is the same for each verification

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+          npx hardhat deploy --network luksoL16 --tags LSP1UniversalReceiverDelegateUP --reset >> URD_deployment.txt
+          URD_ADDRESS=$(grep -o -E '0x(\w|\s){40}' URD_deployment.txt | tail -n 1)
+          npx hardhat verify $URD_ADDRESS --network luksoL16
+
+      - name: Deploy + Verify LSP7Mintable on L16
+        if: always()
+        run: |
+          npx hardhat deploy --network luksoL16 --tags LSP7Mintable --reset >> LSP7_deployment.txt
+          LSP7_Address=$(grep -o -E '0x(\w|\s){40}' LSP7_deployment.txt | tail -n 1)
+          echo "module.exports = [ 'LSP7 Mintable', 'LSP7M', '$DEPLOYER_ADDRESS', false ];" >> lsp7arguments.js
```
CallumGrindle, line 66: This is not in the other deployments, do we need it here? Why specifically for LSP7Mintable deployment?

.github/workflows/deploy-verify.yml
CallumGrindle: You do this for the base contracts so I think we can do something similar

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Deploy + Verify LSP9Vault on L16
+        if: always()
+        run: |
+          npx hardhat deploy --network luksoL16 --tags LSP9Vault --reset >> LSP9_deployment.txt
+          LSP9_ADDRESS=$(grep -o -E '0x(\w|\s){40}' LSP9_deployment.txt | tail -n 1)
+          npx hardhat verify $LSP9_ADDRESS $DEPLOYER_ADDRESS --network luksoL16
+
+  deploy-verify-base:
+    if: always()
```
CallumGrindle, line 85: I think we can remove these

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: "16.x"
+          cache: "npm"
+
+      - name: Install dependencies
+        if: steps.build-cache.outputs.cache-hit != 'true'
+        run: npm ci
+
+      - name: Deploy + Verify ${{ matrix.base_contract }} on L16
```
CallumGrindle, line 122: Cool. This is nice that it uses the matrix so we only need to maintain this code once

DEPLOYMENT.md
```diff
@@ -45,10 +45,14 @@ Available `--tags <options>` are:
 
 - `LSP8Mintable`: deploy a `LSP7Mintable` contract (NFT), using the deployer address as the owner and allowing this deployer address to then mint tokens.
 
-- `LSP7MintableInit`: deploy + initialize (= lock) a `LSP7MintableInit` contract (Token), that can be used as implementation behind proxy.  The base contract is deployed with the `isNFT_` parameter set to `false`, making the implementation token divisible.
+- `LSP7MintableInit`: deploy + initialize (= lock) a `LSP7MintableInit` contract (Token), that can be used as implementation behind proxy. The base contract is deployed with the `isNonDivisible_` parameter set to `false`, making the implementation token divisible.
```
CallumGrindle, line 48: Do we want these changes?

CJ42: yes it was forgotten in a previous PR.

.github/workflows/deploy-verify.yml
CJ42: @CallumGrindle actually using the matrix for deploying does not work.
The account gets confused and see 5 transactions are the same time being submitted to the blockchain, and think each individual transaction aims to replace the previous one.

https://github.com/lukso-network/lsp-smart-contracts/actions/runs/3043284772/jobs/4902570169

This error comes up: 
image

.github/workflows/deploy-verify.yml
CJ42: In LSP7Mintable, there is a 4th parameter `isNonDivisible_` that is a `boolean`.
We tried verifying the contracts with @YamenMerhi and when passing the boolean parameter in the CLI, it does not work.

Related to this issue: https://github.com/NomicFoundation/hardhat/issues/2635

I have re-opened a new one here: https://github.com/NomicFoundation/hardhat/issues/3141

.github/workflows/deploy-verify.yml
CJ42: Yes it would be good if we could find a way to cleanup a bit.
It only differs for the constructor arguments in the verification steps.

.github/workflows/deploy-verify.yml
CJ42: with the pipe operator maybe? `|`

.github/workflows/deploy-verify.yml
CJ42: Yes so the reason for this is that for some random reasons, hardhat etherscan return an error `unknown UID` while the contract actually verified successfully.

See this job for instance: https://github.com/lukso-network/lsp-smart-contracts/actions/runs/3043284772/jobs/4902379921

image

By adding `if: always()`, it enforces the next step to run, regardless of the outcome of the previous step (if it succeeded or not).

CallumGrindle: hum okay interesting. It just means we will never know if this workflow has truly been successful or not

.github/workflows/deploy-verify.yml
CallumGrindle: Haha ah yes indeed, fun nonce issues. Maybe there is a way to wait for the previous tx to be mined

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+        if: always()
+        run: |
+          npx hardhat deploy --network luksoL16 --tags ${{ matrix.base_contract }} --reset >> base_deployment.txt
+          BASE_CONTRACT_ADDRESS=$(grep -o -E '0x(\w|\s){40}' base_deployment.txt.txt | tail -n 1)
+          npx hardhat verify $BASE_CONTRACT_ADDRESS --network luksoL16
```
Hugoo, line 127: i think this code is duplicated a lot of time - move it to a script and call it with parameters

.github/workflows/deploy-verify.yml
CJ42: cf discussion above.

.github/workflows/deploy-verify.yml
```diff
@@ -0,0 +1,127 @@
...
+      # Save the output of the deployment in a text file.
+      # For the UP, we need to specify the contract as UniversalProfile and LSP0ERC725Account have the same runtime code.
+      - name: Deploy + Verify UniversalProfile + LSP6KeyManager on L16
+        if: always()
+        run: |
+          npx hardhat deploy --network luksoL16 --tags LSP6KeyManager --reset >> UP_KM_deployment.txt
+          UP_ADDRESS=$(grep -o -E '0x(\w|\s){40}' UP_KM_deployment.txt | head -n 2 | tail -n 1)
+          npx hardhat verify $UP_ADDRESS $DEPLOYER_ADDRESS --network luksoL16 --contract contracts/UniversalProfile.sol:UniversalProfile
+          KM_ADDRESS=$(grep -o -E '0x(\w|\s){40}' UP_KM_deployment.txt | tail -n 1)
+          npx hardhat verify $KM_ADDRESS $UP_ADDRESS --network luksoL16
```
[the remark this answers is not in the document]
CJ42: @Hugoo implemented a custom shell script ✅ . See latest commits.

.github/workflows/deploy-verify.yml
CJ42: resolved with `matrix` + option `max-parallel: 1` ✅

.github/workflows/deploy-verify.yml
CJ42: Done. Integrated everything in one big shell script.

.github/workflows/deploy-verify.yml
CJ42: I have removed the `if: always()` and use a matrix instead that run each deployment + verification one at a time.
If something goes wrong, we can always re-run it manually via the `workflow_dispatch` trigger.

.github/workflows/deploy-verify.yml
CJ42: @CallumGrindle I haven't implemented that, I couldn't find the way to do it.

Hugoo: Yeah it is fine ;)

CallumGrindle (approved):

Hugoo (approved):
very very cool feature ! let's test it :))

CJ42:
All the deployment passed ✅ See here: https://github.com/lukso-network/lsp-smart-contracts/actions/runs/3051468823/jobs/4919734978

I will configure the workflow just for the `main` branch and merge.

PR #310, opened 2022-09-13
fix:  Reset `pendingOwner` whenever `renounceOwnership(..)` is used.

b00ste:

This PR fixes the bug described in this [tweet](https://twitter.com/pcaversaccio/status/1569373269228142592?t=3i66r7ZBWTL6ZwWREA6MCQ&s=19). Also there is a test added to check that this behaviour is not happening.

Unwanted behaviour:

Step 1: Using `transferOwnership(..)` to set a `pendingOwner`.
Step 2: Using `renounceOwnerhisp(..)` which updates the `owner` but not the `pendingOwner`.
Step 3: Using `claimOwnership(..)` which will make `pendingOwner` the `owner`.

In this way one could renounce ownership of the contract for the public but whenever it is the most profitable the `pendingOwner` will be able to claim ownership of the contract.

To mitigate this unwanted behaviour the `pendingOwner` is reseted whenever `renounceOwnership(..)` is used.

### DIFF
--- a/contracts/Custom/ClaimOwnership.sol
+++ b/contracts/Custom/ClaimOwnership.sol
@@ -63,7 +63,7 @@ abstract contract ClaimOwnership is IClaimOwnership, OwnableUnset {
     function _claimOwnership() internal virtual {
         require(msg.sender == pendingOwner, "ClaimOwnership: caller is not the pendingOwner");
         _setOwner(pendingOwner);
-        pendingOwner = address(0);
+        delete pendingOwner;
     }
 
     function _transferOwnership(address newOwner) internal virtual {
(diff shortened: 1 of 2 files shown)


CJ42 (changes requested):
We need to add some tests for the scenario explained here: https://twitter.com/pcaversaccio/status/1569373269228142592?t=3i66r7ZBWTL6ZwWREA6MCQ&s=19

Also @b00ste can you add a bit more description in the PR body about the issue?
If the tweet ever get deleted, we still have some content to refer to.

b00ste:
> We need to add some tests for the scenario explained here: https://twitter.com/pcaversaccio/status/1569373269228142592?t=3i66r7ZBWTL6ZwWREA6MCQ&s=19
> 
> we need to ensure with a test that this potential scenario is fixed.
> 
> Also @b00ste can you add a bit more description in the PR body about the issue? If the tweet ever get deleted, we still have some content to refer to.

CJ42 (approved):

YamenMerhi (approved):

PR #311, opened 2022-09-15
fix: add reference to `.key` member for SupportedStandards in constants.js

CJ42:
In the `constants.js` file, for the data key `SupportedStandards:...` under `ERC725YKeys`, specify the member `key` to only list data keys under these constants.


Esanim (approved):
Ok for me

YamenMerhi (approved):
Good catch

PR #319, opened 2022-09-20
refactor: refactor LSP1UniversalReceiverDelegate code

YamenMerhi:
- Refactor the code of the URD of the UP: Add Custom errors + adjust return statement to be more descriptive
- Refactor the code of the URD of the Vault
- Remove UtilsLib library
- Split LSP5Utils to LSP5Utils and LSP10Utils
- Add a check in the URD to disallow EOA registering as assets.


contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * @dev Generating the data keys/values to be set on the sender address after sending vaults
+     * @param sender The address sending the vault and where the Keys should be updated
+     * @param vaultMapKey The map key of the vault being sent
+     * @param vaultInterfaceIdAndIndex The value of the map key of the vault being sent
+     */
+    function generateSentVaultKeys(
+        address sender,
+        bytes32 vaultMapKey,
+        bytes memory vaultInterfaceIdAndIndex
```
CJ42, line 101: Just as a note for future. 

You might potentially need some sanitization check here, if this variable is not encoded correctly (`bytes4` + `bytes8`).

In the context of our contracts, it is ok. But in other context if this library is used as standalone, there might be this issue potentially.

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,183 @@
...
+    ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
+        IERC725Y account = IERC725Y(sender);
```
CJ42, line 103: Wouldn't it be better to specify the type of this variable as `IERC725Y` directly (in the function parameter), instead of casting?

contracts/LSP1UniversalReceiver/LSP1Errors.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+/**
+ * @dev reverts when EOA calls the `universalReceiver(..)` function with an asset/vault typeId
+ * @param caller The address of the EOA
+ */
+error CannotRegisterEOAsAsAssets(address caller);
```
CJ42, line 8: Suggesting different name.

As the name here suggest that the address of the EOA is an asset. While the error actually occurs if an EOA make a call to try to register an asset.

```suggestion
error EOAsCannotRegisterAssets(address caller);
```

contracts/LSP1UniversalReceiver/LSP1Errors.sol
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * @dev reverts when the account calling the UniversalReceiverDelegate is not the same account
+ * linked as target in the KeyManager -- Security Check
+ * @param account The address of the account implementing the `universalReceiver(..)` function
+ * @param target The address of the target linked to the KeyManager
+ */
+error MsgSenderIsNotTheLinkedTarget(address account, address target);
```
CJ42, line 16: Suggesting a bit shorter name:

```suggestion
error CallerNotLinkedTarget(address account, address target);
```

**Or maybe to be very explicit, and states that this relates to LSP6?**

```suggestion
error CallerNotLSP6LinkedTarget(address account, address target);
```

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,183 @@
...
+        address receiver,
+        address vault,
+        bytes32 vaultMapKey,
+        bytes4 interfaceID
+    ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
+        keys = new bytes32[](3);
+        values = new bytes[](3);
+
+        IERC725Y account = IERC725Y(receiver);
+        bytes memory encodedArrayLength = account.getData(_LSP10_VAULTS_ARRAY_KEY);
```
CJ42, line 55: We can probably create a simple getter convenience function for this in the future named:

```solidity
function getLSP10ReceivedVaultsCount(IERC725Y account) public view returns(uint256) {
    // ...
}
```

That would make a very nice and handy convenience function.

Just a note, we can do that in a future PR :)

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
```diff
@@ -33,34 +37,124 @@ import {
...
+     */
+    function _whenReceiving(
+        bytes32 typeId,
+        address notifier,
+        address keyManager,
+        bytes32 notifierMapKey,
+        bytes4 interfaceID
+    ) internal returns (bytes memory result) {
+        // if it's a token transfer (LSP7/LSP8)
+        if (typeId != _TYPEID_LSP9_VAULTRECIPIENT) {
```
CJ42, line 98: I am not sure about this `if / else` statement. As this means:
```
 "if the `typeId` IS NOT `_TYPEID_LSP9_VAULTRECIPIENT` {
     // then register an asset
 } 
 ```
 
 Which technically is not the case. An asset should be registered **only if it matches LSP7 or LSP8 Received Assets `typeId`s.**

File: tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts [hunk 7]
```diff
@@ -81,6 +90,48 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+      });
+    });
+
+    describe("when calling with random bytes32 typeId", () => {
+      it("should pass and return typeId out of scope return value", async () => {
+        let result = await context.universalProfile1
+          .connect(context.accounts.any)
+          .callStatic.universalReceiver(LSP1_HOOK_PLACEHOLDER, "0x");
+
+        expect(result).to.be.equal(
```
CJ42, line 126:
```suggestion
        expect(result).to.equal(
```

contracts/LSP10ReceivedVaults/LSP10Utils.sol
YamenMerhi: Would be a great idea! Added the idea as a ticket in the Backlog

contracts/LSP1UniversalReceiver/LSP1Errors.sol
YamenMerhi: I will go with the second one

contracts/LSP1UniversalReceiver/LSP1Errors.sol
YamenMerhi: But actually, this is what's happening, an EOA cannot register a different asset, when the EOA is calling the `universalReceiver(..)` function with an LSP7 recipient typeId, it's registering the address of the same EOA calling as an asset.

contracts/LSP10ReceivedVaults/LSP10Utils.sol
YamenMerhi: It would be the same as i need to typecast the parameter in the UniversalReceiverDelegate, so I prefer to have it here to avoid complexity and more variables in the URD.

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,183 @@
...
+            keys[2] = vaultMapKey;
+            values[2] = bytes.concat(interfaceID, bytes8(0));
+
+            // If the storage is already initiated
+        } else if (encodedArrayLength.length == 32) {
+            uint256 oldArrayLength = uint256(bytes32(encodedArrayLength));
+            uint256 newArrayLength = oldArrayLength + 1;
+
+            keys[0] = _LSP10_VAULTS_ARRAY_KEY;
+            values[0] = bytes.concat(bytes32(newArrayLength));
```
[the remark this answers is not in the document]
YamenMerhi: Sure will remove it

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
YamenMerhi: The typeIds that reach this internal function are just 3 typeIds:
`_TYPEID_LSP7_TOKENRECIPIENT`
`_TYPEID_LSP8_TOKENRECIPIENT`
`_TYPEID_LSP9_VAULTRECIPIENT`
Because we check the typeIds in `LSP1Utils.getTransferDetails` and if it's an unknow typeId we return.

So in this case I am choosing between these 3 typeIds that it's not a typeId relative to the vault, then for sure it's relative to the token.

I can do `if (typeId == _TYPEID_LSP7_TOKENRECIPIENT || typeId == _TYPEID_LSP8_TOKENRECIPIENT)` But it would be more expensive in both conditions.

contracts/LSP10ReceivedVaults/LSP10Utils.sol
YamenMerhi: Good idea, will create a ticket for it

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * @dev Generating the data keys/values to be set on the receiver address after receiving vaults
+     * @param receiver The address receiving the vault and where the Keys should be added
+     * @param vault The address of the vault being received
+     * @param interfaceID The interfaceID of the vault being received
+     */
+    function generateReceivedVaultKeys(
+        address receiver,
+        address vault,
+        bytes32 vaultMapKey,
```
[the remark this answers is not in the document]
YamenMerhi: We already need to have it in the URD code as we need to check if the map value for this key exists or not.
Instead of constructing the KeyTwice we do it here for gas optimizations

b00ste (approved):

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault/LSP1UniversalReceiverDelegateVault.sol
```diff
@@ -26,29 +29,50 @@ import {
...
+        // solhint-disable avoid-tx-origin
+        if (notifier == tx.origin) revert CannotRegisterEOAsAsAssets(notifier);
+
+        bytes32 notifierMapKey = LSP2Utils.generateMappingKey(mapPrefix, bytes20(notifier));
+        bytes memory notifierMapValue = IERC725Y(msg.sender).getData(notifierMapKey);
+
+        if (isReceiving) {
+            // if the map value is already set, then do nothing
+            if (bytes12(notifierMapValue) != bytes12(0))
+                return "URD: asset received is already registered";
```
skimaharvey, line 58: string of over 32 bytes

YamenMerhi: I can't think of a shorter description for now 😅

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,182 @@
...
+    function generateSentVaultKeys(
+        address sender,
+        bytes32 vaultMapKey,
+        bytes memory vaultInterfaceIdAndIndex
+    ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
+        IERC725Y account = IERC725Y(sender);
+
+        // Updating the number of the received vaults
+        uint256 oldArrayLength = uint256(bytes32(account.getData(_LSP10_VAULTS_ARRAY_KEY)));
+        uint256 newArrayLength = oldArrayLength - 1;
```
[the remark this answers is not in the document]
YamenMerhi: Same reason commented above

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -0,0 +1,182 @@
...
+        address receiver,
+        address vault,
+        bytes32 vaultMapKey,
+        bytes4 interfaceID
+    ) internal view returns (bytes32[] memory keys, bytes[] memory values) {
+        keys = new bytes32[](3);
+        values = new bytes[](3);
+
+        IERC725Y account = IERC725Y(receiver);
+        bytes memory encodedArrayLength = account.getData(_LSP10_VAULTS_ARRAY_KEY);
```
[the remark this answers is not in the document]
YamenMerhi: `encodedArrayValues` is more generic than `encodedArrayLength` so I think it's better if we keep it as it is.

PR #320, opened 2022-09-21
fix: add checks for ADD / CHANGE permissions when data key is `AddressPermissions[index]`

CJ42:

**Bug fix**

When the data key passed to the Key Manager is a key-value pair for `AddressPermissions[index]`, the current behaviour is that the caller/signer must have the permission `CHANGE PERMISSION`.

Add additional check to verify what is stored under this type of data key in the storage.

- if some value are set under `AddressPermissions[index]` = caller must have permission `CHANGE PERMISSION`.
- if no value is set under `AddressPermissions[index]` = caller must have permission `ADD PERMISSION`.


YamenMerhi (approved):
Nice

b00ste (approved):

PR #323, opened 2022-09-22
feat!: Replace custom/ClaimOwnership with LSP14Ownable2Step

b00ste:

## ⚠️ BREAKING CHANGES

In this PR the contracts for custom Ownable are moved from the Custom folder to a new folder, LSP14Ownable2Step.
The contracts are updated according to the [OpenZeppelin - Ownable2Step](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable2Step.sol)
The tests are also updated with the new naming of contracts and methods as well as with the new type ids and interface ids.

The following updates were made:

- References of `ClaimOwnership` as the name of the contracts are changed to either `LSP14` or `LSP14Ownable2Step`
- References of `claimOwnership(..)` as a method name are changed to `acceptOwnership(..)`
-  The `pendingOwner` address variable is transformed from `public` to `private` and a new method is added for getting the address, `pendingOwner(..)`
- Because of the change from `claimOwnership(..)` to `acceptOwnership(..)` and the new `pendingOwner(..)` there will be new interfaces id of LSP0, LSP9 and LSP14(old ClaimOwnership):
    1. LSP0: `0xeb6be62e` -> `0xdca05671`
    2. LSP9: `0xfd4d5c50` -> `0xca86ec0f`
    3. LSP14(old ClaimOwnership): `0xa375e9c6` -> `0x94be5999`
- The following type ids:
    1. `LSP9_VAULTSENDER` -> `LSP14_OwnershipTransferred_SenderNotification`
    4. `LSP9_VAULTRECIEVER` -> `LSP14_OwnershipTransferred_RecipientNotification`
    5. `LSP9_VAULTPENDINGOWNER` -> `LSP14_OwnershipTransferStarted`
Are updated and now used in both LSP0 and LSP9 as URD hooks for notifying when starting a trasnfer ownership process and when the transfer ownership process is complete


contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
```diff
@@ -95,7 +96,7 @@ contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiverDelega
         bytes4 interfaceID
     ) internal returns (bytes memory result) {
         // if it's a token transfer (LSP7/LSP8)
-        if (typeId != _TYPEID_LSP9_VAULTRECIPIENT) {
+        if (typeId != _TYPEID_LSP14_OwnershipTransferred_RecipientNotification) {
```
CJ42, line 99: @YamenMerhi @b00ste 

We should look at how this block of code will evolve once we implement the notification hook inside the `transferOwnership(...)` function of LSP0.

Here anything that matches the LSP14 typeId will fall under the `else` block and generate vault keys via `generateReceivedVaultKeys(...)`.

Once we implement the notification hook inside `transferOwnership(...)` of LSP0, vault keys will be generated as well. **But it's not ownership of a vault that will be transferred in this case, but ownership of an LSP0 ERC725Account**.

YamenMerhi: There is already a ticket created yesterday for this topic -- after discussion with Daniel 
https://app.clickup.com/t/32qj7d9

contracts/LSP14Ownable2Step/LSP14Constants.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+bytes4 constant _INTERFACEID_LSP14 = 0x94be5999;
+
+// --- Type IDs
+
+//keccak256(LSP14OwnershipTransferStarted)
+bytes32 constant _TYPEID_LSP14_OwnershipTransferStarted = 0xee9a7c0924f740a2ca33d59b7f0c2929821ea9837ce043ce91c1823e9c4e52c0;
```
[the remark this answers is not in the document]
YamenMerhi: Once we merge this PR, and as discussed, i will follow up with another PR changing all TypeIds, no need to double work.

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
```diff
@@ -0,0 +1,169 @@
...
+        uint256 confirmationPeriodEnd = confirmationPeriodStart +
+            _RENOUNCE_OWNERSHIP_CONFIRMATION_PERIOD;
+
+        if (currentBlock > confirmationPeriodEnd) {
+            _renounceOwnershipStartedAt = currentBlock;
+            emit RenounceOwnershipInitiated();
+            return;
+        }
+
+        if (currentBlock < confirmationPeriodStart) {
```
skimaharvey, line 145: I think here should be `<=`

YamenMerhi (approved):

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
CJ42: to be discussed

CJ42 (approved):

PR #325, opened 2022-09-27
test: Add test for verifying that one cannot use `acceptOwnership` in the same trx as `transferOwnership`

b00ste:

This PR creates a UP contract for testing, with a modified `universalReceiver(..)` which calls `acceptOwnership(..)` whenever the `universalReceiver(..)` is called with the specific typeId used to notify the new owner that he can accept ownership of a contract.
A new test is introduced that should make sure that the transaction is reverted whenever someone calls `trasferOwnership(..)` and `acceptOwnership(..)` is executed in the same block.

### DIFF
--- a/contracts/Helpers/UPWithInstantAcceptOwnership.sol
+++ b/contracts/Helpers/UPWithInstantAcceptOwnership.sol
@@ -28,6 +28,6 @@ contract UPWithInstantAcceptOwnership is LSP0ERC725AccountCore {
         if (typeId == _TYPEID_LSP14_OwnershipTransferStarted) {
             LSP14Ownable2Step(msg.sender).acceptOwnership();
         }
-        returnedValue = LSP0ERC725AccountCore.universalReceiver(typeId, receivedData);
+        super.universalReceiver(typeId, receivedData);
     }
 }
--- a/contracts/Helpers/UPWithInstantAcceptOwnership.sol
+++ b/contracts/Helpers/UPWithInstantAcceptOwnership.sol
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+// modules
+import {LSP0ERC725AccountCore} from "../LSP0ERC725Account/LSP0ERC725AccountCore.sol";
+import {OwnableUnset} from "@erc725/smart-contracts/contracts/custom/OwnableUnset.sol";
+import {LSP14Ownable2Step} from "../LSP14Ownable2Step/LSP14Ownable2Step.sol";
+
+// constants
+import "../LSP14Ownable2Step/LSP14Constants.sol";
+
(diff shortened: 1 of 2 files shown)


contracts/Helpers/UPWithInstantAcceptOwnership.sol
```diff
@@ -0,0 +1,33 @@
...
+        public
+        payable
+        virtual
+        override
+        returns (bytes memory returnedValue)
+    {
+        if (typeId == _TYPEID_LSP14_OwnershipTransferStarted) {
+            LSP14Ownable2Step(msg.sender).acceptOwnership();
+        }
+        returnedValue = LSP0ERC725AccountCore.universalReceiver(typeId, receivedData);
```
CJ42, line 31: I think here you can simply do:

```suggestion
        super.universalReceiver(typeId, receivedData);
```

And the rest of the logic of the current `universalReceiver(...)` function from LSP0 will run.

But not a big deal, it is just a helper contract in the end :)

tests/LSP14Ownable2Step/LSP14Ownable2Step.behaviour.ts
```diff
@@ -127,6 +133,26 @@ export const shouldBehaveLikeLSP14 = (
         expect(accountBalanceAfter).to.be.lt(accountBalanceBefore);
       });
     });
+
+    describe("when the URD of the `newOwner` calls `acceptOwnership`", () => {
```
CJ42, line 137: I would suggest a different text description for this test.

```suggestion
    describe("when `transferOwnership(...)` is called in the same tx as `acceptOwnership(...)`", () => {
```

> **NB: ** also the text here states *when the URD of the `newOwner` calls `acceptOwnership`*. But this is not true, here the re-entrant callback comes from the UP via `universalReceiver(...)` function. 
> It's just a note, but we should double check that we describe the tests being run properly, to keep track correctly of what we actually test and don't test.

tests/LSP14Ownable2Step/LSP14Ownable2Step.behaviour.ts
```diff
@@ -127,6 +133,26 @@ export const shouldBehaveLikeLSP14 = (
...
+      let upWithCustomURD: UPWithInstantAcceptOwnership;
+      before(async () => {
+        context = await buildContext();
+        upWithCustomURD = await new UPWithInstantAcceptOwnership__factory(
+          context.accounts[0]
+        ).deploy(context.accounts[0].address);
+      });
+
+      it("should revert with 'LSP14: newOwner MUST accept ownership in a separate transaction'", async () => {
```
CJ42, line 146:
```suggestion
      it("should revert (e.g: if `universalReceiver(...)` function of `newOwner` calls directly `acceptOwnership(...)`", async () => {
```

YamenMerhi (approved):
Great to have it tested!!

CJ42 (approved):

PR #326, opened 2022-09-27
feat: add routing (= external call ) to additional contract in `universalReceiver(...)` function based on typeId

YamenMerhi:
- Adding routing to the universalReceiver function
- Add tests to check how ABI encoding works and gas costs
- Resolve test according to new behaviour

### DIFF
--- a/contracts/Helpers/ABIEncoder.sol
+++ b/contracts/Helpers/ABIEncoder.sol
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+contract ABIEncoder {
+    function encode(bytes memory a, bytes memory b)
+        public
+        returns (bytes memory c, uint256 gasUsed)
+    {
+        uint256 gasUsed1 = gasleft();
+        c = abi.encode(a, b);
+        uint256 gasUsed2 = gasleft();
+        gasUsed = gasUsed1 - gasUsed2;
+    }
+
+    function decode(bytes memory c) public pure returns (bytes memory a, bytes memory b) {
+        (a, b) = abi.decode(c, (bytes, bytes));
+    }
+}
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -141,44 +141,46 @@ abstract contract LSP0ERC725AccountCore is
 
     /**
      * @notice Triggers the UniversalReceiver event when this function gets executed successfully.
-     * @dev Forwards the call to the UniversalReceiverDelegate if set.
+     * Forwards the call to the addresses stored in the ERC725Y storage under the LSP1UniversalReceiverDelegate
+     * Key and the typeId Key (param) respectivelly. The call will be discarded if no addresses were set.
+     *
      * @param typeId The type of call received.
      * @param receivedData The data received.
-     * @return returnedValue The ABI encoded return value of the LSP1UniversalReceiverDelegate call
-     * and the LSP1TypeIdDelegate call
+     * @return returnedValues The ABI encoded return value of the LSP1UniversalReceiverDelegate call
+     * and the LSP1TypeIdDelegate call.
      */
     function universalReceiver(bytes32 typeId, bytes calldata receivedData)
         public
         payable
         virtual
-        returns (bytes memory returnedValue)
+        returns (bytes memory returnedValues)
     {
         bytes memory lsp1DelegateValue = _getData(_LSP1_UNIVERSAL_RECEIVER_DELEGATE_KEY);
-        bytes memory returnedValue1;
+        bytes memory resultMainDelegate;
 
         if (lsp1DelegateValue.length >= 20) {
             address universalReceiverDelegate = address(bytes20(lsp1DelegateValue));
 
             if (universalReceiverDelegate.supportsERC165Interface(_INTERFACEID_LSP1_DELEGATE)) {
-                returnedValue1 = ILSP1UniversalReceiverDelegate(universalReceiverDelegate)
+                resultMainDelegate = ILSP1UniversalReceiverDelegate(universalReceiverDelegate)
                     .universalReceiverDelegate(msg.sender, msg.value, typeId, receivedData);
             }
         }
 
         bytes memory lsp1TypeIdDelegateValue = _getData(typeId);
-        bytes memory returnedValue2;
+        bytes memory resultTypeIdDelegate;
 
         if (lsp1TypeIdDelegateValue.length >= 20) {
             address universalReceiverDelegate = address(bytes20(lsp1TypeIdDelegateValue));
 
             if (universalReceiverDelegate.supportsERC165Interface(_INTERFACEID_LSP1_DELEGATE)) {
-                returnedValue2 = ILSP1UniversalReceiverDelegate(universalReceiverDelegate)
+                resultTypeIdDelegate = ILSP1UniversalReceiverDelegate(universalReceiverDelegate)
                     .universalReceiverDelegate(msg.sender, msg.value, typeId, receivedData);
             }
         }
 
-        returnedValue = abi.encode(returnedValue1, returnedValue2);
-        emit UniversalReceiver(msg.sender, msg.value, typeId, receivedData, returnedValue);
+        returnedValues = abi.encode(resultMainDelegate, resultTypeIdDelegate);
+        emit UniversalReceiver(msg.sender, msg.value, typeId, receivedData, returnedValues);
     }
 
     /**
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -142,7 +142,7 @@ abstract contract LSP0ERC725AccountCore is
     /**
      * @notice Triggers the UniversalReceiver event when this function gets executed successfully.
      * Forwards the call to the addresses stored in the ERC725Y storage under the LSP1UniversalReceiverDelegate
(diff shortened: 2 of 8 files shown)


contracts/Helpers/ABIEncoder.sol
```diff
@@ -0,0 +1,18 @@
...
+pragma solidity ^0.8.0;
+
+contract ABIEncoder {
+    function encode(bytes memory a, bytes memory b)
+        public
+        returns (bytes memory c, uint256 gasUsed)
+    {
+        uint256 gasUsed1 = gasleft();
+        c = abi.encode(a, b);
+        uint256 gasUsed2 = gasleft();
```
CJ42, line 11: what are you trying to test? how much gas it cost to do `abi.encode` 2 x `bytes`?
The result might likely depend on how big these two `bytes` values are.

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -143,6 +144,8 @@ abstract contract LSP0ERC725AccountCore is
      * @dev Forwards the call to the UniversalReceiverDelegate if set.
      * @param typeId The type of call received.
      * @param receivedData The data received.
+     * @return returnedValue The ABI encoded return value of the LSP1UniversalReceiverDelegate call
```
CJ42, line 147: I think this docstring is incorrect now because of the changes this PR introduce.

1. I would suggest to rename the variable to plural: `returnValue` > `returnedValues`
2. This returned variable should be described as the result of **both calls**, bot results being abi encoded together.
3. we should also explain what are each of these calls in the docs string (to the default LSP1Delegate, then for the additional LSP1Delegate setup for this typeId).

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -151,20 +154,30 @@ abstract contract LSP0ERC725AccountCore is
         returns (bytes memory returnedValue)
     {
         bytes memory lsp1DelegateValue = _getData(_LSP1_UNIVERSAL_RECEIVER_DELEGATE_KEY);
+        bytes memory returnedValue1;
```
CJ42, line 157: I would recommend using different variable names here.
Because it makes it hard to understand otherwise the actual difference between `returnedValue1` and `returnedValue2`.

Even if it is long.

For instance:

```solidity
        // default LSP1 Delegate.
        bytes memory returnedDataDefaultLSP1Delegate;
        
        // LSP1Delegate for typeId
        bytes memory returnedDataLSP1DelegateForTypeID
```

@skimaharvey @b00ste do you have suggestions for better variable names?

CJ42 (commented):
added first round of review comments.

contracts/Helpers/ABIEncoder.sol
YamenMerhi: Exactly that's what i did in ABIEncoder.test.ts :)

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: I tried having these names, it will make the whole function looks ugly if we have very long name.
Anyway I am happy to do it.
My suggestion: we can leave it with shortnames or add the long names with a prettier-ignore to preserve the style of the function

CJ42: ok we can try to shorten it then. Probably no need to specify `LSP1` in the name of the variable since we are inside the `universalReceiver(...)` function, so the context is LSP1.

Let's name it:

```js
// default LSP1 Delegate.
bytes memory resultDelegate;
        
// LSP1Delegate for typeId
bytes memory resultTypeID
```

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -143,6 +144,8 @@ abstract contract LSP0ERC725AccountCore is
      * @dev Forwards the call to the UniversalReceiverDelegate if set.
```
CJ42, line 144: this docstring needs to be changed. To explain that it forwards the call to the other `LSP1TypeIdDelegate` contract.

It might be a good place here to explain what each of these calls are (briefly).

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -151,20 +154,30 @@ abstract contract LSP0ERC725AccountCore is
...
+                returnedValue1 = ILSP1UniversalReceiverDelegate(universalReceiverDelegate)
                     .universalReceiverDelegate(msg.sender, msg.value, typeId, receivedData);
             }
         }
+
+        bytes memory lsp1TypeIdDelegateValue = _getData(typeId);
+        bytes memory returnedValue2;
+
+        if (lsp1TypeIdDelegateValue.length >= 20) {
+            address universalReceiverDelegate = address(bytes20(lsp1TypeIdDelegateValue));
```
CJ42, line 172: maybe rename to:

```suggestion
            address typeIdDelegate = address(bytes20(lsp1TypeIdDelegateValue));
```

?

You can also declare the variable locally outside of each `if (...) { ... }`, and re-assign it. It might save like 10 gas, but not sure if it would make it more readable.

```solidity
address universalReceiverDelegate;

bytes memory lsp1DelegateValue = _getData(_LSP1_UNIVERSAL_RECEIVER_DELEGATE_KEY);

if (lsp1DelegateValue.length >= 20) {
    universalReceiverDelegate = address(bytes20(lsp1DelegateValue));
    
    // ...
}

bytes memory lsp1TypeIdDelegateValue = _getData(typeId);

if (lsp1TypeIdDelegateValue.length >= 20) {
    universalReceiverDelegate = address(bytes20(lsp1TypeIdDelegateValue));
    
    // ...
}
```

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -40,6 +40,7 @@ import {_INTERFACEID_LSP9} from "./LSP9Constants.sol";
  * @dev Could be owned by a UniversalProfile and able to register received asset with UniversalReceiverDelegateVault
  */
 contract LSP9VaultCore is ERC725XCore, ERC725YCore, LSP14Ownable2Step, ILSP1UniversalReceiver {
+    using ERC165Checker for address;
```
CJ42, line 43: I agree on this 👍
it looks much easier to read below compare to before

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -148,6 +149,8 @@ contract LSP9VaultCore is ERC725XCore, ERC725YCore, LSP14Ownable2Step, ILSP1Univ
      * @dev Forwards the call to the UniversalReceiverDelegate if set.
      * @param typeId The type of call received.
      * @param receivedData The data received.
+     * @return returnedValue The ABI encoded return value of the LSP1UniversalReceiverDelegate call
+     * and the LSP1TypeIdDelegate call
      */
     function universalReceiver(bytes32 typeId, bytes calldata receivedData)
```
CJ42, line 155: same comments I put on LSP0 apply here for LSP9.

tests/Helpers/ABIEncoder.test.ts
```diff
@@ -0,0 +1,187 @@
...
+describe("ABI Encoder Contract", () => {
+  let accounts: SignerWithAddress[];
+  let contract: ABIEncoder;
+
+  before(async () => {
+    accounts = await ethers.getSigners();
+    contract = await new ABIEncoder__factory(accounts[0]).deploy();
+  });
+
+  const verifyResult = async (txParameterA, txParameterB) => {
```
CJ42, line 15: nicely written tests 👌

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: Great idea, even tried this and it looks good ⬇️ 
```js
        bytes memory lsp1DelegateValue = _getData(_LSP1_UNIVERSAL_RECEIVER_DELEGATE_KEY);
        bytes memory resultMainDelegate;
        [..]
        bytes memory lsp1TypeIdDelegateValue = _getData(typeId);
        bytes memory resultTypeIdDelegate;
         [..]
        returnedValues = abi.encode(resultMainDelegate, resultTypeIdDelegate);
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -124,8 +125,14 @@ export const shouldBehaveLikeLSP1Delegate = (
           .callStatic.universalReceiver(LSP1_HOOK_PLACEHOLDER, "0x");
 
         expect(result).to.equal(
-          ethers.utils.hexlify(
-            ethers.utils.toUtf8Bytes("LSP1: typeId out of scope")
+          abiCoder.encode(
```
CJ42, line 128: _Suggestion_: maybe destructure the returned parameters?
And then do two `expect` for each.

```js
const [resultDelegate, resultTypeID] = abiCoder.decode(result)
expect(resultDelegate).to.equal(
    ethers.utils.hexlify(
        ethers.utils.toUtf8Bytes("LSP1: typeId out of scope")
    ))
expect(resultTypeID).to.equal("0x").
```

Otherwise, looks good to me :)

CJ42 (changes requested):
added review comments.

CJ42:
@YamenMerhi I will rename the title of the PR for the release CHANGELOG (in order to show the description properly under the feature subheading)

CJ42 (approved):

PR #327, opened 2022-09-28
refactor: construct typeIds based on mapping keyType

YamenMerhi:
- construct typeIds based on mapping keyType
- Update the typeIds in constant.js
- Remove hardoced constant in tests

TypeIds now are constructed as Mapping KeyType according to LSP2-ERC725YJSONSchema 

The usage of typeId in mapping form will make it more easy to identify which Keys and Values being set in the ERC725Y storage are relevant to a typeId based on the 10bytes prefix. 

This will also allow KeyManagers in the future to implement specific permission to set these kind of keys based on the fixed bytes10 prefix.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -200,20 +200,20 @@ const PERMISSIONS = {
 }
 
 const LSP1_TYPE_IDS = {
-	// keccak256('LSP7TokensSender')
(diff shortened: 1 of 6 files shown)


b00ste (approved):

CJ42 (approved):

CJ42:
@YamenMerhi I went too quickly on this PR. 
Since it's a new feature, we need to write some tests for it.
Can you do that, when the a an address is set for a specific typeId (inside this new LSP1 Mapping).

PR #328, opened 2022-10-04
refactor: Separate `fallback()` and `receive()` functions in LSP0 & LSP9

b00ste:

This PR separates the `fallback()` and `receive()` functions for the safe of separating the concern in LSP0 & LSP9: 
- receiving data (arbitrary or not) with no value -> `fallback()`
- receiving data (arbitrary or not) with value -> `fallback()`
- receiving just value with no data -> `receive()`


CJ42 (changes requested):
added review comments.

CJ42 (approved):

YamenMerhi (approved):

PR #330, opened 2022-10-05
feat: Add the `ValueReceived` event in all payable methods

b00ste:

In this PR the event `ValueReceived(address indexed sender, uint256 indexed value)` is emitted in the payable methods if `msg.value` is bigger than 0, meaning if value was received.

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -20,6 +20,7 @@ import {OwnableUnset} from "@erc725/smart-contracts/contracts/custom/OwnableUnse
 import {LSP14Ownable2Step} from "../LSP14Ownable2Step/LSP14Ownable2Step.sol";
 
 // constants
+import "@erc725/smart-contracts/contracts/constants.sol";
 import {
     _INTERFACEID_LSP0,
     _INTERFACEID_ERC1271,
@@ -209,7 +210,16 @@ abstract contract LSP0ERC725AccountCore is
     }
 
     /**
-     * @dev Emit the event ValueReceived in execute(..)
+     * @param operation The operation to execute: CALL = 0 CREATE = 1 CREATE2 = 2 STATICCALL = 3 DELEGATECALL = 4
+     * @param to The smart contract or address to interact with, `to` will be unused if a contract is created (operation 1 and 2)
+     * @param value The amount of native tokens to transfer (in Wei).
+     * @param data The call data, or the bytecode of the contract to deploy
+     * @dev Executes any other smart contract.
+     * SHOULD only be callable by the owner of the contract set via ERC173
+     *
+     * Emits a {Executed} event, when a call is executed under `operationType` 0, 3 and 4
+     * Emits a {ContractCreated} event, when a contract is created under `operationType` 1 and 2
+     * Emits a {ValueReceived} event, when receives native token
      */
     function execute(
         uint256 operation,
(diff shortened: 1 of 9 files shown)


YamenMerhi (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -206,4 +208,54 @@ abstract contract LSP0ERC725AccountCore is
...
+     * Emits a {ContractCreated} event, when a contract is created under `operationType` 1 and 2
+     * Emits a {ValueReceived} event, when receives native token
+     */
+    function execute(
+        uint256 operation,
+        address to,
+        uint256 value,
+        bytes memory data
+    ) public payable virtual override onlyOwner returns (bytes memory) {
+        require(address(this).balance >= value, "ERC725X: insufficient balance");
```
CJ42, line 230: Is it necessary to copy the whole code from ERC725X here? I think it will increase the size of the contract bytecode, and so the deployment cost. Can you compare the number of bytes with this current implementation _vs_ the changes I am suggesting below.

Maybe we can just emit the event first for now, but **I am not sure if emitting first is a good practice**. (probably not, as it violates the check-effect-interaction pattern. The effect is not a state change here, so it's not as important, but still an event being an **effect**, I would suggest we follow this pattern even for events when we can for best practices :) ).

This is my suggestion:

```suggestion
        if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
        super.execute(operation, to, value, data);
```

@YamenMerhi @b00ste I suggest in ERC725X that we separate the `require(...)` with the rest by putting the rest in an internal `_execute(...)` function. Let me know what you think.

So `ERC725X.execute(...)` would be changed to:

```solidity
function execute(
        uint256 operation,
        address to,
        uint256 value,
        bytes memory data
    ) public payable virtual override onlyOwner returns (bytes memory) {
        require(address(this).balance >= value, "ERC725X: insufficient balance");
        _execute(operation, to, value, data);
    }
```

So afterwards, we can refactor in LSP0 and LSP9 to:

```solidity
function execute(
        uint256 operation,
        address to,
        uint256 value,
        bytes memory data
    ) public payable virtual override onlyOwner returns (bytes memory) {
        require(address(this).balance >= value, "ERC725X: insufficient balance");
        if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
        _execute(operation, to, value, data);
    }
```

contracts/LSP9Vault/LSP9Vault.sol
```diff
@@ -30,7 +30,7 @@ contract LSP9Vault is LSP9VaultCore {
      * @notice Sets the owner of the contract and sets the SupportedStandards:LSP9Vault key
      * @param newOwner the owner of the contract
      */
-    constructor(address newOwner) {
+    constructor(address newOwner) payable {
```
CJ42, line 33: Same for check-effect-interaction pattern, I would move this above as the first thing:

```solidity
constructor(address newOwner) payable {
        if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
        
        OwnableUnset._setOwner(newOwner);

        // set key SupportedStandards:LSP9Vault
        _setData(_LSP9_SUPPORTED_STANDARDS_KEY, _LSP9_SUPPORTED_STANDARDS_VALUE);
        _notifyUniversalReceiver(
            newOwner,
            _TYPEID_LSP14_OwnershipTransferred_RecipientNotification,
            ""
        );
    }
```

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -107,6 +108,7 @@ contract LSP9VaultCore is ERC725XCore, ERC725YCore, LSP14Ownable2Step, ILSP1Univ
         bytes memory data
     ) public payable virtual override onlyOwner returns (bytes memory) {
         require(address(this).balance >= value, "ERC725X: insufficient balance");
+        if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
```
CJ42, line 111: same comment as in LSP0 apply here.

contracts/LSP9Vault/LSP9VaultInitAbstract.sol
```diff
@@ -37,5 +37,6 @@ abstract contract LSP9VaultInitAbstract is Initializable, LSP9VaultCore {
             _TYPEID_LSP14_OwnershipTransferred_RecipientNotification,
             ""
         );
+        if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
```
CJ42, line 40: Same comment in LSP0 Init Abstract apply here.

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -194,6 +195,7 @@ abstract contract LSP0ERC725AccountCore is
 
         returnedValues = abi.encode(resultDefaultDelegate, resultTypeIdDelegate);
         emit UniversalReceiver(msg.sender, msg.value, typeId, receivedData, returnedValues);
+        if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
```
CJ42, line 198: I would move this at the top. The first line in the function.
Same here, because of the check-effect-interaction pattern.

The reason I am suggesting is because I tried running the Solidity Static Analysis from Remix once, and it suggested it. See screenshot below:

Screenshot 2022-10-06 at 08 31 59

Screenshot 2022-10-06 at 08 30 17

As suggested, emitting an event is not a big effect (compare to editing something in the contract storage). But it is still an effect in the sense that it changes the state of the blockchain (writing to the logs).

So for best practices, I would suggest to move it at the top 🙂 (first line in the function).

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
b00ste: regarding the suggested changes, the gas used for calling the method will be bigger because it will call `onlyOwner` modifier twice. I would suggest keeping it like this even if the bytecode is bigger but with a cheaper execution and change it when we make the internal `_execute()`.

YamenMerhi: Agree with @CJ42 about splitting into _execute function later.
Agree with @b00ste about not using super as we care more about execution cost rather than deployment cost (Also I think it will not increase the gas deployment by alot)

PR #332, opened 2022-10-07
feat: Allows reentrancy in LSP6 only from controller addresses with REENTRANCY permission + new tests

b00ste:

This PR introduces a new modifier in LSP6KeyManagerCore.sol for blocking reentrancy unless the caller contract has the `REENTRANCY` permission.

Additionally it introduces tests for checking that random contracts cannot reenter the execution and that only a controller contract with `REENTRANCY` permission is allowed to reenter during the execution.

This PR also introduces a new file `Reentrancy.test.ts` in which wi test the REENTRANCY permission combined with the following permissions:
- CALL, AllowedCalls
- SETDATA, AllowedERC725YDataKeys
- ADDPERMISSIONS
- CHANGEPERMISSIONS
- ADDUNIVERSALRECEIVERDELEGATE
- CHANGEUNIVERSALRECEIVERDELEGATE

Expected behaviour:
- Reentrant calls from random contracts MUST fail.
- Reentrant calls from the contracts with `REENTRANCY` permission MUST be allowed.


contracts/Helpers/Security/Reentrancy.sol
```diff
@@ -15,7 +19,11 @@ contract Reentrancy {
     }
 
     fallback() external payable {
-        _target.call(_payload);
+        if (!switchFallback) {
```
CJ42, line 22: What does this `!switchFallback` check does? Re-enter only once?

contracts/Custom/LSP6ReentrancyGuard.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @dev Update the status from `_NON_ENTERED` to `_ENTERED` and checks if
+     * the status is `_ENTERED` in order to revert the call unless the caller is the URD address
+     * Used in the beggining of the `nonReentrant` modifier, before the method execution starts
+     */
+    function _nonReentrantBefore(address upAddress) private {
+        // On the first call to nonReentrant, _status will be _NOT_ENTERED
+        if (_status == _ENTERED) {
+            //compare URD address and msg.sender
```
CJ42, line 50: this comment can be removed since it is mentioned in the Natspec comments

contracts/Custom/LSP6ReentrancyGuard.sol
```diff
@@ -0,0 +1,72 @@
...
+    modifier nonReentrant(address upAddress) {
+        _nonReentrantBefore(upAddress);
+        _;
+        _nonReentrantAfter();
+    }
+
+    /**
+     * @dev Update the status from `_NON_ENTERED` to `_ENTERED` and checks if
+     * the status is `_ENTERED` in order to revert the call unless the caller is the URD address
+     * Used in the beggining of the `nonReentrant` modifier, before the method execution starts
```
CJ42, line 45:
```suggestion
     * Used in the beginning of the `nonReentrant` modifier, before the method execution starts
```

contracts/Custom/LSP6ReentrancyGuard.sol
```diff
@@ -0,0 +1,72 @@
...
+    function _nonReentrantBefore(address upAddress) private {
+        // On the first call to nonReentrant, _status will be _NOT_ENTERED
+        if (_status == _ENTERED) {
+            //compare URD address and msg.sender
+
+            address urdAddress = address(
+                bytes20(IERC725Y(upAddress).getData(_LSP1_UNIVERSAL_RECEIVER_DELEGATE_KEY))
+            );
+
+            if (msg.sender != urdAddress) revert ReentrantAddressNotURD();
```
CJ42, line 56: I don't think we should have this revert error like this since we are just making an exception for the URD.

Why don't we keep the same one as OpenZeppelin?

```solidity
        // On the first call to nonReentrant, _status will be _NOT_ENTERED
        require(_status != _ENTERED, "ReentrancyGuard: reentrant call");
```

or we just state the caller address in the custom error to have a bit more info as well about which contract tried to re-enter the call:

```solidity
revert ReentrantCall(address reentrantCaller)
```

or a bit more explicit, following your changes @skimaharvey @b00ste 

```solidity
revert ReentrantCallNotFromURD(address reentrantCaller) // how about this?
revert ReentrantCallNotFromLSP1Delegate(address reentrantCaller) // or this?
```

@YamenMerhi any feedbacks?

CJ42 (commented):
added review comments

contracts/Custom/LSP6ReentrancyGuard.sol
b00ste: `revert ReentrantCallNotFromLSP1Delegate(address reentrantCaller)` i like this one. since we have an allowed address to reenter i think that we have to specify in the error that the reentrant call is not made from the allowed address, not only that it is a reentrant call, because we kind of allow reentrant calls.

YamenMerhi: I don't mind going with `ReentrantCallNotFromLSP1Delegate` but doesn't make much sense for me, as when it will revert it will be a `ReentrantCall` regardless if the URD is allowed or not.

CJ42: That's true, then let's just keep it like OZ but just use a custom error for gas savings.

```
revert ReentrantCall(address reentrantCaller)
```

contracts/Helpers/Security/Reentrancy.sol
skimaharvey: yes because if it does not we would end up with a loop

YamenMerhi (changes requested):
I don't think it should be applied to `msg.sender`, as we're getting the permissions of relayer in case of `executeRelayCall(..)` which doesn't make any sense. Right ? @b00ste @skimaharvey @CJ42 

The modifier should not be applied to `execute(..)` and `executeRelayCall(..)` in this case, it should be `_verifyPermissions(..)` with removing `view` option as in the modifier we are writing to the state. After adding the modifier to the  `_verifyPermissions(..)` function, we pass the address `from` to the modifier, which pass it to the internal function `_nonReentrantBefore(..)`, in this function we replace `msg.sender` with `from`.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -68,6 +68,32 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     address public target;
     mapping(address => mapping(uint256 => uint256)) internal _nonceStore;
 
+    // Booleans are more expensive than uint256 or any type that takes up a full
```
CJ42, line 71: can we add a reference link maybe to the OpenZeppelin contract we took it from?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -106,7 +132,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     /**
      * @inheritdoc ILSP6KeyManager
      */
-    function execute(bytes calldata payload) public payable returns (bytes memory) {
+    function execute(bytes calldata payload) public payable nonReentrant() returns (bytes memory) {
```
CJ42, line 135: parentheses are not necessary if the modifier does not have arguments.

```suggestion
    function execute(bytes calldata payload) public payable nonReentrant returns (bytes memory) {
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -842,5 +869,37 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+    }
+
+    /**
+     * @dev Update the status from `_NON_ENTERED` to `_ENTERED` and checks if
+     * the status is `_ENTERED` in order to revert the call unless the caller is the URD address
+     * Used in the beginning of the `nonReentrant` modifier, before the method execution starts
+     */
+    function _nonReentrantBefore() private {
+        if (_status == _ENTERED) {
+            // get and check the permissions of the caller
```
CJ42, line 886:
```suggestion
            // CHECK the caller has REENTRANCY permission
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -842,5 +869,37 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+    /**
+     * @dev Initialise _status to _NOT_ENTERED.
+     */
+    function _setupLSP6ReentrancyGuard() internal {
+        _status = _NOT_ENTERED;
+    }
+
+    /**
+     * @dev Update the status from `_NON_ENTERED` to `_ENTERED` and checks if
+     * the status is `_ENTERED` in order to revert the call unless the caller is the URD address
```
CJ42, line 881: this comment is outdated.

```suggestion
     * the status is `_ENTERED` in order to revert the call unless the caller has the REENTRANCY permission
```

tests/LSP6KeyManager/tests/Security.test.ts
```diff
@@ -192,7 +195,11 @@ export const testSecurityScenarios = (
       );
 
       // send LYX to malicious contract
-      await context.keyManager.connect(context.owner).execute(transferPayload);
+      await expect(
+        context.keyManager.connect(context.owner).execute(transferPayload)
+      )
+        .to.be.revertedWithCustomError(context.keyManager, "NotAuthorised")
+        .withArgs(maliciousContract.address, "REENTRANCY");
       // at this point, the malicious contract fallback function
```
CJ42, line 203: can we update this comment and put it back above. So like that:

```js
// send LYX to malicious contract
// at this point, the malicious contract fallback function try to drain funds by re-entering the call
// this should not be possible since it does not have the permission `REENTRANCY`
await expect(
        context.keyManager.connect(context.owner).execute(transferPayload)
      )
        .to.be.revertedWithCustomError(context.keyManager, "NotAuthorised")
        .withArgs(maliciousContract.address, "REENTRANCY");
```

tests/LSP6KeyManager/tests/Security.test.ts
```diff
@@ -272,4 +279,168 @@ export const testSecurityScenarios = (
...
+      const executeTransferPayload = context.keyManager
+        .connect(context.owner)
+        .execute(transferPayload);
+
+      await expect(executeTransferPayload)
+        .to.be.revertedWithCustomError(context.keyManager, "NotAuthorised")
+        .withArgs(maliciousContract.address, "REENTRANCY");
+    });
+
+    it("should pass when reentered by URD", async () => {
```
CJ42, line 312:
```suggestion
    it("should pass when reentered by URD and the URD has REENTRANCY permission", async () => {
```

CJ42 (changes requested):
added review comments

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -68,6 +68,32 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+
+    // The values being non-zero value makes deployment a bit more expensive,
+    // but in exchange the refund on every call to nonReentrant will be lower in
+    // amount. Since refunds are capped to a percentage of the total
+    // transaction's gas, it is best to keep them low in cases like this one, to
+    // increase the likelihood of the full refund coming into effect.
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = 2;
+
+    uint256 private _status;
```
skimaharvey, line 85: I tgink we should rename this variable to `_reantrancyStatus`

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -68,6 +68,23 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+    // Variables, methods and modifier which are used for ReentrancyGuard
+    // are taken from the link below and modified according to our needs.
+    // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = 2;
+    uint256 private _reentrancyStatus;
+
+    /**
+     * @dev This modifier doesn't allow for reentrancy calls unless
+     * it's the URD of the contract that makes the call.
```
skimaharvey, line 80: URD or contract with ReentrancyPermission

skimaharvey (approved):
Nice

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -69,6 +69,23 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+    // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = 2;
+    uint256 private _reentrancyStatus;
+
+    /**
+     * @dev This modifier doesn't allow for reentrancy calls unless the caller
+     * is a controller address with _PERMISSION_REENTRANCY`.
+     */
+    modifier nonReentrant(address from) {
```
YamenMerhi, line 83: Can be removed, since it's no longer used, rigth ?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -188,7 +216,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
      * @param from the address making the request
      * @param payload the payload that will be run on `target`
      */
-    function _verifyPermissions(address from, bytes calldata payload) internal view {
+    function _verifyPermissions(address from, bytes calldata payload) internal {
```
YamenMerhi, line 219: can we re-add view since we didn't change anything here?

contracts/Helpers/Reentrancy/ReentrancyWithValueTransfer.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// interfaces
+import {ILSP6KeyManager} from "../../LSP6KeyManager/ILSP6KeyManager.sol";
+import {LSP14Ownable2Step} from "../../LSP14Ownable2Step/LSP14Ownable2Step.sol";
+
+contract ReentrancyWithValueTransfer {
+    receive() external payable { }
```
YamenMerhi, line 9: linter warning

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -108,9 +125,14 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
      * @inheritdoc ILSP6KeyManager
      */
     function execute(bytes calldata payload) public payable returns (bytes memory) {
+        _nonReentrantBefore(msg.sender);
```
YamenMerhi, line 128: I would move this line under `_verifyPermissions(msg.sender, payload);` to allow proper error bubbling

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -131,6 +153,8 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
 
         address signer = address(this).toDataWithIntendedValidator(encodedMessage).recover(signature);
 
+        _nonReentrantBefore(signer);
```
YamenMerhi, line 156: I would move this line under `_verifyPermissions(signer, payload);` to allow proper error bubbling

YamenMerhi (commented):
For the first round of review, let's see if we agree about changing the location of the check of reentrancy.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: same comment as above

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: also we are fetching permissions in  _nonReentrantBefore(msg.sender) + in the `verifyPermission`

skimaharvey (approved):

YamenMerhi (approved):

YamenMerhi (approved):

contracts/Helpers/Security/Reentrancy.sol
```diff
@@ -15,7 +19,11 @@ contract Reentrancy {
...
+            switchFallback = true;
+            (bool success, bytes memory returnData) = _target.call(_payload);
+            bytes memory result = Address.verifyCallResult(success, returnData, "");
```
CJ42, line 25: a small detail but in this Helper contract, maybe we could add a default revert handler string. Like:

```suggestion
            bytes memory result = Address.verifyCallResult(success, returnData, "Reentrancy Helper Contract: failed to re-enter contract");
```

This way if we add tests in the future, we will know more easily if the reentrant call failed. Otherwise with an empty revert reason string, it might be harder to debug.

contracts/Helpers/UniversalReceivers/UniversalReceiverDelegateDataUpdater.sol
```diff
@@ -0,0 +1,41 @@
...
+        address sender,
+        uint256 value, // solhint-disable no-unused-vars
+        bytes32 typeId,
+        bytes memory data // solhint-disable no-unused-vars
+    ) public virtual returns (bytes memory result) {
+        if (typeId == _TYPEID_LSP7_TOKENSSENDER) {
+            address keyManager = LSP14Ownable2Step(sender).owner();
+            bytes memory setDataPayload = abi.encodeWithSignature(
+                "setData(bytes32,bytes)",
+                bytes32(keccak256(bytes("some random data key"))),
```
CJ42, line 35: The `bytes32` casting is not needed here. `keccak256` will return a `bytes32` value.

```suggestion
                keccak256(bytes("some random data key")),
```

See the Solidity docs:

image

https://docs.soliditylang.org/en/v0.8.17/units-and-global-variables.html#mathematical-and-cryptographic-functions

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -72,6 +72,13 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     address public target;
     mapping(address => mapping(uint256 => uint256)) internal _nonceStore;
 
+    // Variables, methods and modifier which are used for ReentrancyGuard
+    // are taken from the link below and modified according to our needs.
+    // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol
```
CJ42, line 77: I think a frozen commit hash, or a specific release version of OZ might be better here for the link.
In case. If the folder structure of the OpenZeppelin repo changes, the link might break and return a "404 Page Not Found" on Github.

contracts/Helpers/Reentrancy/RelayReentrancy.sol
```diff
@@ -0,0 +1,37 @@
...
+        uint256 newNonce,
+        bytes memory newPayload
+    ) external {
+        _signature = newSignature;
+        _nonce = newNonce;
+        _payload = newPayload;
+    }
+
+    // solhint-disable no-empty-blocks
+    receive() external payable {}
```
CJ42, line 24: is this needed? Does the Helper contract needs to receive funds? 
If not, we can remove it.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -950,5 +967,37 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         if (permission == _PERMISSION_SIGN) return "SIGN";
     }
 
+    /**
+     * @dev Initialise _reentrancyStatus to _NOT_ENTERED.
+     */
+    function _setupLSP6ReentrancyGuard() internal {
```
CJ42, line 973: I am a bit worried about this function being `internal`.
I understand that we use it to setup the `_reentrancyStatus` to `_NOT_ENTERED` **on deployment**.

But now since the function is `internal`, through inheritance, any custom implementation of the KeyManager can recall this function in an other function. Which is not really the expected behaviour, since we just want to **set it up only on deployment**.

I am not sure how we could solve this now. Let me know your feedbacks.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42: > I would move this line under _verifyPermissions(signer, payload); to allow proper error bubbling

@YamenMerhi not sure to understand what you mean by _"allowing proper error bubbling"_. Could you give some details or an example? 🙂 
 
I would leave this line before `_verifyPermissions` actually, because it setup some state in the contract (the `reentrancyStatus` state variable. See below.

image

Since `_verifyPermissions` does a lot of external calls (e.g: reading the permissions to the linked target contract, checking if the contract we want to interact with supports an interface if the payload is `ERC725X.execute(...)`), I think it might be safer to setup the reentrancy status before anything else happen. In case one of the external calls mentioned before is malicious and try to re-enter the KeyManager

(not likely for reading the permissions on the linked target, more more likely when checking with `supportsInterface`, since this can be easily faked as we know).

I think by putting this function here, we follow the _check-effect-interaction_ pattern.

CJ42 (commented):
added review comments

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
YamenMerhi: @CJ42 `_verifyPermissions` is view function, so no risk of adding the line after because any external call will be view and will not change the state of any contract.

I was thinking about reverting with the normal permission first if the caller doesn't have them, then reentrancy can come after. But the team agreed about reverting first with reentrancy, as in both cases, in a case of a caller doesn't have any permission, he will need the permissions to interact and to reenter.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
YamenMerhi: Yep better

contracts/Helpers/Reentrancy/RelayReentrancy.sol
b00ste: Yep, it is needed.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
b00ste: It should be solved right now, the only case when we can initialize is if the status i different to _ENTERED and _NOT_ENERED

CJ42 (approved):

PR #333, opened 2022-10-10
fix: emit `OwnershipTransferStarted` and `Executed` events before external calls

CJ42:

## Overview

In the LSP14Ownable2Step, the function `transferOwnership(...)` performs 3 steps that are similar to the check-effect-interaction pattern.

1. **check** -> validate the required inputs provided as arguments (the `newOwner` is not the address of the contract itself).
2. **effect** -> perform write operations on the contract and blockchain state.
        2.a) set up the  `pendingOwner` as the `newOwner`.
        2.b) emit an event `OwnershipTransferStarted` which write to the blockchain logs.
3. **interactions** -> if the pending owner is a contract that implements the LSP1, notify the pending owner through an **external call** to its `universalReceiver(...)` function.

Written in this order, the steps appear to follow clearly the check-effect-interaction pattern in the right order. However, in the current implementation, **the event `OwnershipTransferStarted` occurs only AFTER the external call happens** ❗ 

> see: https://github.com/lukso-network/lsp-smart-contracts/blob/eb9919fb82414657b721cfb6e70f06659366af52/contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol#L100-L112

The same issue occurs on the `LSP6KeyManagerCore` contract. The `Executed` event gets emitted after the external call. If a re-entrant call occurs on the `execute(...)` / `executeRelayCall(...)` functions _(e.g: the LSP1Delegate contract)_, the 2 x `Executed` events will appear in the logs in the wrong order. The 2nd `Executed` event (from the re-entrant call) will be emitted and appear first.

### Events emitted out of order

The check-effect-interaction pattern often applies to contract state, meaning the 2nd step **effect** often refers to **making necessary updates to the contract state**. 

Events are often not considered as an "effect". However, events are state changing operation. They do not change the state of the contract directly, but they do change the state of the blockchain by writing in the receipt logs.

If any event is emitted in the contract being called, this will lead to an incorrect order of emitted events. 

This might be a low severity issue, but it can lead to undefined behaviour in certain Dapps that listen for events and want to react and interact with the contracts depending on the order of the emitted events.

### Running out of gas

If the external call consumes most of the remaining gas left, if the call the succeed but there is not enough gas left after the external to emit the event, this will lead the tx to fail and revert, although the external call was successful.

## References

This issue was reported in the Liquidity audit report from Trail Of Bits. See page 23.

> https://github.com/trailofbits/publications/blob/master/reviews/Liquity.pdf

image

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -203,22 +203,27 @@ const PERMISSIONS = {
 
 const LSP1_TYPE_IDS = {
 	// keccak256('LSP7Tokens_SenderNotification')
-	LSP7Tokens_SenderNotification: '0x429ac7a06903dbc9c13dfcb3c9d11df8194581fa047c96d7a4171fc7402958ea',
+	LSP7Tokens_SenderNotification:
+		'0x429ac7a06903dbc9c13dfcb3c9d11df8194581fa047c96d7a4171fc7402958ea',
 	// keccak256('LSP7Tokens_RecipientNotification')
-	LSP7Tokens_RecipientNotification: '0x20804611b3e2ea21c480dc465142210acf4a2485947541770ec1fb87dee4a55c',
+	LSP7Tokens_RecipientNotification:
+		'0x20804611b3e2ea21c480dc465142210acf4a2485947541770ec1fb87dee4a55c',
 	// keccak256('LSP8Tokens_SenderNotification')
-	LSP8Tokens_SenderNotification: '0xb23eae7e6d1564b295b4c3e3be402d9a2f0776c57bdf365903496f6fa481ab00',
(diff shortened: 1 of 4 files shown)


CJ42:
Need to do the same for the Executed event in the Key Manager.
(Put the event before the external call).
If someone can do it, otherwise I will do it once I am back on my laptop.

b00ste (approved):

YamenMerhi (approved):

CJ42:
See the incorrect order of events here:
Screenshot 2022-10-13 at 02 37 00

PR #339, opened 2022-10-20
test: add echidna.yml to workflows

skimaharvey:


skimaharvey:
Will go back to it later

PR #340, opened 2022-10-24
fix: [WP-M4] avoid "returnbomb" in ERC165Checker

skimaharvey:
Screenshot 2022-10-24 at 14 11 53

### DIFF
--- a/contracts/Custom/ERC165Checker.sol
+++ b/contracts/Custom/ERC165Checker.sol
@@ -123,8 +123,24 @@ library ERC165Checker {
             IERC165.supportsInterface.selector,
             interfaceId
         );
-        (bool success, bytes memory result) = account.staticcall{gas: 30000}(encodedParams);
-        if (result.length < 32) return false;
-        return success && abi.decode(result, (uint256)) > 0;
+
+        bool success;
+        uint256 returnSize;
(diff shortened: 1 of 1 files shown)


contracts/Custom/ERC165Checker.sol
```diff
@@ -123,8 +123,24 @@ library ERC165Checker {
...
-        if (result.length < 32) return false;
-        return success && abi.decode(result, (uint256)) > 0;
+
+        bool success;
+        uint256 returnSize;
+        uint256 returnValue;
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            success := staticcall(
```
CJ42, line 133: Nice, is it the prettier that automatically indented like that?

Because it looks actually super nice and readable for opcodes that take a lot of arguments like `staticall`.
By having the arguments on each lines, we can see properly through the code **what arguments are positioned/taken from the top of the stack**. 👌 (for instance here `30000` is the gas param).

See this screenshot from evm.codes to understand.

Screenshot 2022-10-24 at 14 01 41

CJ42 (approved):

b00ste (approved):

YamenMerhi (approved):

PR #341, opened 2022-10-24
fix!: [WP-L12] - LSP-6: change parameter of `getNonce(address,uint256)` to a `uint128` -> `getNonce(address,uint128)`

CJ42:

## ⚠️ BREAKING CHANGES

Because of the left bit shifting occurring to generate the idx (= concatenation of channel ID + nonce on a specific channel), the maximum `channelId` that can be provided to the function `getNonce(...)` is the max value of a `uint128`.

Otherwise, the explicit casting will result in wrapping back to 0 and lead to an invalid nonce value for a specific channel ID.

- In the function `getNonce(...)`, change the type of the `channelId` parameter from `uint256` -> `uint128`.
- ⚠️ **update the interface ID of LSP6KeyManager**, since the change in the function signature from `getNonce(address,uint256)` to `getNonce(address,uint128)`.

image

### DIFF
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -72,9 +72,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     /**
      * @inheritdoc ILSP6KeyManager
      */
-    function getNonce(address from, uint256 channelId) public view returns (uint256) {
-        uint128 nonceId = uint128(_nonceStore[from][channelId]);
-        return (uint256(channelId) << 128) | nonceId;
+    function getNonce(address from, uint128 channelId) public view returns (uint256) {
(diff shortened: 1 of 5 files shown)


contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -72,9 +72,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     /**
      * @inheritdoc ILSP6KeyManager
      */
-    function getNonce(address from, uint256 channelId) public view returns (uint256) {
-        uint128 nonceId = uint128(_nonceStore[from][channelId]);
-        return (uint256(channelId) << 128) | nonceId;
+    function getNonce(address from, uint128 channelId) public view returns (uint256) {
+        uint256 nonceInChannel = _nonceStore[from][channelId];
```
skimaharvey, line 76: I like the new name

skimaharvey (approved):

b00ste (approved):

PR #342, opened 2022-10-24
fix: [WP-M9] LSP-2: `LSP2Utils.isEncodedArray()` Incomplete implementation

CJ42:

## 🐛 Bug Fix

The function `isEncodedArray(...)` in `LSP2Utils` behaves incorrectly when the input argument is a bytes32 long value:

e.g:

- any `uint256` value (e.g: **32**, **12345**, ...)
- any `bytes32` value

This causes the function to revert with the following error.

image

This error originates from the `solidity-bytes-utils` library: https://github.com/GNSPS/solidity-bytes-utils/blob/6458fb2780a3092bc756e737f246be1de6d3d362/contracts/BytesLib.sol#L375

This PR fixes this behaviour.

image

image

### DIFF
--- a/contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
+++ b/contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
@@ -202,10 +202,7 @@ library LSP2Utils {
     function isEncodedArray(bytes memory data) internal pure returns (bool) {
         uint256 nbOfBytes = data.length;
 
-        // 1) there must be at least 32 bytes to store the offset
-        if (nbOfBytes < 32) return false;
-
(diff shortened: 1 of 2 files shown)


contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -205,12 +205,9 @@ library LSP2Utils {
         // 1) there must be at least 32 bytes to store the offset
         if (nbOfBytes < 32) return false;
```
skimaharvey, line 206: is this line even needed? @CJ42

b00ste (approved):

YamenMerhi (approved):
LGTM

PR #343, opened 2022-10-24
fix: [WP-M8] LSP7 allow zero-value transfers for `transfer(..)`, `mint(..)` and `burn(..)`

b00ste:

In this PR:
- We allow zero-value transfers, mints and burns in LSP7
- We introduce a check in LSP1 that verifies if the recipient's balance is updated, in order to register a new asset


YamenMerhi (changes requested):
Looks good to me, can we add the same check in URDVault

YamenMerhi (approved):

CJ42:
This is an adjustment in the behaviour of #329

PR #344, opened 2022-10-25
perf: [WP-M24] LSP-6: Recalculate the same `mask` in the for loop is unnecessary

CJ42:

In the Key Manager, the `mask` to calculate the dynamic part of the allowed ERC725Y data key to check is calculated on every iteration in the inner `for` loop (loop that iterates over the input data keys).

Move the mask calculation one level up, outside of the inner `for` loop to save gas. This reduce:

- deployment cost by 626 gas
- runtime cost by 63 gas


b00ste (approved):

skimaharvey (approved):

YamenMerhi (approved):
Looks good to me

PR #348, opened 2022-10-27
chore: [WP-I21] Add comment in helper contracts to mark them for testing only

YamenMerhi:
- [x] Add comment in helper contracts to mark them for testing only.

### DIFF
--- a/contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
+++ b/contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
-// This contract is used only for testing purposes.
-
 // interfaces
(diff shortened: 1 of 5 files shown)


b00ste (approved):

contracts/Helpers/UniversalReceivers/UniversalReceiverTester.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
+// This contract is used only for testing purposes.
```
CJ42, line 4: I think it would be more explicit to mention it just above the contract.

For instance:

```solidity
/**
 * @dev this contract is used only for testing purposes
 */
contract UniversalReceiverTester {
    // ...
}
```

CJ42 (changes requested):

CJ42 (approved):

PR #351, opened 2022-11-01
fix!: [WP-M5] LSP-6: Replace `AllowedStandards/Address/Functions` by `AllowedCalls` - Signature collisions can be exploited with phishing attack

CJ42:

## ⚠️ BREAKING CHANGE!

Combine the following three data keys from LSP6 - Key Manager:

- `AddressPermissions:AllowedStandards:<address>`
- `AddressPermissions:AllowedAddresses:<address>`
- `AddressPermissions:AllowedFunctions:<address>`

...into a single data key:

`AddressPermissions:AllowedCalls:<address>`

## Rationale

image


contracts/LSP2ERC725YJSONSchema/LSP2Errors.sol
```diff
@@ -7,3 +7,10 @@ pragma solidity ^0.8.0;
  * @param valueType the type of the value to check for an abi-encoded array
  */
 error InvalidABIEncodedArray(bytes value, string valueType);
+
+/**
+ * @dev reverts when `value` is not a properly encoded
```
skimaharvey, line 12:
```suggestion
 * @dev reverts when `value` is not a properly encoded compact fixed size bytes array
```

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -274,4 +274,16 @@ library LSP2Utils {
...
+
+    /**
+     * @dev verify that `data` a valid CompactFixedSizeBytesArray
+     * @param compactFixedSizeBytesArray The value that is to be verified
+     * @param fixedSize The size of each element in bytes
+     */
+    function checkValidCompactFixedSizeBytesArray(
+        bytes memory compactFixedSizeBytesArray,
+        uint256 fixedSize
+    ) internal pure returns (bool isValid) {
```
skimaharvey, line 286: we are not using the `isValid` variable

contracts/LSP6KeyManager/LSP6Errors.sol
```diff
@@ -16,27 +16,14 @@ error NoPermissionsSet(address from);
 error NotAuthorised(address from, string permission);
 
 /**
- * @dev reverts when address `from` is not authorised to interact with `disallowedAddress` via the linked account
+ * @dev reverts when `from` is not authorised to interact with `to` address, because of a
```
skimaharvey, line 19: I am not sure about this line because we could be allowed to interact with `to` address just may not be to interact with a specific functions or standard? even though it is explained below I think it is kind of confusing.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -45,6 +45,8 @@ import {
 } from "../LSP0ERC725Account/LSP0Constants.sol";
 import "./LSP6Constants.sol";
 
+import "hardhat/console.sol";
```
skimaharvey, line 48:
```suggestion
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -292,42 +294,17 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
-            bytes12(dataKey) == _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDSTANDARDS_PREFIX ||
-            bytes12(dataKey) == _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDFUNCTIONS_PREFIX
-        ) {
             bool isClearingArray = dataValue.length == 0;
 
-            // AddressPermissions:AllowedFunctions:<address>
-            // AddressPermissions:AllowedStandards:<address>
-            if (!isClearingArray && !LSP2Utils.isBytes4EncodedArray(dataValue)) {
-                revert InvalidABIEncodedArray(dataValue, "bytes4");
+            if (!isClearingArray && !LSP2Utils.checkValidCompactFixedSizeBytesArray(dataValue, 28)) {
```
skimaharvey, line 301: I think `isClearingArray` is not needed because if `dataValue.length == 0` `LSP2Utils.checkValidCompactFixedSizeBytesArray(dataValue, 28)` would return `true` and not revert (which is what we want)

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -553,18 +530,46 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
-            // CHECK for ALLOWED STANDARDS
-            _verifyAllowedStandard(from, to);
+        if (containsFunctionCall) selector = bytes4(payload[164:168]);
 
-            // CHECK for ALLOWED FUNCTIONS
-            // extract bytes4 function selector from payload passed to ERC725X.execute(...)
-            if (payload.length >= 168) _verifyAllowedFunction(from, bytes4(payload[164:168]));
+        uint256 allowedCallsLength = allowedCalls.length;
+
+        // TODO: change behaviour to disallow if nothing is set
```
skimaharvey, line 549: I think we need to discuss with @YamenMerhi and @b00ste if we want to add this revert here since it will revert anyway with `NotAllowedCall` anyway

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -274,4 +274,16 @@ library LSP2Utils {
...
+        if (compactFixedSizeBytesArray.length % (fixedSize + 1) == 0) return true;
+    }
```
skimaharvey, line 288:
```suggestion
    function checkValidCompactFixedSizeBytesArray(
        bytes memory compactFixedSizeBytesArray,
        uint256 fixedSize
    ) internal pure returns (bool) {
        if (compactFixedSizeBytesArray.length % (fixedSize + 1) == 0) return true;
        return false;
    }
```

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -274,4 +274,16 @@ library LSP2Utils {
...
+    /**
+     * @dev verify that `data` a valid CompactFixedSizeBytesArray
+     * @param compactFixedSizeBytesArray The value that is to be verified
+     * @param fixedSize The size of each element in bytes
+     */
+    function checkValidCompactFixedSizeBytesArray(
+        bytes memory compactFixedSizeBytesArray,
+        uint256 fixedSize
+    ) internal pure returns (bool isValid) {
+        if (compactFixedSizeBytesArray.length % (fixedSize + 1) == 0) return true;
```
b00ste, line 287:
```suggestion
        if (compactFixedSizeBytesArray.length % (fixedSize + 1) == 0) isValid = true;
```

skimaharvey: we are trying to avoid to have too many variable in the stack

contracts/LSP6KeyManager/LSP6Constants.sol
```diff
@@ -19,27 +19,19 @@ bytes16 constant _LSP6KEY_ADDRESSPERMISSIONS_ARRAY_PREFIX = 0xdf30dba06db6a30e65
...
-bytes10 constant _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDADDRESSES_PREFIX = 0x4b80742de2bfc6dd6b3c; // AddressPermissions:AllowedAddresses:<address> --> address[]
-
-// bytes6(keccak256('AddressPermissions')) + bytes4(keccak256('AllowedFunctions')) 
-bytes10 constant _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDFUNCTIONS_PREFIX = 0x4b80742de2bf8efea1e8; // AddressPermissions:AllowedFunctions:<address> --> bytes4[]
-
-// bytes6(keccak256('AddressPermissions')) + bytes4(keccak256('AllowedStandards')) 
-bytes10 constant _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDSTANDARDS_PREFIX = 0x4b80742de2bf3efa94a3; // AddressPermissions:AllowedStandards:<address> --> bytes4[]
-
-// bytes6(keccak256('AddressPermissions')) + bytes4(keccak256('AllowedERC725YKeys')) 
+// bytes6(keccak256('AddressPermissions')) + bytes4(keccak256('AllowedERC725YKeys')) // bytes6(keccak256('AddressPermissions')) + bytes4(keccak256('AllowedERC725YKeys'))
```
YamenMerhi, line 25: Duplicate

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -286,48 +286,18 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
-        ) {
-            bool isClearingArray = dataValue.length == 0;
+        } else if (bytes12(dataKey) == _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDCALLS_PREFIX) {
 
-            // AddressPermissions:AllowedFunctions:<address>
-            // AddressPermissions:AllowedStandards:<address>
-            if (!isClearingArray && !LSP2Utils.isBytes4EncodedArray(dataValue)) {
-                revert InvalidABIEncodedArray(dataValue, "bytes4");
-            }
+            if (dataValue.length % 29 != 0) revert InvalidCompactBytes28Array(dataValue);
```
YamenMerhi, line 296: If this error is used only here, and there is no function in `LSP2Utils.sol` that fires this error, why the error should be importted from LSP2Errors? Why not LSP6Errors.sol ?

CJ42: In my opinion, we should not put it in LSP6Utils because `CompactBytesArray` are a `valueType` that will be specific only to the LSP2 standard (once we add them in the LSP2 specs). That's the reason why I think it should stay in the `LSP2Errors.sol` file although it is not used in the LSP2Utils library directly.

YamenMerhi: What about making the error and the validation function more generic, so instead of doing `datavalue .. % 29 != 0`, we can have a function in LSP2Utils like 
```solidity
function isValidCompactBytesArray(bytes memory bytesArray, uint256 bytesLength) {
    if(bytesArray % bytesLength != 0) revert InvalidCompactBytesArray(bytesArray, bytesLength) 
} 
```
It may be overkil a little bit for LSP6, but makes it more generic for everyone that will use LSP2, and make the error more generic not too specific.

CJ42: This is my suggestion:

```solidity
// for checking AllowedCalls
function isValidCompactBytes28Array(bytes memory bytesArray) internal pure returns (bool) {
        if (bytesArray.length % 29 == 0) return true;
        return false;
}
```

CJ42: @YamenMerhi you are right about this imo:

> If this error is used only here, and there is no function in LSP2Utils.sol that fires this error, why the error should be imported from LSP2Errors? Why not LSP6Errors.sol ?

We talked about it with @skimaharvey and think that best option to match both our arguments would be:

- have the function `isValidCompactBytes28Array(...)` in `LSP2Utils.sol` that returns a `bool` true or false.
- replace the custom error `InvalidCompactBytesArray` -> by `InvalidEncodedAllowedCalls`, and define it in `LSP6Errors.sol`

Let me know what you think :)

YamenMerhi: Agree about the error name and location 👍 

I would still think that a more generic function in LSP2Utils is better, specially that it can be re-used, `bytes28` specific function will not be that commonly used, i would still prefer to have the same logic in a more generic function like: 

```js
function isValidCompactBytesArray(bytes memory bytesArray, uint256 bytesLength) internal pure returns (bool) {
        if (bytesArray.length % bytesLength == 0) return true;
        return false;
}
```

Anyways, it's a small detail and if we don't agree about this i don't see a problem going with your suggestion.

CJ42: Yes a more generic error will be better in the future. As long as the implementation is safe and well tested.

@b00ste do you have a snippet for a generic validation function like this one we are talking about?

skimaharvey: I would also prefer however as we saw earlier with 
```js
function isValidCompactBytesArray(bytes memory bytesArray, uint256 bytesLength) internal pure returns (bool) {
        if (bytesArray.length % bytesLength == 0) return true;
        return false;
}
```
if `byteLength == 0` or `byteLength == 1` then it will always return true with any `bytesArray` even when the CompactBytesArray is not valid. I think it could be a problem, however we should probably create a function that iterates over the fixed bytesArray, check the length then move the point by the length etc

skimaharvey: actually this one should work?
https://github.com/lukso-network/lsp-smart-contracts/blob/f3fc8fc2f3aa36b37ba767e3e27c622a72c65fcc/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol#L470-L499

CJ42: @YamenMerhi @skimaharvey @b00ste once we have #352 merged, I will introduce the check function you created as a replacement.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -312,7 +312,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
 
         } else if (bytes12(dataKey) == _LSP6KEY_ADDRESSPERMISSIONS_ALLOWEDCALLS_PREFIX) {
 
-            if (dataValue.length % 29 != 0) revert InvalidEncodedAllowedCalls(dataValue);
+            if (!LSP2Utils.isCompactBytesArray(dataValue)) revert InvalidEncodedAllowedCalls(dataValue);
```
skimaharvey, line 314: I think we need to reintroduce the `isClearingArray` because it will revert if we try to clear array? (we removed it because the function we used was fine with `0x`)

skimaharvey (approved):
👌

YamenMerhi (approved):

b00ste (approved):

skimaharvey (approved):

PR #352, opened 2022-11-02
refactor!: [WP-C2] Change the verification logic of AllowedERC725YKeys in LSP6

b00ste:

### Main

- Change in how we store the: `AllowedERC725YKeys` in the ERC725Y storage:
      - before: `bytes32[] encoded array`
     - now: `compactedBytesArray` 

- Change in how we check the `AllowedERC725YKeys` in the LSP6KeyManager according to `compactedBytesArray` type.

### Periphery:
- Add new custom error to `constant.js`

## Rational

9A9E65E3-4CF5-4C52-A77F-D288F1EF4B5B

### DIFF
--- a/contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
+++ b/contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 // libraries
 import {LSP6Utils} from "../../LSP6KeyManager/LSP6Utils.sol";
+import {LSP2Utils} from "../../LSP2ERC725YJSONSchema/LSP2Utils.sol";
 
 // modules
 import {ERC725Y} from "@erc725/smart-contracts/contracts/ERC725Y.sol";
@@ -46,24 +47,24 @@ contract KeyManagerInternalTester is LSP6KeyManager {
         pure
         returns (bool)
     {
-        return super._checkValidCompactBytesArray(compactBytesArray);
+        return LSP2Utils.isValidCompactBytesArray(compactBytesArray);
     }
 
     function verifyAllowedERC725YSingleKey(
-        address _from,
+        address from,
         bytes32 inputKey,
         bytes memory allowedERC725YKeysFor
     ) public pure returns (bool) {
-        super._verifyAllowedERC725YSingleKey(_from, inputKey, allowedERC725YKeysFor);
+        super._verifyAllowedERC725YSingleKey(from, inputKey, allowedERC725YKeysFor);
         return true;
     }
 
     function verifyAllowedERC725YKeys(
-        address _from,
-        bytes32[] memory _inputKeys,
+        address from,
+        bytes32[] memory inputKeys,
         bytes memory allowedERC725YKeysCompacted
-    ) public view returns (bool) {
-        super._verifyAllowedERC725YKeys(_from, _inputKeys, allowedERC725YKeysCompacted);
+    ) public pure returns (bool) {
+        super._verifyAllowedERC725YKeys(from, inputKeys, allowedERC725YKeysCompacted);
         return true;
     }
 
--- a/contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
+++ b/contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
@@ -42,12 +42,8 @@ contract KeyManagerInternalTester is LSP6KeyManager {
         super._verifyAllowedFunction(_sender, _function);
     }
 
-    function checkValidCompactBytesArray(bytes memory compactBytesArray)
-        public
-        pure
-        returns (bool)
-    {
-        return LSP2Utils.isValidCompactBytesArray(compactBytesArray);
+    function isCompactBytesArray(bytes memory compactBytesArray) public pure returns (bool) {
+        return LSP2Utils.isCompactBytesArray(compactBytesArray);
     }
 
     function verifyAllowedERC725YSingleKey(
--- a/contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
+++ b/contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
@@ -41,8 +41,30 @@ contract KeyManagerInternalTester is LSP6KeyManager {
         super._verifyAllowedFunction(_sender, _function);
     }
 
-    function verifyAllowedERC725YKeys(address _from, bytes32[] memory _inputKeys) public view {
-        super._verifyAllowedERC725YKeys(_from, _inputKeys);
+    function checkValidCompactBytesArray(bytes memory compactBytesArray)
+        public
+        pure
+        returns (bool)
+    {
+        return super._checkValidCompactBytesArray(compactBytesArray);
+    }
+
+    function verifyAllowedERC725YSingleKey(
+        address _from,
+        bytes32 inputKey,
+        bytes memory allowedERC725YKeysFor
+    ) public pure returns (bool) {
+        super._verifyAllowedERC725YSingleKey(_from, inputKey, allowedERC725YKeysFor);
+        return true;
+    }
+
+    function verifyAllowedERC725YKeys(
+        address _from,
+        bytes32[] memory _inputKeys,
+        bytes memory allowedERC725YKeysCompacted
+    ) public view returns (bool) {
+        super._verifyAllowedERC725YKeys(_from, _inputKeys, allowedERC725YKeysCompacted);
+        return true;
     }
 
     function hasPermission(bytes32 _addressPermission, bytes32 _permissions)
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -84,3 +84,9 @@ error AddressPermissionArrayIndexValueNotAnAddress(bytes32 dataKey, bytes invali
  * @dev reverts if there are no AllowedERC725YDataKeys set for the caller
  */
 error NoERC725YDataKeysAllowed();
+
+/**
+ * @dev reverts when `value` is not encoded properly using the CompactBytesArray
+ * @param value the value to check for an CompactBytesArray
+ */
+error InvalidEncodedAllowedERC725YKeys(bytes value);
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -79,3 +79,13 @@ error InvalidERC725Function(bytes4 invalidFunction);
  * @param invalidValue the invalid value that was attempted to be set under AddressPermissions[index]
  */
 error AddressPermissionArrayIndexValueNotAnAddress(bytes32 dataKey, bytes invalidValue);
+
+/**
+ * @dev reverts if there are no AllowedERC725YDataKeys set for the caller
+ */
+error NoERC725YDataKeysAllowed();
+
+/**
+ * @dev reverts when the AllowedERC725YDataKeys are not encoded properly using the CompactBytesArray
+ */
+error InvalidCompactedAllowedERC725YDataKeys();
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -84,8 +84,3 @@ error AddressPermissionArrayIndexValueNotAnAddress(bytes32 dataKey, bytes invali
  * @dev reverts if there are no AllowedERC725YDataKeys set for the caller
  */
 error NoERC725YDataKeysAllowed();
-
-/**
- * @dev reverts when the AllowedERC725YDataKeys are not encoded properly using the CompactBytesArray
- */
-error InvalidCompactedAllowedERC725YDataKeys();
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -82,8 +82,9 @@ error AddressPermissionArrayIndexValueNotAnAddress(bytes32 dataKey, bytes invali
 
 /**
  * @dev reverts if there are no AllowedERC725YDataKeys set for the caller
+ * @param from the address that has no AllowedERC725YDataKeys
  */
-error NoERC725YDataKeysAllowed();
+error NoERC725YDataKeysAllowed(address from);
 
 /**
  * @dev reverts when `value` is not encoded properly using the CompactBytesArray
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -57,6 +57,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     using ECDSA for bytes32;
     using ERC165Checker for address;
     using EIP191Signer for address;
+    using BytesLib for bytes;
 
     address public target;
     mapping(address => mapping(uint256 => uint256)) internal _nonceStore;
@@ -199,11 +200,8 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
                 // CHECK for permission keys
                 _verifyCanSetPermissions(inputKey, inputValue, from, permissions);
 
-            } else {
-                bytes32[] memory wrappedInputKey = new bytes32[](1);
-                wrappedInputKey[0] = inputKey;
-
-                _verifyCanSetData(from, permissions, wrappedInputKey);
+            } else {    
+                _verifyCanSetData(from, permissions, inputKey);
             }
         } else if (erc725Function == SETDATA_ARRAY_SELECTOR) {
             (bytes32[] memory inputKeys, bytes[] memory inputValues) = abi.decode(
@@ -250,6 +248,28 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         } else {
             revert InvalidERC725Function(erc725Function);
         }
+    }
+
+    /**
+     * @dev verify if `_from` has the required permissions to set some dataKeys
+     * on the linked target
+     * @param from the address who want to set the dataKeys
+     * @param permissions the permissions
+     * @param inputKey the dataKeys being set
+     * containing a list of key-value pairs
+     */
+    function _verifyCanSetData(
+        address from,
+        bytes32 permissions,
+        bytes32 inputKey
+    ) internal view {
+        // Skip if caller has SUPER permissions
+        if (permissions.hasPermission(_PERMISSION_SUPER_SETDATA)) return;
+
+        _requirePermissions(from, permissions, _PERMISSION_SETDATA);
+
+        bytes memory allowedERC725YKeysCompacted = ERC725Y(target).getAllowedERC725YKeysFor(from);
+        _verifyAllowedERC725YSingleKey(from, inputKey, allowedERC725YKeysCompacted);
     }
 
     /**
@@ -342,8 +362,8 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
             bool isClearingArray = dataValue.length == 0;
 
             // AddressPermissions:AllowedERC725YKeys:<address>
-            if (!isClearingArray && !LSP2Utils.isEncodedArray(dataValue)) {
-                revert InvalidABIEncodedArray(dataValue, "bytes32");
+            if (!isClearingArray && !_checkValidCompactBytesArray(dataValue)) {
+                revert InvalidCompactedAllowedERC725YDataKeys();
             }
 
             bytes memory storedAllowedERC725YKeys = ERC725Y(target).getData(dataKey);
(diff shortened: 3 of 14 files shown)


contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
```diff
@@ -41,8 +41,30 @@ contract KeyManagerInternalTester is LSP6KeyManager {
         super._verifyAllowedFunction(_sender, _function);
     }
 
-    function verifyAllowedERC725YKeys(address _from, bytes32[] memory _inputKeys) public view {
-        super._verifyAllowedERC725YKeys(_from, _inputKeys);
+    function checkValidCompactBytesArray(bytes memory compactBytesArray)
```
skimaharvey, line 44: I think since it returns a bool pretending by `is` is better?
```suggestion
    function isValidCompactBytesArray(bytes memory compactBytesArray)
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,137 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
 
     /**
-     * @dev verify if `from` is allowed to change the `inputKey`
-     * @param from the address who want to set the dataKeys
-     * @param inputKeys the dataKey that is verified
+     * @dev Verify the validity of the `compactBytesArray` according to LSP2
      */
-    function _verifyAllowedERC725YKeys(address from, bytes32[] memory inputKeys) internal view {
-        bytes memory allowedERC725YKeysEncoded = ERC725Y(target).getAllowedERC725YKeysFor(from);
+    function _checkValidCompactBytesArray(bytes memory compactBytesArray)
```
skimaharvey, line 470:
```suggestion
    function _isValidCompactBytesArray(bytes memory compactBytesArray)
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -342,8 +363,8 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
             bool isClearingArray = dataValue.length == 0;
 
             // AddressPermissions:AllowedERC725YKeys:<address>
-            if (!isClearingArray && !LSP2Utils.isEncodedArray(dataValue)) {
-                revert InvalidABIEncodedArray(dataValue, "bytes32");
+            if (!isClearingArray && !_checkValidCompactBytesArray(dataValue)) {
```
skimaharvey, line 366: I think `isClearingArray` is not needed here. if `dataValue.length == 0` `_checkValidCompactBytesArray(dataValue))` should return true

contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
```diff
@@ -41,8 +41,30 @@ contract KeyManagerInternalTester is LSP6KeyManager {
...
+        public
+        pure
+        returns (bool)
+    {
+        return super._checkValidCompactBytesArray(compactBytesArray);
+    }
+
+    function verifyAllowedERC725YSingleKey(
+        address _from,
```
skimaharvey, line 53: why `_` in front of the from? (not a big deal since it a tester but I don't think we use `_` elsewhere?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: also same as before if `compactBytesArray` is `0x`, it  will be considered as valid. should it? @YamenMerhi @CJ42 @b00ste

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,137 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+         * 03 a00000
+         * 05 fff83a0011
+         * 20 aa0000000000000000000000000000000000000000000000000000000000cafe
+         * 12 bb000000000000000000000000000000beef
+         * 19 cc00000000000000000000000000000000000000000000deed
+         * ↑↑
+         *
+         * the pointer can only land on the length of the following bytes value.
+         */
+        uint256 pointer = 0;
```
skimaharvey, line 523: to me it is a little weird that we don't initialise`foundAllowedKey` but initialize `uint256 pointer = 0;` when we don't need too. Out of consistency I think we should initialize or don't.

tests/utils/helpers.ts
```diff
@@ -127,3 +127,34 @@ export function combinePermissions(..._permissions: string[]) {
...
+      "0x" +
+        compactBytesArray
+          .toString()
+          .substring(pointer + 2, pointer + 2 * (length + 1))
+    );
+
+    pointer += 2 * (length + 1);
+  }
+  return keysToExport;
+}
```
skimaharvey, line 160: this is super cool

contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
YamenMerhi: Yep agree

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
YamenMerhi: I would say let's not initialize, not needed as you said

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
YamenMerhi: Should we move this function to LSP2Utils and rename it to `isValidDynamicCompactBytesArray`?
and then the error could be in sync with the one in PR #351 : `InvalidDynamicCompactBytesArray` and maybe adding the bytes as param ?

CJ42: Yes this function should not be in the KeyManager contract. Since `CompactBytesArray` is something specific to LSP2, not LSP6.

skimaharvey: > Should we move this function to LSP2Utils and rename it to `isValidDynamicCompactBytesArray`? and then the error could be in sync with the one in PR #351 : `InvalidDynamicCompactBytesArray` and maybe adding the bytes as param ?

Yes I think that is what we should do. Also create a similar one for `FixCompactBytesArray`

skimaharvey: actually, we should review this function as it does not entirely work. For example with the value `0x02000000001100`, after the `2` we would expect 2 bytes however in this case we have 4 `00` bytes. At the moment I dont see any use case where it could be a treat but it is not reassuring. @b00ste @CJ42 @YamenMerhi

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,137 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+         * The pointer can only land on the length of the following bytes value.
+         */
+        uint256 pointer = 0;
+
+        /**
+         * Check each length byte and make sure that when you reach the last length byte.
+         * Make sure that the last length describes exactly the last bytes value and you do not get out of bounds.
+         */
+        while (pointer < compactBytesArray.length) {
+            uint256 elementLength = uint256(uint8(bytes1(compactBytesArray[pointer])));
```
skimaharvey, line 495: probably should add a check that returns false when `elementLength == 0` 
```suggestion
            uint256 elementLength = uint256(uint8(bytes1(compactBytesArray[pointer])));
if(elementLength == 0) return false;
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: added a code suggestion following the comment above 
https://github.com/lukso-network/lsp-smart-contracts/pull/352#discussion_r1012520961

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,137 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+     * @dev Verify the validity of the `compactBytesArray` according to LSP2
      */
-    function _verifyAllowedERC725YKeys(address from, bytes32[] memory inputKeys) internal view {
-        bytes memory allowedERC725YKeysEncoded = ERC725Y(target).getAllowedERC725YKeysFor(from);
+    function _checkValidCompactBytesArray(bytes memory compactBytesArray)
+        internal
+        pure
+        returns (bool isValid)
+    {
+        /**
```
skimaharvey, line 475: with current implementation `0x` will be considerer as a valid CompactBytesArray, I don't think it makes sense. Especially if we move this function to LSP2Utils.

```suggestion
        if(compactBytesArray.length == 0) return false;
        /**
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
b00ste: i made this change instead:
`if (elementLength == 0) pointer = compactBytesArray.length + 1;`

YamenMerhi (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,103 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+             *                       vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
+             *        mask = 0xffffff0000000000000000000000000000000000000000000000000000000000
+             */
+            bytes32 mask = bytes32(
+                0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
+            ) << (8 * (32 - length));
+
+            if (allowedKey == (inputKey & mask)) {
+                // voila you found the key ;)
+                foundAllowedKey = true;
```
skimaharvey, line 547: As seens with @CJ42 and @YamenMerhi, here `foundAllowedKey` is not necessary and also should `return` instead so that we don't continue within the loop for nothing. If `allowedKey` is found then we don't need to check the other keys and should just return to save on gas.
I think we should remove the `foundAllowedKey` variable from function and return instead

contracts/LSP6KeyManager/LSP6Errors.sol
```diff
@@ -79,3 +79,8 @@ error InvalidERC725Function(bytes4 invalidFunction);
  * @param invalidValue the invalid value that was attempted to be set under AddressPermissions[index]
  */
 error AddressPermissionArrayIndexValueNotAnAddress(bytes32 dataKey, bytes invalidValue);
+
+/**
+ * @dev reverts if there are no AllowedERC725YDataKeys set for the caller
+ */
+error NoERC725YDataKeysAllowed();
```
CJ42, line 86: is this custom error really necessary?
Why don't we simply re-use the custom error we had before?

image

Since we reverted the logic (= nothing is allowed if nothing is set), now if a caller is given the permission `SETDATA`, it MUST be given some allowed data keys as well to make its permissions work. So if nothing is listed under its allowed ERC725Y data keys and the caller is trying to `setData(...)`, then the data key provided as input is not allowed in the end.

Unless you see a specific reason and argument to states to the caller "you have no data keys allowed".

I guess it depends on the perspective, but it feels a bit strange imo.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -252,6 +250,27 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         }
     }
 
+    /**
+     * @dev verify if `_from` has the required permissions to set some dataKeys
+     * on the linked target
```
CJ42, line 255: Minor but can probably put that on the same line as the Natspec comment above

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -252,6 +250,27 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+        address from,
+        bytes32 permissions,
+        bytes32 inputKey
+    ) internal view {
+        // Skip if caller has SUPER permissions
+        if (permissions.hasPermission(_PERMISSION_SUPER_SETDATA)) return;
+
+        _requirePermissions(from, permissions, _PERMISSION_SETDATA);
+
+        bytes memory allowedERC725YKeysCompacted = ERC725Y(target).getAllowedERC725YKeysFor(from);
```
CJ42, line 270: Very nice variable name with the word "Compacted" in it 👌

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,103 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
-                if (allowedERC725YKeys[ii] == (inputKeys[jj] & mask)) {
-                    // if the input key matches the allowed key
-                    // make it null to mark it as allowed
-                    inputKeys[jj] = bytes32(0);
-                }
+            /*
+             * transform the allowed key situated from `pointer + 1` until `pointer + 1 + length` to a bytes32 value
+             * E.g. 0xfff83a -> 0xfff83a0000000000000000000000000000000000000000000000000000000000
+             */
+            //bytes32 allowedKey = bytes32(allowedERC725YKeysCompacted[pointer + 1:pointer + 1 + length]);
```
CJ42, line 517: this line can be deleted (commented code)

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -270,7 +289,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
 
         _requirePermissions(from, permissions, _PERMISSION_SETDATA);
 
-        _verifyAllowedERC725YKeys(from, inputKeys);
+        bytes memory allowedERC725YKeysCompacted = ERC725Y(target).getAllowedERC725YKeysFor(from);
+
```
CJ42, line 293: maybe we can remove this blank like, to put the variable `allowedERC725YKeysCompacted` and the call to the internal function `_verifyAllowedERC725YKeys(...)` next to each other :)

See the difference in the screenshot below.

image

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,103 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     }
 
     /**
-     * @dev verify if `from` is allowed to change the `inputKey`
-     * @param from the address who want to set the dataKeys
-     * @param inputKeys the dataKey that is verified
+     * @dev Verify if the `inputKey` is present in `alloedERC725KeysCompacted` stored on the `from`'s ERC725Y contract
```
CJ42, line 468:
```suggestion
     * @dev Verify if the `inputKey` is present in `allowedERC725KeysCompacted` stored on the `from`'s ERC725Y contract
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -444,59 +465,103 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
      */
-    function _verifyAllowedERC725YKeys(address from, bytes32[] memory inputKeys) internal view {
-        bytes memory allowedERC725YKeysEncoded = ERC725Y(target).getAllowedERC725YKeysFor(from);
+    function _verifyAllowedERC725YSingleKey(address from, bytes32 inputKey, bytes memory allowedERC725YKeysCompacted) internal pure {
```
CJ42, line 470: We can refactor this function like this:

- we can remove the variable `foundAllowedKey`, and `return` directly as soon as we find a match.
image

- this below can be removed, since we don't allow a single byte length of `0`

image

- If we did not find anything, after the `while` loop, we revert.

```solidity
/**
     * @dev Verify if the `inputKey` is present in `allowedERC725KeysCompacted` stored on the `from`'s ERC725Y contract
     */
    function _verifyAllowedERC725YSingleKey(address from, bytes32 inputKey, bytes memory allowedERC725YKeysCompacted) internal pure {
        if (allowedERC725YKeysCompacted.length == 0) revert NoERC725YDataKeysAllowed();
        if (!LSP2Utils.isValidCompactBytesArray(allowedERC725YKeysCompacted)) revert InvalidCompactBytesArray(allowedERC725YKeysCompacted);

        /**
         * pointer will always land on these values:
         *
         * ↓↓
         * 03 a00000
         * 05 fff83a0011
         * 20 aa0000000000000000000000000000000000000000000000000000000000cafe
         * 12 bb000000000000000000000000000000beef
         * 19 cc00000000000000000000000000000000000000000000deed
         * ↑↑
         *
         * the pointer can only land on the length of the following bytes value.
         */
        uint256 pointer;

        /**
         * iterate over each key by saving in the `pointer` variable the index for
         * the length of the following key until the `pointer` reaches an undefined value
         *
         * 0x 03 a00000 03 fff83a 20 aa00...00cafe
         *    ↑↑        ↑↑        ↑↑
         *  first  |  second  |  third
         *  length |  length  |  length
         */
        while (!foundAllowedKey && pointer < allowedERC725YKeysCompacted.length) {
            /**
             * save the length of the following allowed key
             * which is saved in `allowedERC725YKeys[pointer]`
             */
            uint256 length = uint256(uint8(bytes1(allowedERC725YKeysCompacted[pointer])));

            /*
             * transform the allowed key situated from `pointer + 1` until `pointer + 1 + length` to a bytes32 value
             * E.g. 0xfff83a -> 0xfff83a0000000000000000000000000000000000000000000000000000000000
             */
            bytes32 allowedKey = bytes32(allowedERC725YKeysCompacted.slice(
                pointer + 1,
                length
            ));

            /**
             * the bitmask discard the last `32 - length` bytes of the input key via ANDing &
             * so to compare only the relevant parts of each ERC725Y keys
             *
             * E.g.:
             *
             * allowed key = 0xa00000
             *
             *      &     compare this part
             *                 vvvvvv
             * checked key = 0xa00000cafecafecafecafecafecafecafe000000000000000000000011223344
             *
             *                                              discard this part
             *                       vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
             *        mask = 0xffffff0000000000000000000000000000000000000000000000000000000000
             */
            bytes32 mask = bytes32(
                0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
            ) << (8 * (32 - length));

            if (allowedKey == (inputKey & mask)) {
                return;
            } else {
                // move the pointer to the first index of the first fixed key
                pointer += length + 1;
            }
        }

        revert NotAllowedERC725YKey(from, inputKey);
    }
```

tests/utils/helpers.ts
```diff
@@ -127,3 +127,34 @@ export function combinePermissions(..._permissions: string[]) {
 
   return ethers.utils.hexZeroPad(result.toHexString(), 32);
 }
+
+export function encodeCompactedBytes(inputKeys: BytesLike[]) {
```
CJ42, line 131: suggesting renaming:

```suggestion
export function encodeCompactBytesArray(inputKeys: BytesLike[]) {
```

tests/utils/fixtures.ts
```diff
@@ -137,7 +137,7 @@ export async function setupProfileWithKeyManagerWithURD(
         EOA.address,
         lsp1universalReceiverDelegateUP.address,
         ALL_PERMISSIONS,
-        ethers.utils.hexZeroPad(PERMISSIONS.SETDATA, 32),
+        ethers.utils.hexZeroPad(PERMISSIONS.SUPER_SETDATA, 32),
```
CJ42, line 140: nice, will help to better benchmark gas costs 👌

tests/LSP6KeyManager/tests/PermissionSetData.test.ts
```diff
@@ -224,7 +271,7 @@ export const shouldBehaveLikePermissionSetData = (
         });
       });
 
-      describe("For address that has permission SETDATA", () => {
+      describe("For address that has permission SETDATA with AlloedERC725YDataKeys", () => {
```
CJ42, line 274:
```suggestion
      describe("For address that has permission SETDATA with AllowedERC725YDataKeys", () => {
```

tests/LSP6KeyManager/tests/PermissionSetData.test.ts
```diff
@@ -638,12 +802,59 @@ export const shouldBehaveLikePermissionSetData = (
...
+      const payload = bobContext.universalProfile.interface.encodeFunctionData(
+        "setData(bytes32,bytes)",
+        [
+          ERC725YKeys.LSP6["AddressPermissions:AllowedERC725YKeys"] +
+            aliceContext.universalProfile.address.substring(2),
+          encodeCompactedBytes([key]),
+        ]
+      );
+      await bobContext.keyManager.connect(bob).execute(payload);
+      // move on with the test
```
CJ42, line 856:
```suggestion

```

CJ42 (changes requested):
added comments

contracts/LSP6KeyManager/LSP6Errors.sol
b00ste: We use this error only when caller has zero AllowedERC725YKeys. I might be misleading to get revert error `NotAllowedERC725YKey(..)` whenever you have nothing allowed. Receiving `NoERC725YDataKeysAllowed(..)` error can be more insightful in some situations.

YamenMerhi: Yep technically in both cases we can use `NotAllowedERC725YKey` because the input key cannot be set either because there is no keys allowed at all, or because the input key is not part of the allowedERC725YKeyCompacted, but i think that in case there is no allowedERC725YCompacted keys in the storage its more precise to point to this case because it's a cause of `NotAllowedERC725YKey`.

If we have this seperation and i got:
- `NotAllowedERC725YKey` i know that there is a list of allowedERC725YKey in the storage of the UP but the input key that i am providing is not part of them.
- `NoAllowedERC725YKey` i know that there is nothing in the `allowedERC725YKey` of the storage of UP. that's why setting the input key was reverted

skimaharvey: yes I think it is disputable, only thing is that for allowed calls permissions we are using the generic `NotAllowedCall` error when there are no allowed calls or when the call is not in the list of allowed calls permission. We could also implement a new custom error `NoCallsAllowed` when the allowed calls permission is empty. 
I am fine with both but I think we should be consistent with the behaviors between allowed calls and allowed keys. If we implement `NoAllowedERC725YKey` we should as well implement `NoAllowedCalls`

skimaharvey: when introducing a specific custom error for when the permission is empty: 
cons: 
- more bytecode at deployment
pro: 
- more precise and also I believe more efficient at runtime since we are not returning any params

skimaharvey (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -176,7 +177,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
     }
 
     /**
-     * @dev verify the permissions of the _from address that want to interact with the `target`
+     * @dev verify the permissions of the _from abddress that want to interact with the `target`
```
CJ42, line 180:
```suggestion
     * @dev verify the permissions of the _from address that want to interact with the `target`
```

tests/LSP6KeyManager/tests/PermissionSetData.test.ts
```diff
@@ -441,9 +602,20 @@ export const shouldBehaveLikePermissionSetData = (
           context.owner.address.substring(2),
         ERC725YKeys.LSP6["AddressPermissions:Permissions"] +
           contractCanSetData.address.substring(2),
+        ERC725YKeys.LSP6["AddressPermissions:AllowedERC725YKeys"] +
+          contractCanSetData.address.substring(2),
       ];
 
-      const permissionValues = [ALL_PERMISSIONS, PERMISSIONS.SETDATA];
+      const compactedAllowedERC725YDataKeys = encodeCompactBytesArray([
+        "0x562d53c1631c0c1620e183763f5f6356addcf78f26cbbd0b9eb7061d7c897ea1",
```
CJ42, line 610: name this variable `hardcodedDataKey` (some that we know it comes from a hardcoded value in the Solidity contract.

tests/LSP6KeyManager/internals/AllowedERC725YKeys.internal.ts
```diff
@@ -1,168 +1,728 @@
...
+        dataKeys.fourthDynamicKey.length.toString().substring(2) +
+        dataKeys.fourthDynamicKey.key.toString().substring(2) +
+        dataKeys.thirdFixedKey.length.toString().substring(2) +
+        dataKeys.thirdFixedKey.key.toString().substring(2);
+    });
 
-  describe("keyType: Singleton", () => {
-    let controllerCanSetOneKey: SignerWithAddress;
+    describe("`isCompactBytesArray(..)`", () => {
+      it("Verifying the validity of a CompactBytesArray containing 2 dynamic keys", async () => {
```
CJ42, line 117: Please explain what it should return (`true` or `false`).

```suggestion
      it("should return true for a CompactBytesArray containing 2 dynamic keys", async () => {
```

tests/LSP6KeyManager/internals/AllowedERC725YKeys.internal.ts
```diff
@@ -1,168 +1,728 @@
...
+        thirdFixedKey: {
+          length: "0x20",
+          key: "0x7b68176ed8c5774f16c8040df3f2c4aac9959612242613785716c4263670fe18",
+        },
+        fourthFixedKey: {
+          length: "0x20",
+          key: "0x2b58178172d258515ef1d9e7c467f6f6a09510e863ef5ad383dbfc50721183df",
+        },
+      };
+      dataKeysToCheck = [
```
CJ42, line 68: Not so important, but maybe to clarify:

```suggestion
      dataKeysToRevert = [
```

tests/LSP6KeyManager/tests/PermissionChangeAddPermissions.test.ts
```diff
@@ -3006,7 +2974,7 @@ export const shouldBehaveLikePermissionChangeOrAddPermissions = (
       });
 
       describe("when caller is an address with permission SETDATA + CHANGEPERMISSIONS", () => {
```
CJ42, line 2976:
```suggestion
      describe("when caller is an address with permission SETDATA + CHANGEPERMISSIONS + no Allowed ERC725Y Keys", () => {
```

CJ42 (commented):
added 2nd round of review comments

CJ42 (changes requested):

CJ42 (approved):

tests/LSP6KeyManager/tests/PermissionChangeAddPermissions.test.ts
```diff
@@ -2860,7 +2825,7 @@ export const shouldBehaveLikePermissionChangeOrAddPermissions = (
       });
 
       describe("when caller is an address with permission SETDATA + ADDPERMISSIONS", () => {
-        it("(should pass): 2 x keys + add 2 x new permissions + increment AddressPermissions[].length by +2", async () => {
+        it("(should fail): 2 x keys + add 2 x new permissions + increment AddressPermissions[].length by +2", async () => {
```
CJ42, line 2828: this test should not have been changed, but re-adapted in the`before` setup. As now, we do not have any tests to test if it works.

tests/LSP6KeyManager/tests/PermissionSetData.test.ts
```diff
@@ -95,14 +115,43 @@ export const shouldBehaveLikePermissionSetData = (
...
+            .connect(canSetDataWithAllowedERC725YDataKeys)
+            .execute(payload);
           const fetchedResult = await context.universalProfile.callStatic[
             "getData(bytes32)"
           ](key);
           expect(fetchedResult).to.equal(value);
         });
       });
 
+      describe("For address that has permission SETDATA without any AllowedERC725YDataKeys", () => {
```
CJ42, line 128: nice perfect 👌

contracts/Helpers/KeyManager/KeyManagerInternalsTester.sol
```diff
@@ -41,8 +42,26 @@ contract KeyManagerInternalTester is LSP6KeyManager {
...
+        return LSP2Utils.isCompactBytesArray(compactBytesArray);
+    }
+
+    function verifyAllowedERC725YSingleKey(
+        address from,
+        bytes32 inputKey,
+        bytes memory allowedERC725YKeysFor
+    ) public pure returns (bool) {
+        super._verifyAllowedERC725YSingleKey(from, inputKey, allowedERC725YKeysFor);
+        return true;
```
CJ42, line 55: @b00ste I missed this change while reviewing. This should have never been changed like this.

The point of this file and functions is to expose the `internal` functions publicly so they can be tested via unit tests.
Neither the `verifyAllowedERC725YSingleKey`, nor the `verifyAllowedERC725YKeys` functions return a `boolean` inside the KeyManagerCore.

We should simply test if the check pass (= don't revert), or fail (= revert).

These functions (and their unit tests in TS) **must be exactly the same as it is in the Key Manager Core.**

PR #353, opened 2022-11-04
chore: remove return parameters in LSP2Utils.sol

CJ42:


skimaharvey (approved):
also prefer using return statements

YamenMerhi (approved):

PR #354, opened 2022-11-07
fix: [WP-M7] LSP-6: `CALL` permission with empty `AllowedAddresses` and `AllowedFunctions` should not be equal to `SUPER_CALL`

CJ42:

## ⚠️ BREAKING CHANGE

Change the behaviour of the permissions `TRANSFER_VALUE`, `CALL` and `STATICCALL` to not allow any interaction if nothing is set under `AllowedCall`.


CJ42:
closed because of issue during rebase. Reopened in #356

PR #356, opened 2022-11-07
fix!: [WP-M7] LSP-6: do not allow any interaction if `AllowedCalls` is empty for a caller/signer

CJ42:

## ⚠️ BREAKING CHANGE

Change the behaviour of the permissions `TRANSFER_VALUE`, `CALL` and `STATICCALL` to not allow any interaction if nothing is set under `AllowedCall`.

image

image

image


skimaharvey (approved):
nice

skimaharvey (approved):

PR #357, opened 2022-11-07
refactor: optimise `_verifyAllowedERC725YKeys` and properly test bytes32(0) key

b00ste:

- Optimising `_verifyAllowedERC725YKeys` so it is more gas efficient.
- Disallow `bytes32(0)` data key to be set a data value through the LSP6KeyManager, for security reasons.
- Properly test that one is not allowed to use `setData(bytes32,bytes)` & `setData(bytes32[],bytes[])` if one of the input data keys is equal to `bytes32(0)`.

### DIFF
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -191,6 +191,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         if (erc725Function == SETDATA_SELECTOR) {
             (bytes32 inputKey, bytes memory inputValue) = abi.decode(payload[4:], (bytes32, bytes));
 
+            // We don't allow the datakey: "0x0000000000000000000000000000000000000000000000000000000000000000" to be set
+            if (inputKey == bytes32(0)) revert ZeroDataKeyNotAllowed();
+
             if (bytes16(inputKey) == _LSP6KEY_ADDRESSPERMISSIONS_ARRAY_PREFIX) {
                 // CHECK if key = AddressPermissions[] or AddressPermissions[index]
                 _verifyCanSetPermissionsArray(inputKey, inputValue, from, permissions);
@@ -214,6 +217,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
             for (uint256 ii = 0; ii < inputKeys.length; ii = GasLib.uncheckedIncrement(ii)) {
                 bytes32 key = inputKeys[ii];
                 bytes memory value = inputValues[ii];
+
+                // We don't allow the datakey: "0x0000000000000000000000000000000000000000000000000000000000000000" to be set
+                if (key == bytes32(0)) revert ZeroDataKeyNotAllowed();
 
                 if (bytes16(key) == _LSP6KEY_ADDRESSPERMISSIONS_ARRAY_PREFIX) {
                     // CHECK if key = AddressPermissions[] or AddressPermissions[index]
@@ -424,7 +430,6 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
      * @dev Verify if the `inputKey` is present in `allowedERC725KeysCompacted` stored on the `from`'s ERC725Y contract
      */
     function _verifyAllowedERC725YSingleKey(address from, bytes32 inputKey, bytes memory allowedERC725YKeysCompacted) internal pure {
-        if (inputKey == bytes32(0)) revert NotAllowedERC725YKey(from, bytes32(0)); 
         if (allowedERC725YKeysCompacted.length == 0) revert NoERC725YDataKeysAllowed(from);
         if (!LSP2Utils.isCompactBytesArray(allowedERC725YKeysCompacted)) revert InvalidEncodedAllowedERC725YKeys(allowedERC725YKeysCompacted);
 
@@ -604,13 +609,6 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         for (uint256 i = 0; i < inputKeys.length; i++) {
             if (inputKeys[i] != bytes32(0)) revert NotAllowedERC725YKey(from, inputKeys[i]);
         }
-
-        /**
-         * If you got to this point, it means that `allowedKeysFound` is smaller than 
-         * the number of `inputKeys` and all of the `inputKeys` are bytes32(0).
-         * From that we can derive that one of the `inputKeys` was bytes32(0) from the start.
-         */
-        revert NotAllowedERC725YKey(from, bytes32(0));
     }
 
     /**
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -491,7 +491,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
                 // voila you found the key ;)
                 return;
             } else {
-                // move the pointer to the first index of the first fixed key
+                // move the pointer to the index of the next key
                 pointer += length + 1;
             }
         }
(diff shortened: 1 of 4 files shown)


tests/LSP6KeyManager/tests/AllowedERC725YKeys.test.ts
```diff
@@ -1673,88 +1679,197 @@ export const shouldBehaveLikeAllowedERC725YKeys = (
...
+        {
+          datakeyToSet:
+            "0x0000000000000000000000000000000000000000000000000000000000000000",
+        },
+        {
+          datakeyToSet:
+            "0x00000000000000000000000000000000000000000000000000000000000000ca",
+        },
+      ].forEach((testCase) => {
+        it(`should revert when trying to set any random data key (e.g: ${testCase.datakeyToSet})`, async () => {
```
CJ42, line 1774: Maybe we can encapsulate this in a `describe` block?

```suggestion
describe("should revert when trying to set any random data key that is not in the list of AllowedERC725YKeys")
        it(`e.g: ${testCase.datakeyToSet})`, async () => {
```

tests/LSP6KeyManager/tests/AllowedERC725YKeys.test.ts
```diff
@@ -1673,88 +1679,197 @@ export const shouldBehaveLikeAllowedERC725YKeys = (
...
+            ethers.utils.toUtf8Bytes("some value for " + key)
+          );
+
+          const setDataPayload =
+            context.universalProfile.interface.encodeFunctionData(
+              "setData(bytes32,bytes)",
+              [key, value]
+            );
+
+          if (
```
CJ42, line 1786: Why do we have a `if / else` block inside the tests?
We should not have such conditionals inside the tests imo, as this will end up making it:
- very hard to maintain
- very hard to debug if it fails in the CI.

We should create a separate test here, especially in the case pf the zero data key.

Especially as now, the name of the test states "any random data key", but the data key tested is not random at it is the zero data key.

image

Please put this in a separate test, so that we can easily find it 🙏

tests/LSP6KeyManager/tests/AllowedERC725YKeys.test.ts
```diff
@@ -1769,12 +1884,40 @@ export const shouldBehaveLikeAllowedERC725YKeys = (
...
-            "NotAllowedERC725YKey"
-          )
-          .withArgs(controllerCanSetSomeKeys.address, testCase.datakeyToSet);
+        ).to.be.revertedWithCustomError(
+          context.keyManager,
+          "ZeroDataKeyNotAllowed"
+        );
+      });
+
+      it("should pass when trying to set an array of data keys including bytes32(0)", async () => {
```
CJ42, line 1893: I think there is an error here. It states "should pass" but inside the test, it tests that it reverts.

```suggestion
      it("should revert when trying to set an array of data keys including bytes32(0)", async () => {
```

CJ42 (changes requested):

CJ42:
@b00ste @YamenMerhi @skimaharvey 
We need to describe it very explicitly. Both in the tests + the Solidity file that "**we currently do not allow to set the zero data key**".

But this might change in the future.

Because I do imagine a use case where someone want to:

- set a value for the data key: `0x00000000...00000000`.
- set a value for a data key / or have an allowed ERC725Y data key that starts with a certain number of leading zero bytes.
_example:_ input data key = `0x00000000abcdefabcdef...`, so in the allowed ERC725Y data keys, we have as one of the entry 

```
length
v 
0400000000
  ^^^^^^^^^
  anything starting with `0x00000000`
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -506,8 +512,103 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
      * @param inputKeys the dataKey that is verified
      */
     function _verifyAllowedERC725YKeys(address from, bytes32[] memory inputKeys, bytes memory allowedERC725YKeysCompacted) internal pure {
+        if (allowedERC725YKeysCompacted.length == 0) revert NoERC725YDataKeysAllowed(from);
+        if (!LSP2Utils.isCompactBytesArray(allowedERC725YKeysCompacted)) revert InvalidEncodedAllowedERC725YKeys(allowedERC725YKeysCompacted);
+
```
skimaharvey, line 517: instead of checking the isCompactBytesArray before the while loop, on line 550 we could add a check: 
`if(lengths += length > allowedERC725YKeysCompacted.length) revert InvalidEncodedAllowedERC725YKeys(allowedERC725YKeysCompacted)`

b00ste: you cannot be 100% sure that the loop will iterate over all AllowedERC725YKeys, it might end on the first key and the CompactBytesArray can still be invalid.

skimaharvey: yes but I think this is what we discussed with Fabian this morning. Even when checking the lengths we cannot be sure they are valid

skimaharvey: may be I misunderstood @CJ42 @YamenMerhi

b00ste: in the case of the lsp2 check we have a valid CompactBytesArray, in this case we don’t have even that. i am not sure this is the way to check it

skimaharvey: yes we have a valid `CompactBytesArray` but may be not a valid keys and looping twice of the `allowedERC725YKeysCompacted` is expensive as discussed this morning

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -506,8 +512,103 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+         *  first  |  second  |  third
+         *  length |  length  |  length
+         */
+        while (allowedKeysFound < inputKeys.length && pointer < allowedERC725YKeysCompacted.length) {
+            /**
+             * save the length of the following allowed key
+             * which is saved in `allowedERC725YKeys[pointer]`
+             */
+            uint256 length = uint256(uint8(bytes1(allowedERC725YKeysCompacted[pointer])));
+
```
skimaharvey, line 550: should we add a check here that prevents having a length of over 32 to avoid an overflow error? @b00ste @CJ42 @YamenMerhi

skimaharvey:
> > @b00ste @YamenMerhi @skimaharvey We need to describe it very explicitly. Both in the tests + the Solidity file that "**we currently do not allow to set the zero data key**".
> > ⚠️ But this might change in the future.
> > Because I do imagine a use case where someone want to:
> > 
> > * set a value for the data key: `0x00000000...00000000`.
> > * set a value for a data key / or have an allowed ERC725Y data key that starts with a certain number of leading zero bytes.
> >   _example:_ input data key = `0x00000000abcdefabcdef...`, so in the allowed ERC725Y data keys, we have as one of the entry
> > 
> > ```
> > length
> > v 
> > 0400000000
> >   ^^^^^^^^^
> >   anything starting with `0x00000000`
> > ```
> 
> In the current implementation we can set dynamic keys which are bytes1(0) … bytes31(0), the only one that is not allowed is bytes32(0)

hey @b00ste, where is this check made? with the dynamic bytes can't we set a length of 32 and bytes32(0) as a corresponding bytes?

YamenMerhi:
Can we add a test case where:
- You have under your allowedERC725YKey the bytes32(0) dataKey as allowed key.
- Try to set a random key and see if it passes or not.

b00ste:
> Can we add a test case where:
> 
> * You have under your allowedERC725YKey the bytes32(0) dataKey as allowed key.
> * Try to set a random key and see if it passes or not.

we have test for that already

b00ste:
> > > @b00ste @YamenMerhi @skimaharvey We need to describe it very explicitly. Both in the tests + the Solidity file that "**we currently do not allow to set the zero data key**".
> > > ⚠️ But this might change in the future.
> > > Because I do imagine a use case where someone want to:
> > > 
> > > * set a value for the data key: `0x00000000...00000000`.
> > > * set a value for a data key / or have an allowed ERC725Y data key that starts with a certain number of leading zero bytes.
> > >   _example:_ input data key = `0x00000000abcdefabcdef...`, so in the allowed ERC725Y data keys, we have as one of the entry
> > > 
> > > ```
> > > length
> > > v 
> > > 0400000000
> > >   ^^^^^^^^^
> > >   anything starting with `0x00000000`
> > > ```
> > 
> > 
> > In the current implementation we can set dynamic keys which are bytes1(0) … bytes31(0), the only one that is not allowed is bytes32(0)
> 
> hey @b00ste, where is this check made? with the dynamic bytes can't we set a length of 32 and bytes32(0) as a corresponding bytes?

Line 1770 in the tests, that’s what we do there, we have the bytes32(0) key as an allowed key and we still revert, as discussed. we revert for random keys as well and for bytes31(0) dynamic key because those are not part of AllowedERC725YKeys

YamenMerhi (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: TBD: https://app.clickup.com/t/3ua07je

CJ42 (approved):

PR #359, opened 2022-11-08
refactor!: mark `initialize(...)` functions as `external`

CJ42:

## ⚠️ BREAKING CHANGE

Change the visibility of the `initialize(...)` functions from `public` to `external` to make it behave more like a `constructor` (since constructor are only called externally on deployment).

This also prevent calling the `initialize(...)` internally by mistake when the `Init` contracts are used via inheritance.

Finally, this enables to **"open-up** the visibility/scope of the function via overriding, as it is not possible to "close-down/restrict" the visibility at the moment because of the current Solidity behaviour. See this issue on the OpenZeppelin repository for more details: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3750


b00ste (approved):

YamenMerhi (approved):
LGTM

skimaharvey (approved):

PR #361, opened 2022-11-08
fix: add check for ValueReceived event in LSP0/LSP9

YamenMerhi:
- An additional check made in the receive function to ensure that `msg.value` is greater than 0, then we emit the event.
If not the event will not be emitted and the call should go through.

In solidity, if both `receive` and `fallback` function are implemented, and the contract receive an empty call `{value: 0, data: "0x"}`, the call goes to the `receive` function. 

### Current Behavior

We emit `ValueReceived` in `receive` without checking that msg.value is greater than 0.

### Wanted Behavior

Emit  `ValueReceived` in `receive` only when `msg.value > 0`


skimaharvey (approved):

b00ste (approved):

CJ42 (approved):
Very nice finding! 👌 ⭐

PR #363, opened 2022-11-09
feat: add ADD/CHANGE UniversalReceiverDelegate checks in LSP6

YamenMerhi:
- [x] Adding a check in LSP6 KeyManager to have permission of `ADD/CHANGE UNIVERSALRECEIVERDELEGATE` when adding keys that starts with `bytes10(_LSP1_UNIVERSAL_RECEIVER_DELEGATE_PREFIX)`.

- [x] Remove LSP1 Key from file test "SETDATA" because they are no longer normal ERC725YKey.

- [x] Add test for the new LSP6 Check for `ADD/CHANGE UNIVERSALRECEIVERDELEGATE`

CI will pass after merging #357


YamenMerhi:
Currently the test is failing saying that the `bytes32(0)` is not an allowedKey.

Screen Shot 2022-11-09 at 1 15 09 PM

Knowing that the payload doesn't contain setting a `bytes32(0)` key, the issue is coming from the fact that we are nullify the URD Key and after refactoring `_verifyAllowedERC725YKey(..)` in a previous PR, we didn't add a check that skips the `bytes32(0)` keys. 

The check will be introduced with the PR mentioned in the description, after being merged to develop, we can merge develop to this branch, and the test should automatically pass without adding new logic.

b00ste (approved):

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+        combinePermissions(
+          PERMISSIONS.ADDUNIVERSALRECEIVERDELEGATE,
+          PERMISSIONS.CHANGEUNIVERSALRECEIVERDELEGATE
+        ),
+        PERMISSIONS.ADDUNIVERSALRECEIVERDELEGATE,
+        PERMISSIONS.CHANGEUNIVERSALRECEIVERDELEGATE,
+        PERMISSIONS.SUPER_SETDATA,
+        PERMISSIONS.SETDATA,
+        encodeCompactBytesArray([
+          // Adding the LSP1 Keys as AllowedERC725Ykey to test if it break the behavior
```
CJ42, line 112: nice! 🚀 💪

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+      permissionKeys = permissionKeys.concat(permissionArrayKeys);
+      permissionValues = permissionValues.concat(permissionArrayValues);
+
+      await setupKeyManager(context, permissionKeys, permissionValues);
+    });
+
+    describe("when setting one UniversalReceiverDelegate key", () => {
+      describe("when caller is an address with ALL PERMISSIONS", () => {
+        it("should be allowed to ADD a UniversalReceiverDelegate key", async () => {
+          const payloadParam = {
```
CJ42, line 160: minor but is it necessary to put them in an object?

why not creating them as 2 separate constants?

```js
const dataKey = ERC725YKeys.LSP1.LSP1UniversalReceiverDelegate;
const dataValue = universalReceiverDelegateA.address;
```

So the final `expect` takes only one line.

```js
const result = await context.universalProfile["getData(bytes32)"](dataKey);
expect(ethers.utils.getAddress(result)).to.equal(dataValue);
```

Might take a bit less lines ;)

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+
+          const result = await context.universalProfile["getData(bytes32)"](
+            payloadParam.dataKey
+          );
+          expect(ethers.utils.getAddress(result)).to.equal(
+            payloadParam.dataValue
+          );
+        });
+
+        it("should be allowed to CHANGE a UniversalReceiverDelegate key", async () => {
```
CJ42, line 180: I would suggest for writing the tests that we use the following terms:

- "add"
- "edit"
- "remove"

so that we don't mix up the words "ADD" / "CHANGE" (that are the permissions) with the actions being done (adding, editing, removing).

```suggestion
        it("should be allowed to edit a UniversalReceiverDelegate key", async () => {
```

skimaharvey (approved):
LGTM

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+
+          const result = await context.universalProfile["getData(bytes32)"](
+            payloadParam.dataKey
+          );
+          expect(ethers.utils.getAddress(result)).to.equal(
+            payloadParam.dataValue
+          );
+        });
+
+        it("shouldn't be allowed to CHANGE the UniversalReceiverDelegate key set even if it's setting existing data", async () => {
```
CJ42, line 314:
```suggestion
        it("should NOT be allowed to CHANGE the UniversalReceiverDelegate key set even if it's setting existing data", async () => {
```

>  even if it's setting existing data

what does that mean?

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+            payloadParam.dataKey
+          );
+
+          expect(result).to.equal(payloadParam.dataValue);
+        });
+      });
+
+      describe("when caller is an address with SUPER_SETDATA permission", () => {
+        before(async () => {
+          // Adding a data key by the owner to test if address with SUPER_SETDATA
```
CJ42, line 457: Is it not this instead?

```suggestion
          // Adding a LSP1 data key by the owner to test if address with SUPER_SETDATA
```

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+
+          let payload = context.universalProfile.interface.encodeFunctionData(
+            "setData(bytes32,bytes)",
+            [payloadParam.dataKey, payloadParam.dataValue]
+          );
+
+          await context.keyManager.connect(context.owner).execute(payload);
+        });
+
+        it("shouldn't be allowed to ADD another UniversalReceiverDelegate key even when UniversalReceiverDelegate key is allowed in AllowedERC725YKey", async () => {
```
CJ42, line 552: this is a great test 👌

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+                "NotAuthorised"
+              )
+              .withArgs(
+                canAddAndChangeUniversalReceiverDelegate.address,
+                "SETDATA"
+              );
+          });
+        });
+
+        describe("when adding and changing in same tx UniversalReceiverDelegate", () => {
```
CJ42, line 835: nice this is a very nice edgy case.

Can you test instead in the dataValues to have a different address for the second array entry? Like this.

```js
const payloadParam = {
              dataKeys: [
                ERC725YKeys.LSP1.LSP1UniversalReceiverDelegate,
                ERC725YKeys.LSP1.LSP1UniversalReceiverDelegate,
              ],
              dataValues: [
                universalReceiverDelegateA.address,
                universalReceiverDelegateB.address,
              ],
            };
```

And then the assertion could be:

```
const [result] = await context.universalProfile[
              "getData(bytes32)"
            ](payloadParam.dataKeys[0]);

            expect(result).to.equal(universalReceiverDelegateB.address);
            // or
            expect(result).to.equal(payloadParam.dataValues[1]);
```

So to test that it is set to the last value.

CJ42 (commented):
added review comments

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -203,7 +207,11 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
                 // CHECK for permission keys
                 _verifyCanSetPermissions(inputKey, inputValue, from, permissions);
 
-            } else {
+            } else if(bytes10(inputKey) == _LSP1_UNIVERSAL_RECEIVER_DELEGATE_PREFIX ) {
```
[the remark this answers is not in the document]
YamenMerhi: Good point  👍

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
YamenMerhi: Honestly i like the object to emphasize that these 2 constants are related, like we do in tokens with `txParams` object
also now its taking one because i removed `ethers.utils.getAddress(result)` :)

tests/utils/helpers.ts
```diff
@@ -99,6 +99,16 @@ export function getRandomString() {
   return randoms.join("");
 }
 
+export function getRandomBytes32() {
```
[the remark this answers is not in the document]
YamenMerhi: Nice!

tests/LSP6KeyManager/tests/PermissionsChangeAddURD.test.ts
```diff
@@ -0,0 +1,1353 @@
...
+        ERC725YKeys.LSP1.LSP1UniversalReceiverDelegatePrefix +
+        "0000" +
+        getRandomBytes32().substring(2, 42);
+
+      universalReceiverDelegateKey4 =
+        ERC725YKeys.LSP1.LSP1UniversalReceiverDelegatePrefix +
+        "0000" +
+        getRandomBytes32().substring(2, 42);
+
+      universalReceiverDelegateA = ethers.Wallet.createRandom();
```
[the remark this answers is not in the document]
YamenMerhi: Yep good idea 👍

PR #364, opened 2022-11-10
refactor: add `ContractCreated` event and refactor UniversalFactory contract

YamenMerhi:
- [x] Adding an event emitted whenever a contract is created
- [x] Refactoring the logic of the UniversalFactory to a more precise behavior 
( In case of {deployCreate2} if the contract is initializable the `msg.value` now goes to the external call, not the contract creation )  
- [x] Supporting Custom Errors
- [x] Improving the naming and the nastpec comments

### Old behavior

When deploying a normal contract using `deployCreate2(..)`, wether the contract is initializable or not, the `msg.value` passed was directly sent to the creation of the contract, which means to the constructor that needs to be payable.

And if the contract is initializable, there is no possibility to send value along calling the supposed `initialize(..)` function.
Most valid cases where the contract is initializable, the value needs to be send to the initialize function not the constructor. 

### New behavior

If the normal contract to deploy using `deployCreate2(..)`  is not initializable, we pass the `msg.value` to the contract creation (constructor), if it's intiializable we pass the `msg.value` to the external call that initialize the contract.

> NB: These changes are only valid for normal contract not Clones, as you cannot send `msg.value` to clones on deployment as they don't have a constructor.

### DIFF
--- a/contracts/Factories/UniversalFactory.sol
+++ b/contracts/Factories/UniversalFactory.sol
@@ -34,6 +34,8 @@ error SendingValueNotAllowed();
  * More information: https://weka.medium.com/how-to-send-ether-to-11-440-people-187e332566b7
  */
 contract UniversalFactory {
+    using BytesLib for bytes;
+    
     /**
      * @dev Emitted whenever a contract is created
      * @param contractCreated The address of the contract created
@@ -47,8 +49,6 @@ contract UniversalFactory {
         bool indexed clone,
         bytes initialCalldata
     );
-
-    using BytesLib for bytes;
 
     // The bytecode hash of EIP-1167 Minimal Proxy
     bytes32 private constant _MINIMAL_PROXY_BYTECODE_HASH_PT1 =
--- a/contracts/Factories/UniversalFactory.sol
+++ b/contracts/Factories/UniversalFactory.sol
@@ -49,7 +49,6 @@ contract UniversalFactory {
     );
 
     using BytesLib for bytes;
-    using Address for bool;
 
     // The bytecode hash of EIP-1167 Minimal Proxy
     bytes32 private constant _MINIMAL_PROXY_BYTECODE_HASH_PT1 =
@@ -128,7 +127,11 @@ contract UniversalFactory {
             (bool success, bytes memory returnData) = contractCreated.call{value: msg.value}(
                 initializeCallData
             );
-            success.verifyCallResult(returnData, "UF: could not initialize the created contract");
+            Address.verifyCallResult(
+                success,
+                returnData,
+                "UF: could not initialize the created contract"
+            );
         }
     }
 
@@ -160,7 +163,11 @@ contract UniversalFactory {
             (bool success, bytes memory returnData) = proxy.call{value: msg.value}(
                 initializeCallData
             );
-            success.verifyCallResult(returnData, "UF: could not initialize the created contract");
+            Address.verifyCallResult(
+                success,
+                returnData,
+                "UF: could not initialize the created contract"
+            );
         }
     }
 
--- a/contracts/Factories/UniversalFactory.sol
+++ b/contracts/Factories/UniversalFactory.sol
@@ -7,6 +7,19 @@ import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 
+// errors
+
+/**
+ * @dev reverts when the `byteCode` passed to the {deployCreate2} function is the EIP-1167 Minimal Proxy bytecode
+ */
+error MinimalProxiesDeploymentNotAllowed();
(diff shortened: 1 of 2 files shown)


contracts/Factories/UniversalFactory.sol
```diff
@@ -7,6 +7,19 @@ import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 
+// errors
+
+/**
+ * @dev reverts when the `byteCode` passed to the {deployCreate2} function is the EIP-1167 Minimal Proxy bytecode
+ */
+error MinimalProxiesDeploymentNotAllowed();
```
skimaharvey, line 15: I think some context would be nice on why cant we deploy minimal proxies through `deployCreate2()`? may be should add it over `deployCreate2()` or the `NotMinimalProxy` modifier

contracts/Factories/UniversalFactory.sol
```diff
@@ -84,19 +111,26 @@ contract UniversalFactory {
...
     ) public payable notMinimalProxy(byteCode) returns (address contractCreated) {
-        bytes32 generatedSalt = _generateSalt(initializeCallData, salt);
-        contractCreated = Create2.deploy(msg.value, generatedSalt, byteCode);
+        bytes32 generatedSalt = _generateSalt(initializeCallData, providedSalt);
+
+        if (initializeCallData.length == 0) {
+            contractCreated = Create2.deploy(msg.value, generatedSalt, byteCode);
+            emit ContractCreated(contractCreated, providedSalt, false, initializeCallData);
+        } else {
+            contractCreated = Create2.deploy(0, generatedSalt, byteCode);
```
skimaharvey, line 123: in the event, I send some value through `msg.value` with `initializeCallData.length > 0`. `contractCreated` will be created with 0 value and value will be stuck in Factory?

skimaharvey: I am not sure why we associate the `msg.value` to the `initalizeCallData` and not the `constructor` for example? @YamenMerhi

contracts/Factories/UniversalFactory.sol
```diff
@@ -84,19 +111,26 @@ contract UniversalFactory {
      */
     function deployCreate2(
         bytes memory byteCode,
-        bytes32 salt,
-        bytes memory initializeCallData
+        bytes32 providedSalt,
+        bytes calldata initializeCallData
     ) public payable notMinimalProxy(byteCode) returns (address contractCreated) {
```
CJ42, line 116: What I am not sure to understand here is this modifier.

```
notMinimalProxy(byteCode)
```

So that mean that this function only deploy:
a) contracts with `constructor`
b) contracts used as base implementation behind proxies (like EIP1167).

But if we are deploying base implementation contracts, that are supposed to be initialized/locked directly via the `constructor`, not through a separate `initialize(...)` function call.

Like we do here for instance.

https://github.com/lukso-network/lsp-smart-contracts/blob/99a439f0176b53a0298d964f9d8ae6218ca5a686/contracts/LSP7DigitalAsset/presets/LSP7MintableInit.sol#L11-L18

I am not sure to understand the modifier, or I am missing something here?

contracts/Factories/UniversalFactory.sol
YamenMerhi: I will add them in my next PR, like we agreed :)

contracts/Factories/UniversalFactory.sol
```diff
@@ -84,19 +111,26 @@ contract UniversalFactory {
...
-        bytes32 generatedSalt = _generateSalt(initializeCallData, salt);
-        contractCreated = Create2.deploy(msg.value, generatedSalt, byteCode);
+        bytes32 generatedSalt = _generateSalt(initializeCallData, providedSalt);
+
+        if (initializeCallData.length == 0) {
+            contractCreated = Create2.deploy(msg.value, generatedSalt, byteCode);
```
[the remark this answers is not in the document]
YamenMerhi: Already added :)
```js
      it("should revert when sending value while deploying a non payable constructor contract", async () => {
        let salt = ethers.utils.solidityKeccak256(["string"], ["OtherSalt"]);

        let KMBytecode =
          LSP6KeyManagerBytecode +
          AddressOffset +
          ethers.constants.AddressZero.substr(2);

        await expect(
          context.universalFactory.deployCreate2(KMBytecode, salt, "0x", {
            value: 100,
          })
        ).to.be.revertedWith("Create2: Failed on deploy");
      });
```

contracts/Factories/UniversalFactory.sol
YamenMerhi: Will add more context on this in my next PR, as we agreed :)

CJ42 (approved):

b00ste (approved):

skimaharvey (approved):

PR #366, opened 2022-11-11
refactor: Add a way to differentiate LSP0 from LSP9 inside the `universalReceiverDelegate(..)` of the LSP1

b00ste:

In this PR we introduce a check inside LSP1’s function, `universalReceiverDelegate(..)` that checks for the LSP9 interface id in order to differentiate between ownership transfers of an LSP9 and an LSP0

### DIFF
--- a/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
@@ -64,6 +64,7 @@ contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiverDelega
         (address keyManager, bool ownerIsKeyManager) = _validateCallerViaKeyManager();
         if (!ownerIsKeyManager) return "LSP1: account owner is not a LSP6KeyManager";
 
+        // if receiving ownership of an LSP0, do not register it as a received vault
         if (mapPrefix == _LSP10_VAULTS_MAP_KEY_PREFIX) {
             if (notifier.code.length != 0) {
                 if (!notifier.supportsERC165Interface(_INTERFACEID_LSP9))
--- a/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
@@ -23,6 +23,7 @@ import {LSP10Utils} from "../../LSP10ReceivedVaults/LSP10Utils.sol";
 import "../LSP1Constants.sol";
 import "../../LSP6KeyManager/LSP6Constants.sol";
 import "../../LSP9Vault/LSP9Constants.sol";
+import "../../LSP10ReceivedVaults/LSP10Constants.sol";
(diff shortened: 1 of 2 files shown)


contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
```diff
@@ -63,6 +64,14 @@ contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiverDelega
         (address keyManager, bool ownerIsKeyManager) = _validateCallerViaKeyManager();
         if (!ownerIsKeyManager) return "LSP1: account owner is not a LSP6KeyManager";
 
+        // if receiving ownership of an LSP0, do not register it as a received vault
```
YamenMerhi, line 67: If the account being transferred doesn't support LSP9, do not register it as a received vault, something like that. Let's not be specific to LSP0 as it could be any other account or type of LSP

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -140,6 +142,51 @@ export const shouldBehaveLikeLSP1Delegate = (
     });
   });
 
+  describe("when testing LSP0-ERC725Account", () => {
```
YamenMerhi, line 145: This test doesn't make much sense, as there is no URD for the account to receive, also not controlled by a KM, so its expected to not have anything registered

YamenMerhi (changes requested):
I suggest to add the test in `LSP1UniversalReceiverDelegateUP.behavior`, where you have already an account with a `URD` and a `KM` setup, you just deploy a new account, `transferOwnership(..)`, `acceptOwnership(..)` from the account to receive, and `execute(..)` the payload through the KM, and then you can expect that nothing has changed or added to the `LSP10Vaults` keys.

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
```diff
@@ -63,6 +64,14 @@ contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiverDelega
...
+        if (mapPrefix == _LSP10_VAULTS_MAP_KEY_PREFIX) {
+            if (notifier.code.length != 0) {
```
CJ42, line 69: How should we handle the case when the notification comes from the LSP9 constructor?

https://github.com/lukso-network/lsp-smart-contracts/blob/d1d19f1e7e4d11558700903002fec67b7cbff47d/contracts/LSP9Vault/LSP9Vault.sol#L33-L46

CJ42 (approved):

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
YamenMerhi: We can't. That's why we wrapped it with `if (notifier.code.length != 0)`

YamenMerhi (approved):

skimaharvey (approved):
LGTM

skimaharvey (approved):
LGTM

PR #367, opened 2022-11-11
feat: introduce batch `execute(bytes[])` and`executeRelayCall(bytes[],uint256[],bytes[])` on the LSP6 Key Manager

CJ42:

## ⭐ Feat

Introduce batch execution functions on the Key Manager to group multiple interactions in one single transactions.
Both batch execution functionalities for direct interactions, or interactions via relayers.

Some use cases scenarios were tested. See below.

### 🧪 Tests for batch `execute(bytes[])`

- [x]  send some LYX to three different addresses.
- [x]  send some native tokens (LYX) + some tokens to the same recipient (= to the same recipient).
- [x]  sending multiple tokens to the same recipient (e.g: 3 different tokens).
- [x]  A Universal Profile:
    - deploy the token contract.
    - mint some tokens for the UP address (for yourself).
    - then `transferBatch(…)` to 3 addresses.
- [ ]  Buy multiple NFTs from a marketplace (let’s say 3x NFTs). The Market place is a UP (e.g: a Brand) and send multiple NFTs to the same UP buyer.
- [x]  A Universal Profile:
    - deploy a Token contract (the EIP1169 proxy)
    - initialize the Token contract (call the `initialize(…)` function).
    - setting the metadata of the token contract.
    
### 🧪 Tests for batch `executeRelayCall(bytes[],uint256[],bytes[])`

- [x]  give the permission to someone to mint, this address mint, then remove him the permission - `executeRelayCall(…)` (= ephemeral permission on one transaction).
    - test that the address does not have the AllowedCalls anymore.
    - test that the address cannot mint anymore afterwards.


contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -108,9 +108,26 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
 
-        return _executePayload(payload);
+    /**
+     * @inheritdoc ILSP6KeyManager
+     */
+    function execute(bytes[] calldata payloads) public payable returns (bytes[] memory) {
+        bytes[] memory results = new bytes[](payloads.length);
+        bool msgValueSent;
+
+        for (uint256 ii = 0; ii <= payloads.length - 1; ii++) {
```
skimaharvey, line 121: why `payloads.length - 1`? also should we not store the calculation in a memory variable instead of doing at each step?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -108,9 +108,26 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+            if (!msgValueSent) {
+                results[ii] = _execute(payloads[ii], msg.value);
```
skimaharvey, line 123: I think it could interesting to add a test where: 
- I call `execute(bytes[] calldata payloads)` with a value 
- one of my payload, will call the execute function of the UP which will call back this function.
I think it could renter with the same`msg.value` ? @CJ42

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
YamenMerhi: cleaner `uint256 ii; ii < payloads.length; ii++`

CJ42: yes this might be a bit cheaper. I can try it out.

YamenMerhi (commented):
Cool, Initially looks good to me! Would be good to add more tests, such as 
- If one payload reverts at first and last it reverts with `executeRelayCall` and `execute`
- Trying to pass an array of same signature would fail in `executeRelayCall`
- If we are sending value and the first payload is setData, it reverts,
- etc ..

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: Yes also we are using the batch `executeRelayCall`func we are using `uint256 ii; ii < payloads.length; ii++`. I think out of consistency we should use the same

contracts/LSP6KeyManager/LSP6Constants.sol
```diff
@@ -54,3 +54,9 @@ bytes32 constant _PERMISSION_SETDATA                          = 0x00000000000000
 bytes32 constant _PERMISSION_ENCRYPT                          = 0x0000000000000000000000000000000000000000000000000000000000080000;
 bytes32 constant _PERMISSION_DECRYPT                          = 0x0000000000000000000000000000000000000000000000000000000000100000;
 bytes32 constant _PERMISSION_SIGN                             = 0x0000000000000000000000000000000000000000000000000000000000200000;
+
+// LSP6 overloaded function selectors
+bytes4 constant _LSP6_EXECUTE_SELECTOR = 0x09c5eabe;
+bytes4 constant _LSP6_EXECUTE_BATCH_SELECTOR = 0x44471415;
+bytes4 constant _LSP6_EXECUTE_RELAYCALL_SELECTOR = 0x902d5fa0;
+bytes4 constant _LSP6_EXECUTE_RELAYCALL_BATCH_SELECTOR = 0x4edb7c01;
```
skimaharvey, line 62: i dont think `_LSP6_EXECUTE_BATCH_SELECTOR`, `_LSP6_EXECUTE_RELAYCALL_SELECTOR` and `_LSP6_EXECUTE_BATCH_SELECTOR` are used anywhere

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -121,6 +138,43 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+    }
+
+    /**
+     * @inheritdoc ILSP6KeyManager
+     */
+    function executeRelayCall(
+        bytes[] memory signatures,
+        uint256[] calldata nonces,
+        bytes[] calldata payloads
+    ) public payable returns (bytes[] memory) {
```
skimaharvey, line 151: do you think we should add a check that all parameters`.length` are equal?
@CJ42 @YamenMerhi @b00ste

skimaharvey: for example I send 3 `signature`, 3 `nonce` and 2 `payload`. Since we look depending on the `payloads.length`, we will not loop over the last `signatures` and `nonces`.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -150,12 +204,12 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
       * @param payload the payload to execute
       * @return bytes the result from calling the target with `_payload`
       */
-     function _executePayload(bytes calldata payload) internal returns (bytes memory) {
+     function _executePayload(bytes calldata payload, uint256 msgValue) internal returns (bytes memory) {
 
         emit Executed(msg.value, bytes4(payload));
```
skimaharvey, line 209: I think we should use `msgValue` instead of `msg.value`?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
b00ste: I think we should verify that all the arrays have the same length as we do in `batchTransfer(..)` in LSP7 & LSP8 out of consistency

b00ste (approved):
I think we should check the array lengths in `executeRelayCall(bytes[],uint256[],bytes[])` otherwise LGTM
Also the test are interesting, maybe we can think of some other interesting scenarios

skimaharvey (approved):
Nice

skimaharvey (changes requested):
see thread on SC channel

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -121,11 +145,56 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+     */
+    function executeRelayCall(
+        bytes[] memory signatures,
+        uint256[] calldata nonces,
+        uint256[] calldata values,
+        bytes[] calldata payloads
+    ) public payable returns (bytes[] memory) {
+        if (signatures.length != nonces.length || nonces.length != payloads.length) {
+            revert BatchExecuteRelayCallParamsLengthMismatch();
+        }
```
skimaharvey, line 162: how about checking the values array length as well?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -121,11 +145,56 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+                revert LSP6BatchInsufficientMsgValue(totalValues, msg.value);
+            }
+
+            results[ii] = _executeRelayCall(signatures[ii], nonces[ii], payloads[ii], values[ii]);
+        }
+
+        if (totalValues < msg.value) {
+            revert LSP6BatchCannotLeaveFundsOnKeyManager(totalValues, msg.value);
+        }
+
```
skimaharvey, line 178: I am not sure why we just simply add up all the values and make sure that they add up?
something like this
```js
        uint256 totalValues;

        for (uint256 ii; ii < payloads.length; ii = GasLib.uncheckedIncrement(ii)) {

            results[ii] = _executeRelayCall(signatures[ii], nonces[ii], payloads[ii], values[ii]);
            totalValues += values[ii]
        }
        if (totalValues != msg.value){revert  CustomError}
   ```

File: contracts/LSP6KeyManager/LSP6KeyManagerCore.sol [hunk 8]
```diff
@@ -121,11 +145,56 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+        return results;
+    }
+
+    function _execute(bytes calldata payload, uint256 msgValue) internal returns (bytes memory) {
```
skimaharvey, line 182: I think we should start with msgValue and the payload

skimaharvey (changes requested):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -108,9 +108,33 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
+            if ((totalValues += values[ii]) > msg.value) {
+                revert LSP6BatchInsufficientMsgValue(totalValues, msg.value);
+            }
+
+            results[ii] = _execute(payloads[ii], values[ii]);
+        }
+
+        if (totalValues < msg.value) {
+            revert LSP6BatchCannotLeaveFundsOnKeyManager(totalValues, msg.value);
+        }
```
skimaharvey, line 135: same comment as: https://github.com/lukso-network/lsp-smart-contracts/pull/367/files#r1031359410

contracts/LSP6KeyManager/LSP6Errors.sol
```diff
@@ -105,3 +105,14 @@ error InvalidWhitelistedCall(address from);
...
 error ZeroDataKeyNotAllowed();
+
+/**
+ * @dev reverts when providing array parameters of different sizes to `executeRelayCall(bytes[],uint256[],bytes[])`
+ */
+error BatchExecuteRelayCallParamsLengthMismatch();
+
+error BatchExecuteParamsLengthMismatch();
+
+error LSP6BatchInsufficientMsgValue(uint256 totalValues, uint256 msgValue);
```
b00ste, line 116: I would be nice to have more or less similar naming for this and the following custom error.
```suggestion
error LSP6BatchInsufficientValueSent(uint256 totalValues, uint256 msgValue);
```

contracts/LSP6KeyManager/LSP6Errors.sol
```diff
@@ -105,3 +105,14 @@ error InvalidWhitelistedCall(address from);
...
+
+error LSP6BatchCannotLeaveFundsOnKeyManager(uint256 totalValues, uint256 msgValue);
```
b00ste, line 118:
```suggestion
error LSP6BatchExcessiveValueSent(uint256 totalValues, uint256 msgValue);
```

YamenMerhi (approved):

b00ste (approved):
LGTM 😎

skimaharvey (approved):

PR #368, opened 2022-11-14
feat!: add ADD/CHANGE Extensions permissions and check in LSP6

YamenMerhi:

- [x] Add new permissions:
         - `ADDEXTENSIONS`
         - `CHANGEEXTENSIONS`
         
ALL PERMISSIONS UPDATED TO:
`0x00000000000000000000000000000000000000000000000000000000003f3fff // except DELEGATECALL + SUPER_DELEGATECALL`

- [x] Adding a check in LSP6 KeyManager to have permission of `ADD/CHANGE Extensions` when adding keys that starts with `bytes12(_LSP0_EXTENSION_HANDLER_PREFIX)`.

- [x] Add test for the new LSP6 Check for `ADD/CHANGE Extensions`


skimaharvey (approved):
LGTM

constants.js
```diff
@@ -69,6 +69,10 @@ const SupportedStandards = {
  * @see https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md
  */
 const ERC725YKeys = {
+	LSP0: {
+		// bytes10(keccak256('ExtensionHandler'))
+		ExtensionHandlerPrefix: '0x21c768241ee86a8d564b',
```
CJ42, line 74: Can we add the 2 x zero bytes `0x0000` separator at then end. So that people that import this constant don't have to add it themselves.

```suggestion
		ExtensionHandlerPrefix: '0x21c768241ee86a8d564b0000',
```

Like we do for LSP5.

https://github.com/lukso-network/lsp-smart-contracts/blob/d1d19f1e7e4d11558700903002fec67b7cbff47d/constants.js#L104

constants.js
```diff
@@ -69,6 +69,10 @@ const SupportedStandards = {
...
+	},
 	LSP1: {
 		// bytes10(keccak256('LSP1UniversalReceiverDelegate'))
 		LSP1UniversalReceiverDelegatePrefix: '0x0cfc51aec37c55a4d0b1',
```
CJ42, line 78: Same here, can we add the 2 x zero bytes `0x0000` separator.

```suggestion
		LSP1UniversalReceiverDelegatePrefix: '0x0cfc51aec37c55a4d0b10000',
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -216,6 +217,10 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
                 // CHECK for Universal Receiver Delegate key
                 _verifyCanSetUniversalReceiverDelegateKey(inputKey, from, permissions);
                 
+            } else if (bytes10(inputKey) == _LSP0_EXTENSION_HANDLER_PREFIX) {
```
CJ42, line 220: I guess this is a LSP2 `Mapping` key type?
If that's the case, we need to cast to `bytes12`.

```suggestion
            } else if (bytes12(inputKey) == _LSP0_EXTENSION_HANDLER_PREFIX) {
```

Same reasons as explained here: https://github.com/lukso-network/lsp-smart-contracts/pull/363#discussion_r1017913820

> NB: I can see it is being done below. 
image

tests/LSP6KeyManager/tests/PermissionChangeAddExtensions.test.ts
```diff
@@ -0,0 +1,1227 @@
...
+
+    // Generate few addresses to be used as dataValue for Extension Handler dataKeys
+    let extensionA, extensionB, extensionC, extensionD;
+
+    before(async () => {
+      context = await buildContext();
+
+      extensionHandlerKey1 =
+        ERC725YKeys.LSP0.ExtensionHandlerPrefix +
+        "0000" +
```
CJ42, line 50: here we can remove this afterwards, once the 2 x zero bytes `0x0000` seperators have been added at the end of the data key in constants.js

CJ42 (changes requested):

b00ste (approved):

CJ42 (approved):

PR #369, opened 2022-11-16
feat!: Add LSP17ContractExtension implementation in LSP0 & LSP9

YamenMerhi:
## ⚠️ BREAKING CHANGE

- Creating the LSP17ContractExtension Folder
- Adding in the folder 4 files:
      - `LSP17Constants.sol` holding the interfaceId of `LSP17Extendable` and `LSP17Extension` and `LSP17Extension` data key prefix
      - `LSP17Extendable.sol` abstract contract containing the logic of the fallback function that calls extension contracts
      - `LSP17Extension.sol` abstract contract containing internal relevant function for an extension
      - `LSP17Utils.sol` library contract to add utility function, like `isExtension(..)`

- Modifying the fallback function of the LSP0 and LSP9 to have the logic of the LSP17Extendable + emitting the `ValueReceived` event in case `msg.value > 0`.

- Add tests for LSP17Behavior

- Add contract extension as helpers to test the LSP17 behavior 

- Updating constants of LSP0 and LSP9 to include the interfaceId of  `LSP17Extendable`. + tests

### The LSP0 and LSP9 fallback function behavior:

**- A)** If `msg.value` > 0 emit `ValueReceived` event.

**- B)** If `msg.data.length  < 4`  don't check for extension, just return

**- C)** If `msg.data.length > 4 `:

        - C-1) Construct the data key relevant for the `msg.sig` received

        - C-2) Check in the storage if there is an address stored under the data key constructed
        
        - C-3) If `addressRetreived == address(0)` return
        
        - C-4) If `addressRetreived != address(0)`:

                 - C-4-1) Append the 20 bytes msg.sender and 32 bytes msg.value to the msg.data
                 
                 - C-4-2) Send the resultant msg.data to the addressRetreived as a .call
                   
                 - C-4-3) Return the returnValue or revert the revert reason


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -66,17 +70,76 @@ abstract contract LSP0ERC725AccountCore is
...
+
+                    // The msg.sender address is shifted to the left by 12 bytes to remove the padding
+                    // Then the address without padding is stored right after the calldata
+                    mstore(calldatasize(), shl(96, caller()))
+
+                    // The msg.value is stored right after the calldata + msg.sender
+                    mstore(add(calldatasize(), 20), callvalue())
+
+                    // Add 52 bytes for the msg.sender and msg.value appended at the end of the calldata
+                    let success := call(gas(), extension, 0, 0, add(calldatasize(), 52), 0, 0)
```
frozeman, line 123: maybe we should forward gas - what we need to execute the rest statements below.

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -77,17 +80,76 @@ contract LSP9VaultCore is ERC725XCore, ERC725YCore, LSP14Ownable2Step, ILSP1Univ
...
+     *
+     * If the msg.data is shorter than 4 bytes or msg.sig is the bytes4(0), do not check for an extension and return
+     *
      * Executed when:
      * - the first 4 bytes of the calldata do not match any publicly callable functions from the contract ABI.
      * - receiving native tokens with some calldata.
      */
     fallback() external payable virtual {
         if (msg.value != 0) emit ValueReceived(msg.sender, msg.value);
+
```
frozeman, line 99: Wouldn't it be better to make LSP17 a utils contract used in both, LSP0 and LSP9?

frozeman (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -66,17 +72,48 @@ abstract contract LSP0ERC725AccountCore is
...
+        address extension = address(bytes20(_getData(extensionHandlerDataKey)));
+
+        return extension;
+    }
+
     /**
      * @dev Emits an event when receiving native tokens
      *
+     * Forwards the call to an extension (address) stored under a _LSP17_FALLBACK_EXTENSIONS_HANDLER_ appended
+     * to a function selector. If there is no extension stored under the data key, return.
```
skimaharvey, line 99: may be "appended by". You use "with" on like 101. I am not sure neither. Probably will need @CJ42 gramarly input

contracts/LSP17FallbackExtensions/LSP17FallbackExtension.sol
```diff
@@ -0,0 +1,75 @@
...
+    function _getExtension(bytes4 functionSelector) internal view virtual returns (address) {
+        // To be overrided
+    }
+
+    /**
+     * Forwards the call to an extension mapped to a function selector. If the address of
+     * the extension is address(0), then return.
+     *
+     * The call to the extension is appended with bytes20 (msg.sender) and bytes32 (msg.value).
+     * Returns the return value on success and revert in case of failure.
```
skimaharvey, line 25: i dont think this is correct? `fallback` functions do not return, also `_fallbackExtension()` doesn't? wdut @YamenMerhi @b00ste @CJ42

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -66,17 +72,48 @@ abstract contract LSP0ERC725AccountCore is
...
+     *
+     * The call to the extension is appended with bytes20 (msg.sender) and bytes32 (msg.value).
+     * Returns the return value on success and revert in case of failure.
```
skimaharvey, line 102: i dont think `fallback()` can return

b00ste (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: You can write the return using inline assembly, like i did

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,85 @@
...
+        return interfaceId == _INTERFACEID_LSP17_EXTENDABLE || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev Returns the extension mapped to a specific function selector
+     * If no extension was found, return the address(0)
+     */
+    function _getExtension(bytes4 functionSelector) internal view virtual returns (address) {
+        // To be overrided
+        // Up to the implementor contract to return an extension based on a function selector
```
skimaharvey, line 32: hmmm I am not sure about this because without this, everything after line 48 wont work. It is a must be overridden.
I saw that you implemented them in LSP0 and LSP9 but may be we should `revert` here just so the devs knows why its extensions don't work.

skimaharvey (approved):
LGTM if we decide to go/go exclusively for `call` 🚀

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
CJ42: Suggestion:

```
Forwards the call to an extension contract (address). This address can be retrieved from the ERC725Y data key-value store using the data key below (function selector appended to the prefix): 

_LSP17_FALLBACK_EXTENSIONS_HANDLER_ + <function-selector>

If there is no extension stored under the data key, return.
```

contracts/Helpers/FallbackExtensions/CheckerExtension.sol
```diff
@@ -0,0 +1,28 @@
...
+    // Checks that the msg.sender + msg.value are appended to the msg.data
+    // Length should be: 4 + 32 + 32 + 20 + 32
+    // 4bytes function selector + 32bytes parameter address with offset + 32bytes parameter uint256
+    // + 20bytes msg.sender appended + 32bytes msg.value appended
+    function checkMsgVariable(address originalMsgSender, uint256 originalMsgValue)
+        public
+        pure
+        returns (bool)
+    {
+        if (msg.data.length != 120) revert();
```
CJ42, line 19: I am not sure to understand this check.

Since the value parameters are `address` and `uint256`, they are direct types. So there should not be any offset in the calldata.

The calldata should be made of just the `bytes4` of the selector + bytes32 words padded for each arguments.

image

image

So the calldata here is 68 bytes, not 120.

```
0xe825d37d000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb9226600000000000000000000000000000000000000000000000000000000000000c8
```

contracts/Helpers/FallbackExtensions/CheckerExtension.sol
```diff
@@ -0,0 +1,28 @@
...
+        if (
+            originalMsgSender !=
+            address(bytes20(msg.data[msg.data.length - 52:msg.data.length - 32]))
```
CJ42, line 22: Here the end of the slice is not necessarily needed, since the `bytes20(...)` casting will cut the first 20 bytes.

```suggestion
            address(bytes20(msg.data[msg.data.length - 52:]))
```

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,87 @@
...
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == _INTERFACEID_LSP17_EXTENDABLE || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev Returns the extension mapped to a specific function selector
+     * If no extension was found, return the address(0)
+     */
+    function _getExtension(bytes4 functionSelector) internal view virtual returns (address) {
```
CJ42, line 30: I am not in favour of this.
This can be easily forgotten.

The best way to ensure that the function is implemented is by not making the compiler compile if the function is not implemented.

We can simply write the function definition and end it with a semicolon, removing the whole function body.

```suggestion
    function _getExtension(bytes4 functionSelector) internal view virtual returns (address);
```

If the contract inherits this contract and the function is not implemented, it will not compile, forcing the developer to implement the function.

We can adjust the comments by putting the comment inside the function above the function definition.

```solidity
/**
     * @dev Returns the extension mapped to a specific function selector
     * If no extension was found, return the address(0)
     * To be overrided.
     * Up to the implementor contract to return an extension based on a function selector
     */
    function _getExtension(bytes4 functionSelector) internal view virtual returns (address);
```

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,87 @@
...
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+
+// constants
+import {_INTERFACEID_LSP17_EXTENDABLE} from "./LSP17Constants.sol";
+
+/**
+ * @title Implementation of the fallback logic according to LSP17ContractExtension
+ * @dev Module to be inherited used to extend the functionality of the parent contract when
+ * calling a function that doesn't exist on the parent contract via forwarding the call
+ * to an extension mapped to the function selector being called, set originaly by the parent contract
```
CJ42, line 14:
```suggestion
 * to an extension mapped to the function selector being called, set originally by the parent contract
```

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,87 @@
...
+ */
+abstract contract LSP17Extendable is ERC165 {
+    // solhint-disable
```
CJ42, line 17: why is this solhint disable comment needed here?

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,87 @@
...
+        // Up to the implementor contract to return an extension based on a function selector
+        // revert was added to make sure that the dev override this function in his code, otherwise revert
+        revert();
+    }
+
+    /**
+     * Forwards the call to an extension mapped to a function selector. If the address of
+     * the extension is address(0), then return.
+     *
+     * The call to the extension is appended with bytes20 (msg.sender) and bytes32 (msg.value).
```
CJ42, line 41: The comment here is incorrect. 

The call to the extension is appended by an `address` and a `uint256` ✅ 
Not a `bytes20` and a `bytes32` ❌ 

The difference here is very important because of the padding. 

Especially between:

- `address` (padded on the left).

Screenshot 2022-11-22 at 10 30 34



- `bytes20` (padded on the right).

image

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,87 @@
...
+     * Returns the return value on success and revert in case of failure.
+     *
+     * If the msg.data is shorter than 4 bytes, do not check for an extension and return
+     */
+    function _fallbackLSP17Extendable() internal virtual {
+        if (msg.data.length < 4) return;
+        // If there is a function selector
+        else {
```
CJ42, line 49: The `else` here is not needed. Can be removed.

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -0,0 +1,87 @@
...
+            address extension = _getExtension(msg.sig);
+
+            // if no extension was found, return
+            if (extension == address(0)) {
+                return;
+            }
+            // if the extension was found, call the extension with the msg.data
+            // appended with bytes20(address) and bytes32(msg.value)
+            else {
```
CJ42, line 58: Same here, the `else` is not needed because the previous statement `return`.

CJ42 (changes requested):
added first round of review

contracts/LSP17ContractExtension/LSP17Extendable.sol
skimaharvey: nice way.

contracts/Helpers/FallbackExtensions/CheckerExtension.sol
YamenMerhi: There is comments already explaining the purpose of the function:
```
// Checks that the msg.sender + msg.value are appended to the msg.data
// Length should be: 4 + 32 + 32 + 20 + 32
```

contracts/Helpers/FallbackExtensions/CheckerExtension.sol
YamenMerhi: Just want to make sure to point that the original msg.sender is from the point x till the point y

contracts/LSP17ContractExtension/LSP17Extendable.sol
YamenMerhi: We don't have padding, we already removed the padding in the fallback function. So it's just bytes20 and bytes32 **appended** to the calldata without any offset or zeropad

contracts/Helpers/ERC165Interfaces.sol
```diff
@@ -143,6 +150,30 @@ contract CalculateLSPInterfaces {
 
         return interfaceId;
     }
+
+    function calculateInterfaceLSP17Extendable() public pure returns (bytes4) {
+        // prettier-ignore
+        bytes4 interfaceId = bytes4(keccak256(abi.encodePacked("LSP17Extendable")));
```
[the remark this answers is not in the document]
YamenMerhi: Yes, same for the LSP17Extension, no public function, just a behavior

CJ42 (approved):

PR #372, opened 2022-11-21
refactor: refactor the UniversalFactory contract to LSP16

YamenMerhi:
- Split the function `deployCreate2(..)` into `deployCreate2(..)` and `deployCreate2Init(..)` to allow more possibilities such as splitting the `msg.value` into the constructor and the initialize function
- Split the tests according to the three functions, and edge cases.
- Improve natspec

> The functions are split to take into consideration simplicity and runtime gas cost, any type of contract can be deployed with 1 function but we'll end up with many if else and ununsed parameters.


.github/workflows/build-lint-test.yml
```diff
@@ -55,7 +55,7 @@ jobs:
             "lsp7",
             "lsp8",
             "lsp9",
-            "universalfactory",
+            "lsp16",
```
CJ42, line 58: good catch 👌

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+    ) public payable returns (address contractCreated) {
+        if (initializeCalldata.length == 0) revert ZeroInitializeCalldataNotAllowed();
+        if (constructorMsgValue + initializeCalldataMsgValue != msg.value)
+            revert InvalidMsgValueDistribution();
+
+        bytes32 generatedSalt = _generateSalt(initializeCalldata, providedSalt);
+        contractCreated = Create2.deploy(constructorMsgValue, generatedSalt, byteCode);
+        emit ContractCreated(contractCreated, providedSalt, true, initializeCalldata);
+
+        (bool success, bytes memory returndata) = contractCreated.call{
```
CJ42, line 134: There is a linter error here.
We can discard it with solhint inline comments.

image

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+        bool initializable = initializeCalldata.length > 0;
+        bytes32 generatedSalt = _generateSalt(initializeCalldata, providedSalt);
+
+        proxy = Clones.cloneDeterministic(baseContract, generatedSalt);
+        emit ContractCreated(proxy, providedSalt, initializable, initializeCalldata);
+
+        if (!initializable) {
+            if (msg.value > 0) revert ValueNotAllowedWithNonInitializableProxies();
+        } else {
+            (bool success, bytes memory returndata) = proxy.call{value: msg.value}(
```
CJ42, line 166: Same here, linter error.

image

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * @dev reverts when msg.value sent to {deployCreate2Init} function is not equal to the sum of the
+ * `initializeCalldataMsgValue` and `constructorMsgValue`
+ */
+error InvalidMsgValueDistribution();
+
+/**
+ * @dev reverts when not passing any initializeCalldata to the {deployCreate2Init} function
+ */
+error ZeroInitializeCalldataNotAllowed();
```
CJ42, line 24: _Suggested renaming options_

```suggestion
error InitializeCalldataRequired();
```

```suggestion
error MissingInitializeCalldata();
```

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * @dev reverts when sending value to the {deployCreate2Proxy} function if the contract being created
+ * is an uninitializable clone.
+ */
+error ValueNotAllowedWithNonInitializableProxies();
+
+/**
+ * @dev UniversalFactory contract can be used to deploy CREATE2 contracts; normal contracts and minimal
```
CJ42, line 33: _Suggested changes_

Very good reference to Nick's article 👌 

```
/**
 * @dev UniversalFactory contract can be used to deploy normal or minimal proxy contracts (EIP-1167) using CREATE2.
 * This gives the ability to deploy the same contract at the same address on different chains.
 * If the contract has a constructor, the arguments will be part of the bytecode
 * If the contract has an `initialize` function, the parameters of this function will be included in
 * the salt to ensure that the parameters of the contract should be the same on each chain.
 *
 * This contract should be deployed using Nick's Method.
 * More information: https://weka.medium.com/how-to-send-ether-to-11-440-people-187e332566b7
 */
```

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
YamenMerhi: Yeah, `InitializeCalldataRequired` is way more better, will do 👍

b00ste (approved):

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
```diff
@@ -0,0 +1,218 @@
...
+        bytes calldata initializeCalldata
+    ) public payable returns (address proxy) {
+        bool initializable = initializeCalldata.length > 0;
+        bytes32 generatedSalt = _generateSalt(initializeCalldata, providedSalt);
+
+        proxy = Clones.cloneDeterministic(baseContract, generatedSalt);
+        emit ContractCreated(proxy, providedSalt, initializable, initializeCalldata);
+
+        if (!initializable) {
+            if (msg.value > 0) revert ValueNotAllowedWithNonInitializableProxies();
```
skimaharvey, line 169: we could use `!=` (more gas efficient)

skimaharvey (approved):
looks great 🚀

CJ42 (approved):

PR #375, opened 2022-11-23
chore: add `indexed` keyword to the `contractCreated` event

YamenMerhi:


b00ste (approved):

contracts/LSP16UniversalFactory/LSP16UniversalFactory.sol
```diff
@@ -48,9 +48,9 @@ contract LSP16UniversalFactory {
      * @param initializeCalldata The bytes provided as initializeCalldata
      */
     event ContractCreated(
-        address contractCreated,
-        bytes32 providedSalt,
-        bool initializable,
+        address indexed contractCreated,
```
CJ42, line 51: I understand the idea of indexing the `contractCreated` address (if the address end up being the same through multiple chain).
But how about the salt? I guess yes it will have to be the same as well to be able to deploy at the same address?

Trying to understand.

YamenMerhi: The provided salt by the deployer should be searchable, as well if the contract is initializable or not, to know how to construct the generated salt.

CJ42 (approved):

PR #355, opened 2022-11-07
build!: upgrade `@erc725/smart-contracts` dependency to 4.0.0

CJ42:
# What does this PR introduce?

## 🏗️ Build

Upgrade dependency of `@erc725/smart-contracts` to version 4.0.0. 
This introduces the new batch `execute(...)` function from ERC725X in the smart contracts ABIs of LSP0ERC725Account and LSP9Vault

## ⚠️ BREAKING CHANGES

The following ERC165 interface IDs have changed:

- `LSP0ERC725Account`: `0xdca05671` -> `0xcf6e8efc`
- `LSP9Vault`: `0xca86ec0f` -> `0xd9483482`

PR #377, opened 2022-11-25
fix: add `ValueReceived` event to batch ERC725X `execute(uint256[],address[],uint256[],bytes[])` functions in LSP0 and LSP9

CJ42:

**In LSP0 and LSP9.**

Emit the `ValueReceived` event when caller send some value while calling the new batch `execute(uint256[],address[],uint256[],bytes[])` function from ERC725X.

This was missed in #355

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -187,6 +187,21 @@ abstract contract LSP0ERC725AccountCore is
     }
 
     /**
+     * @inheritdoc ERC725XCore
+     *
+     * @dev Emits a {ValueReceived} event when receiving native tokens.
+     */
+    function execute(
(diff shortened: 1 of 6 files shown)


YamenMerhi (approved):
Since we are calling execute via `super` onlyOwner modifier will run twice, i don't see that a problem currently, but we can optimize later :)

CJ42:
> Since we are calling execute via `super` onlyOwner modifier will run twice, i don't see that a problem currently, but we can optimize later :)

Oh I see. I did not realise that again. Ok then maybe it is better to revert to how it was before then. @YamenMerhi

CJ42:
@YamenMerhi @b00ste @skimaharvey 
In ERC725XCore, we might be better having the balance check inside the internal `_execute` function instead then, to avoid this duplication.

image

YamenMerhi:
@CJ42 Yeah i don't mind

PR #382, opened 2022-11-28
refactor: prevent calling universalReceiver(...) with value

skimaharvey:
## What does this previous introduce 

Add a check that prevents calling `universalReceiver(...)` function with a value associated


YamenMerhi (approved):
LGTM, but not sure if it make sense to have the param

skimaharvey:
> LGTM, but not sure if it make sense to have the param

Right. No point. Removed it.

CJ42 (approved):

PR #383, opened 2022-11-29
test: Add new tests for the batch `execute(..)` and `executeRelayCall(..)`

b00ste:

In this PR the following paths are tested:
- `execute(bytes)` -> `execute(uint256[],bytes[])`
- `execute(bytes)` -> `executeRelayCall(bytes[],uint256[],uint256[],bytes[])`

Also, this PR removes the REENTRANCY permission from the constant `ALL_PERMISSIONS` inside constants.js


tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -7027,4 +6738,2890 @@ export const testReentrancyScenarios = async (
...
+          context.keyManager.connect(caller)["execute(bytes)"](upExecutePayload)
+        )
+          .to.be.revertedWithCustomError(context.keyManager, "NotAuthorised")
+          .withArgs(
+            signer_with_VALUETRANSFER_permissions_no_calls.address,
+            "REENTRANCY"
+          );
+      });
+
+      it("should revert if the reentrant contract has ONLY TRANSFERVALUE permission with AlowedCalls", async () => {
```
skimaharvey, line 8652: I think in this test. It makes sense that since we are reentering through `executeRelayCall(...)`, we should always mention something about the signer as it is the one that we care about when reentering. For example in this test it is failing with REENTRANCY error because of the signer not the contract.
This comment is valid of tests between line 8258-8697

tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -7027,4 +6738,2890 @@ export const testReentrancyScenarios = async (
...
+        );
+
+        await expect(
+          context.keyManager.connect(caller)["execute(bytes)"](upExecutePayload)
+        )
+          .to.be.revertedWithCustomError(context.keyManager, "NotAuthorised")
+          .withArgs(signer_without_permissions.address, "REENTRANCY");
+      });
+
+      it("should revert if caller has ONLY REENTRANCY permission", async () => {
```
skimaharvey, line 9081: did you mean signer here intead of caller?

tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -7027,4 +6738,2890 @@ export const testReentrancyScenarios = async (
...
+          context.keyManager.connect(caller)["execute(bytes)"](upExecutePayload)
+        )
+          .to.be.revertedWithCustomError(context.keyManager, "NotAuthorised")
+          .withArgs(
+            signer_with_ADDPERMISSION_permissions.address,
+            "REENTRANCY"
+          );
+      });
+
+      it("should pass if caller has ADDPERMISSIONS and REENTRANCY permissions", async () => {
```
skimaharvey, line 9121: call is reentering through `executeRelayCall(...)` I think "signer" is more appropriate than "caller"

skimaharvey (commented):
Test are super solid 🚀  ! Congrats.
Just 2 points: 
- Between line 8258-8697, when reentering through `executeRelayCall(…)`, you are only mentioning context about the contract that reenters but nothing about the signer (which is the one we are checking permission for). I think you should add some context about the signer in each of these tests
- between line 8848 and 9627 you use the term “Caller” instead of the term “signer”, we are reentering through `executeRelayCall(…)` so we are checking signer permission not caller (which you are doing just you are calling it “caller”. I think you should replace it.
 
~~Also, as a side note, when transferring value and having no REENTRANCY permission, it is interesting to note that when reentering we revert with TRANSFERVALUE error over REENTRANCY. Probably nothing we can do about it without major changes but I think it may make more sense to revert with REENTRANCY first (debatable).~~
Anyway good job will probably go back to it. Probably missed some stuff...

tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -201,6 +203,119 @@ const generateValueTransferPayload = async (
...
+      );
+      break;
+    default:
+      upExecutePayload = "0x";
+      break;
+  }
+
+  let latestNonce = await keyManager.callStatic.getNonce(signer.address, 1);
+
+  const signedMessageParams = {
```
CJ42, line 284: Maybe you can reuse this function in the helpers?

https://github.com/lukso-network/lsp-smart-contracts/blob/7a2c3256b327e36ba229cfa94d8a08ad448694db/tests/utils/helpers.ts#L202-L237

tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -500,6 +612,7 @@ export const testReentrancyScenarios = async (
       let contract_with_VALUETRANSFER_permissions_with_calls: ReentrancyWithValueTransfer;
       let contract_with_REENTRANCY_VALUETRANSFER_permissions_no_calls: ReentrancyWithValueTransfer;
       let contract_with_REENTRANCY_VALUETRANSFER_permissions_with_calls: ReentrancyWithValueTransfer;
+      let reentrantCallPayload: BytesLike;
```
CJ42, line 615: could you add a whitespace before the `before` hook for readability please :)

contracts/Helpers/Reentrancy/RelayBatchReentrancy.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+// interfaces
+import {ILSP6KeyManager} from "../../LSP6KeyManager/ILSP6KeyManager.sol";
+import {LSP14Ownable2Step} from "../../LSP14Ownable2Step/LSP14Ownable2Step.sol";
+
+contract RelayBatchReentrancy {
```
CJ42, line 8: It starts to feel like a lot of Helpers contracts for every single new case. 

image

Maybe we can try to make one mock contract for multiple cases. How about if we:
- use one `RelayReentrancy` contract with two functions:
    - `prepareRelayCall`
    - `prepareBatchRelayCall`
    
And in its `universalReceiver` function, we can make it behave to send either the normal one or the batch payload, depending on the `typeId`.

tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -842,6 +901,7 @@ export const testReentrancyScenarios = async (
       let contract_with_SETDATA_permissions_with_allowed_keys: ReentrancyWithSetData;
       let contract_with_REENTRANCY_SETDATA_permissions_no_allowed_keys: ReentrancyWithSetData;
       let contract_with_REENTRANCY_SETDATA_permissions_with_allowed_keys: ReentrancyWithSetData;
+      let reentrantContractPayload: BytesLike;
```
CJ42, line 904: same needing white space before the `before` hook.

CJ42 (commented):
added 1st round of review

contracts/Helpers/Reentrancy/RelayBatchReentrancy.sol
```diff
@@ -0,0 +1,38 @@
...
+        _nonces = newNonces;
+        _values = newValues;
+        _payloads = newPayloads;
+    }
+
+    // solhint-disable no-empty-blocks
+    receive() external payable {}
+
+    function universalReceiver(
+        bytes32 typeId, // solhint-disable no-unused-vars
```
YamenMerhi, line 30: Let's comment the params

contracts/Helpers/Reentrancy/RelayBatchReentrancy.sol
```diff
@@ -0,0 +1,38 @@
...
+        bytes memory data // solhint-disable no-unused-vars
+    ) public virtual returns (bytes memory result) {
+        // solhint-disable no-unused-vars
+        address keyManager = LSP14Ownable2Step(msg.sender).owner();
+
+        ILSP6KeyManager(keyManager).executeRelayCall(_signatures, _nonces, _values, _payloads);
```
YamenMerhi, line 36: Adding a `return` statement before `ILSP6KeyManager ...` to avoid compiler warnings

contracts/Helpers/Reentrancy/RelayBatchReentrancy.sol
b00ste: A soon as Jean's PR will be merged I will merge develop into this PR which will remove the compiler warning.

contracts/Helpers/Reentrancy/RelayBatchReentrancy.sol
b00ste: A soon as Jean's PR will be merged I will merge develop into this PR which will remove the compiler warning.

contracts/Helpers/Reentrancy/RelayBatchReentrancy.sol
b00ste: We could make a single contract that could handle very case  based on a `typeId` but I don't think we should overcomplicate as these are only used for tests and having several different contracts for different cases would make sense as it makes it easier to understand the context of the tests. I would keep them as they are as they give more specifics to the tests.

YamenMerhi (commented):
For the first round, looks good to me.
Let's not forget to remove the `.only` before merging to develop.

Later we can try to wrap the code in some way to avoid having 9k lines as there is shared behavior and scenarios that can be combined in one function maybe.

CJ42 (approved):

PR #384, opened 2022-11-29
Fix Slither error for `transferBatch` in LSP7/8 + remove logic of `_beforeTokenTransfer` hooks

CJ42:

When running Slither against `LSP7DigitalAssetCore` or `LSP8IdentifiableDigitalAssetCore`, the following errors are reported ⬇️ .

Screenshot 2022-11-29 at 15 47 47
Screenshot 2022-11-29 at 15 24 04

In brief, Slither spots some [potential costly operations inside a loop](https://github.com/crytic/slither/wiki/Detector-Documentation#costly-operations-inside-a-loop): the `for` loop of the `transferBatch` function.

This error is a false positive, as the **costly operations inside a loop** here refer to increasing/decreasing the total supply of `_existingTokens`, via the `_beforeTokenTransfer` hook. However, this will never happen as there will not be any token supply increase or decrease on any transferBatch, because neither `from` nor `to` can be `address(0)`. Only the functions `_mint(...)` and `_burn(...)` can increase/decrease the total supply.

## What is the new behaviour?

This Slither error is fixed by making the following changes.

- Inline the increase or decrease of the total supply (`_existingTokens`) inside the internal `_mint` and `_burn` functions.
- Remove any logic inside the `_beforeTokenTransfer` to leave it up to the developer to implement its own custom logic inside the hook.
This follow the same approach used by OpenZeppelin in their [ERC20 implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/8f8fd84f1e60426a5e785d6b5b2524938271bb05/contracts/token/ERC20/ERC20.sol#L364-L368)

See OpenZeppelin docs "Using Hooks" for more details ➡️ https://docs.openzeppelin.com/contracts/4.x/extending-contracts#using-hooks

### DIFF
--- a/contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol
+++ b/contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol
@@ -220,6 +220,9 @@ abstract contract LSP7DigitalAssetCore is ILSP7DigitalAsset {
 
         _beforeTokenTransfer(address(0), to, amount);
 
+        // tokens being minted
+        _existingTokens += amount;
+
         _tokenOwnerBalances[to] += amount;
 
         emit Transfer(operator, address(0), to, amount, force, data);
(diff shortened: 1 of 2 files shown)


YamenMerhi (approved):
Great Catch 👍 Let's resolve the linter warnings and merge 🚀

YamenMerhi (approved):

PR #392, opened 2022-12-07
 feat!: adjust supportsInterface in LSP0/9 to support Interfaces of ERC165Extension

YamenMerhi:
- Adding to LSP0/9 the possibility of supporting an interface through the ERC165Extension.

### Why this changes are introduced?
- Since we implemented Fallback extensions in LSP0, now we can call functions on the LSP0 that are not existing as part of the core code but as extensions, but adding functions as extensions is not enough since some application check InterfaceId before calling. (For example in our `universalReceiver(..)` function we check LSP1 interfaceId before calling the function)

To avoid situations where we implement a new function as extension but we miss supporting the interfaceId, we decided to check if there is more interfaceIds supported in the ERC165Extension whenever someone calls the `supportsInterface(..)` function of the LSP0/LSP9

### Old Behavior

The `supportsInterface(..)` on LSP0 will return true if only called with the interfaceId of `LSP0`, `LSP14`, `LSP17`, `ERC725X`, `ERC725Y` and `ERC1271`.

### New Behavior

If no ERC165Extension is set, the `supportsInterface(..)` on LSP0 will return true if only called with the following interfaceIds: `LSP0`, `LSP14`, `LSP17`, `ERC725X`, `ERC725Y` and `ERC1271`.

If the ERC165Extension is set, the `supportsInterface(..)` on LSP0 will return true if called with the following interfaceIds: `LSP0`, `LSP14`, `LSP17`, `ERC725X`, `ERC725Y` and `ERC1271` **and** if the interfaceId being checked matches the interfaceIds existing in the `supportsInterface(..)` of the ERC165Extension.

If the function doesn't exist on the extension or the function reverts, the call will return false.

### DIFF
--- a/contracts/LSP17ContractExtension/LSP17Extendable.sol
+++ b/contracts/LSP17ContractExtension/LSP17Extendable.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.4;
 
 // modules
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {ERC165Checker} from "../Custom/ERC165Checker.sol";
 
 // constants
 import {_INTERFACEID_LSP17_EXTENDABLE} from "./LSP17Constants.sol";
@@ -19,6 +20,26 @@ abstract contract LSP17Extendable is ERC165 {
      */
     function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
         return interfaceId == _INTERFACEID_LSP17_EXTENDABLE || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev Returns whether the interfaceId being checked is supported in the extension of the
+     * {supportsInterface} selector.
+     */
+    function _supportsInterfaceInERC165Extension(bytes4 interfaceId)
+        internal
+        view
+        virtual
+        returns (bool)
+    {
+        address erc165Extension = _getExtension(ERC165.supportsInterface.selector);
+        if (erc165Extension == address(0)) return false;
+
+        if (ERC165Checker.supportsERC165Interface(erc165Extension, interfaceId)) {
+            return true;
+        } else {
+            return false;
+        }
     }
 
     /**
--- a/contracts/LSP17ContractExtension/LSP17Extendable.sol
+++ b/contracts/LSP17ContractExtension/LSP17Extendable.sol
@@ -35,11 +35,7 @@ abstract contract LSP17Extendable is ERC165 {
         address erc165Extension = _getExtension(ERC165.supportsInterface.selector);
         if (erc165Extension == address(0)) return false;
 
-        if (ERC165Checker.supportsERC165Interface(erc165Extension, interfaceId)) {
-            return true;
-        } else {
-            return false;
-        }
+        return ERC165Checker.supportsERC165Interface(erc165Extension, interfaceId);
     }
 
     /**
--- a/contracts/LSP9Vault/LSP9VaultCore.sol
+++ b/contracts/LSP9Vault/LSP9VaultCore.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.4;
 
 // interfaces
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import {ILSP1UniversalReceiver} from "../LSP1UniversalReceiver/ILSP1UniversalReceiver.sol";
 
 // libraries
@@ -139,6 +140,7 @@ contract LSP9VaultCore is
             interfaceId == _INTERFACEID_LSP9 ||
             interfaceId == _INTERFACEID_LSP1 ||
             interfaceId == _INTERFACEID_LSP14 ||
+            _supportsInterfaceInERC165Extension(interfaceId) ||
             super.supportsInterface(interfaceId);
     }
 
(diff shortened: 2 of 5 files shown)


skimaharvey (commented):
Pretty cool 🚀

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -139,6 +139,7 @@ contract LSP9VaultCore is
             interfaceId == _INTERFACEID_LSP9 ||
             interfaceId == _INTERFACEID_LSP1 ||
             interfaceId == _INTERFACEID_LSP14 ||
+            _supportsInterfaceInERC165Extension(interfaceId) ||
```
skimaharvey, line 142: I dont think `super.supportsInterface(interfaceId);` is reading from storage so probably better to put it after so we don't read from storage for nothing? (same goes for LSP0)

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -21,6 +22,26 @@ abstract contract LSP17Extendable is ERC165 {
...
+    function _supportsInterfaceInERC165Extension(bytes4 interfaceId)
+        internal
+        view
+        virtual
+        returns (bool)
+    {
+        address erc165Extension = _getExtension(ERC165.supportsInterface.selector);
+        if (erc165Extension == address(0)) return false;
+
+        if (ERC165Checker.supportsERC165Interface(erc165Extension, interfaceId)) {
```
CJ42, line 38: no need for a `if else` statement here, the function already returns a boolean.

https://github.com/lukso-network/lsp-smart-contracts/blob/61843b7047ddbe140007cf25213aa0e252b2f6b7/contracts/Custom/ERC165Checker.sol#L117-L121

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -21,6 +22,26 @@ abstract contract LSP17Extendable is ERC165 {
...
+     * @dev Returns whether the interfaceId being checked is supported in the extension of the
+     * {supportsInterface} selector.
+     */
+    function _supportsInterfaceInERC165Extension(bytes4 interfaceId)
+        internal
+        view
+        virtual
+        returns (bool)
+    {
+        address erc165Extension = _getExtension(ERC165.supportsInterface.selector);
```
CJ42, line 35: I am not sure I understand this line here.
Is the parameter passed to `_getExtension` not supposed to be `interfaceId`?

CJ42:
```suggestion
        address erc165Extension = _getExtension(interfaceId);
```

tests/LSP17ContractExtension/LSP17Extendable.behaviour.ts
```diff
@@ -657,6 +669,40 @@ export const shouldBehaveLikeLSP17 = (
...
+              let result = await context.contract.supportsInterface(
+                "0xaabbccdd"
+              );
+              expect(result).to.be.false;
+            });
+          });
+
+          describe("when the ERC165 extension was set", () => {
+            let erc165Extension: ERC165Extension;
+            beforeEach(async () => {
```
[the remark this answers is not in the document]
CJ42: yep agree

tests/LSP17ContractExtension/LSP17Extendable.behaviour.ts
```diff
@@ -65,7 +68,8 @@ export const shouldBehaveLikeLSP17 = (
     revertStringFunctionExtensionHandlerKey,
     revertCustomFunctionExtensionHandlerKey,
     emitEventFunctionExtensionHandlerKey,
-    onERC721ReceivedFunctionExtensionHandlerKey;
+    onERC721ReceivedFunctionExtensionHandlerKey,
+    supportsInterfaceFunctionExtensionHandlerKey;
 
   beforeEach(async () => {
```
CJ42, line 74: Here `before` should enough too.

```suggestion
  before(async () => {
```

CJ42 (commented):
added review comments

contracts/LSP9Vault/LSP9VaultCore.sol
skimaharvey: Hmmm yes but let's say I want to read the interfaceID related to the `super.supportsInterface(interfaceId);`, I will waste reading 2100 gas from storage first in `_supportsInterfaceInERC165Extension(interfaceId)`.
Wdyt @CJ42 ?

contracts/LSP17ContractExtension/LSP17Extendable.sol
YamenMerhi: Extensions operate on function selectors

contracts/LSP9Vault/LSP9VaultCore.sol
YamenMerhi: @skimaharvey The `super` keyword is important for contracts that inherits our contract, so one solution could be supporting all interfaceId by hardcoding (ERC725X, Y, LSP17) and then `_supportsInterfaceInERC165Extension` and then `super`

CJ42 (approved):

contracts/LSP9Vault/LSP9VaultCore.sol
YamenMerhi: Applied @skimaharvey suggested changes ✅

skimaharvey (approved):

PR #386, opened 2022-12-01
test: improve speed for running tests from LSP3 + LSP6 test suites.

CJ42:
# What does this PR introduce?

This PR concerns the following test suites:

- LSP3
- LSP6 (**mainly**)

These test suites take a lot of time to run in the CI, in particular LSP6. See screenshot below.

image.

Improve the speed of these tests, using the following 2 techniques:
- replace `beforeEach` hooks by `before` hooks: avoid heavy setup on every single tests (deploying a UP, deploying a KeyManager, setting the permissions + link to a default LSP1UniversalReceiverDelegate, transferring ownership to the Key Manager, etc...)
- avoid sending separate transactions in the setup of the `before` hooks to fund the UP with LYX (for tests that require native LYX tokens transfers. Fund the UP immediately on deployment).

PR #394, opened 2022-12-09
refactor: reentrancy test for better maintainability

b00ste:

- Split the REENTRANCY permission tests into 6 tests files.
- Create a helper file.
- Automatize the tests related to `NotAuthorised(..)` .


tests/LSP6KeyManager/tests/Reentrancy/reentrancyHelpers.ts
```diff
@@ -0,0 +1,650 @@
...
+    permissionsText: "REENTRANCY, CHANGEUNIVERSALRECEIVERDELEGATE",
+    permissions: combinePermissions(
+      PERMISSIONS.REENTRANCY,
+      PERMISSIONS.CHANGEUNIVERSALRECEIVERDELEGATE
+    ),
+    missingPermission: "",
+  },
+};
+
+export const buildReentrancyContext = async (context: LSP6TestContext) => {
```
CJ42, line 322: I am not sure about having this function in the helpers.
I think it will make it difficult to keep track of the setup for each test suite here, because we have to jump to a different file to look at it, and as a result we loose track of it.

I would suggest to take this code and put it in every single `before` block of each Reentrancy test suite.

I would suggest to take the content of each of these

tests/LSP6KeyManager/tests/Reentrancy/reentrancyHelpers.ts
```diff
@@ -0,0 +1,650 @@
...
+      [
+        reentrantContract.address,
+        singleReentarncyRelayer.address,
+        batchReentarncyRelayer.address,
+      ],
+      ["0xffffffff", "0xffffffff", "0xffffffff"]
+    ),
+  ];
+
+  await context.universalProfile
```
CJ42, line 381: We can remove this and reuse the `setupKeyManager` function in the fixture.

https://github.com/lukso-network/lsp-smart-contracts/blob/a4f7cf34f5607b87cf138580decfed2d70b8c994/tests/utils/fixtures.ts#L47-L74

tests/LSP6KeyManager/tests/Reentrancy.test.ts
```diff
@@ -1,9609 +1,52 @@
...
-          await context.universalProfile.provider.getBalance(
-            contract_with_REENTRANCY_VALUETRANSFER_permissions_with_calls.address
-          )
-        ).to.equal(ethers.utils.parseEther("1"));
-      });
-    });
-
-    describe("when reentering and setting data", () => {
-      let contract_without_permissions: ReentrancyWithSetData;
-      let contract_with_ALL_PERMISSI
```
CJ42, line 49: I put a suggested comment below. I think we can remove the `buildReentrancyContext` and put the content inside the `before` block for each Reentrancy test suite in each file.
Otherwise it will make it difficult to debug, as you have to jump to another file to debug the setup and it makes you loose or hard to keep track and remember the context.

tests/LSP6KeyManager/tests/Reentrancy/SingleExecuteRelayCallToSingleExecuteRelayCall.test.ts
```diff
@@ -0,0 +1,664 @@
...
+        "TRANSFERVALUE",
+        reentrancyContext.singleReentarncyRelayer,
+        reentrancyContext.reentrantSigner,
+        reentrancyContext.newControllerAddress,
+        reentrancyContext.newURDAddress
+      );
+    });
+
+    transferValueTestCases.NotAuthorised.forEach((testCase) => {
+      it(`should revert if the reentrant contract has the following permissions: ${testCase.permissionsText}`, async () => {
```
skimaharvey, line 165: `should revert if the reentrant contract has the following permissions` why? I think you should add the "missingPermissions" value as well because this is what matter the most?

tests/LSP6KeyManager/tests/Reentrancy/SingleExecuteRelayCallToSingleExecuteRelayCall.test.ts
```diff
@@ -0,0 +1,664 @@
...
+const loadTestCase = async (
+  payloadType: string,
+  testCase: TransferValueTestCase | SetDataTestCase | SimplePermissionTestCase,
+  context: LSP6TestContext,
+  reentrancyContext: ReentrancyContext
+) => {
+  let permissionKeys: BytesLike[];
+  let permissionValues: BytesLike[];
+
+  switch (payloadType) {
```
skimaharvey, line 48: I think this switch statement could be added to an helper file because it takes 100 lines and we find on the 4 files

CJ42 (changes requested):
added 1st round of review

skimaharvey (approved):
Tests are solids and clean however I think in the future for tests. Maybe we should prefer Readability over DRY.
Sometimes it is a little challenging to follow. Not sure what @YamenMerhi and @CJ42 think?

tests/LSP6KeyManager/tests/Reentrancy/SingleExecuteRelayCallToSingleExecuteRelayCall.test.ts
CJ42: I agree as well. The tests for revert should explain why it reverts.

a) either because the caller has **only a subset of the required permissions**.

_e.g:_ when caller has (only) permission `SETDATA` but tries to set an array of data keys that includes a data key to add a new controller, it should revert.

or 

b) because the caller is **missing a required permission**

_e.g: when caller does not have permission `_ADDPERMISSIONS_ and tries to set an array of data keys that includes a data key to add a new controller, it should revert.

b00ste: do you think mentioning the given & missing permissions in the describe block would make it easier to understand?

tests/LSP6KeyManager/tests/Reentrancy/SingleExecuteRelayCallToSingleExecuteRelayCall.test.ts
b00ste: good point

tests/LSP6KeyManager/tests/Reentrancy/reentrancyHelpers.ts
```diff
@@ -0,0 +1,650 @@
...
+    newURDAddress,
+    reentrantContract,
+    reentrantSigner,
+    singleReentarncyRelayer,
+    batchReentarncyRelayer,
+    randomLSP1TypeId,
+  };
+};
+
+export const generateRelayCall = async (
```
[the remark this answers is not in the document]
b00ste: I can see it for `generateRelayCall(..)` as it is very generic but `generateSingleRelayPayload(..)` and `generateBatchRelayPayload(..)` seem too specific and tied to specific contracts, `SingleReentrancyRelayer` and `BatchReentrancyRelayer`, not sure in what other context would you use it.
I mean I have no problems with moving all of them to `fixtures.ts`, just not sure how will we use it in other tests.

tests/LSP6KeyManager/tests/Reentrancy/reentrancyHelpers.ts
```diff
@@ -0,0 +1,650 @@
...
+    .transferOwnership(context.keyManager.address);
+
+  const acceptOwnershipPayload =
+    context.universalProfile.interface.encodeFunctionData("acceptOwnership");
+  await context.keyManager
+    .connect(owner)
+    ["execute(bytes)"](acceptOwnershipPayload);
+
+  // Fund Universal Profile with some LYXe
+  await owner.sendTransaction({
```
CJ42, line 386: maybe we can fund the UP immediately on deployment, instead of sending a separate transaction in the `before` block.

This will reduce a bit the CI time and clean up the code.

https://github.com/lukso-network/lsp-smart-contracts/pull/386/commits/167b7c1222a781911b69dac05ac0234e617a7ca3


image

CJ42 (approved):

PR #399, opened 2022-12-16
refactor: add length check generateArrayKey(...)

skimaharvey:

`generateArrayKey(...)` would underflow if the length of the string passed was smaller than 2 characters.


b00ste (approved):

CJ42 (approved):

YamenMerhi (changes requested):
Maybe we can do `require(dataKey.length >= 2)` to emphasize that we require brackets `[]`. Just feels to me more readable

YamenMerhi (approved):

PR #401, opened 2022-12-16
refactor: optimise gas usage of `setData(...)` via LSP6KeyManager, improve code readability and storage layout.

CJ42:

This PR is a large refactor of the `LSP6KeyManagerCore` contract. It aims to:
- simplify the Solidity code of this contract for better readability.
- optimise the gas usage for common UP interactions via `setData(...)`.
- optimise the deployment cost of the `LSP6KeyManager` contract.

## ⚠️ BREAKING CHANGES

This PR introduces 4 main breaking changes on the Key Manager. Both related to internals and the behaviours.

### New Behaviours

1) With this PR, it is now possible to set the `bytes32(0)` data key (zero data key).

2) Also, the Key Manager does not reverts anymore when passed an invalid operation type for `ERC725X.execute(...)`. The reverts will now occur at ERC725X level. This is to avoid duplicate check following suggestions by @YamenMerhi in https://github.com/ERC725Alliance/ERC725/pull/184#discussion_r1044241245.

### New Main internal changes

3) the `target` variable is now changed from `public` to `internal _target`.`target()` interface function is now implemented as its own getter function. This is to reduce the deployment cost (-8,841 gas).

4) The `_reentrancyStatus` was changed from a `uint256` to a `bool` to be tightly pack into the same storage slot next to `_target`.

## Gas Costs Optimisations

The table below summarises the gas cost difference for common UP interactions (from the Gas Benchmark Bot).

image

### ♻️ code refactoring

- Improved the code for the paths `setData(bytes32,bytes)` to `setData(bytes32[],bytes[])`.

Removed the duplicated `if { .... } else if { ... } else { ... }` blocks on each data key checks, and moved everything in an `internal` function that retrieves the required `bytes32` permission based on the ERC725Y data key being set.

- used simple assembly instead of the `BytesLib.slice` library function that is expensive.

- optimised the code of `_verifyAllowedERC725YDataKeys` and `_verifyAllowedERC725YKeySingle`.

- reduced deployment and runtime cost by inlining with ternary operators instead of using nested `if { ... } else { ... }`.

### 🔴  revert errors

- revert reason string for Delegate Call disallowed was changed with a custom error.
- the revert reason string for the default error handler was changed from `"LSP6: unknown Error occured on the linked target contract" to a shorter string (less than 32 characters) --> `"LSP6: failed executing payload"`

### 📄 Documentation

Improved most of the Natspec comments and added comments in most part of the code to explain the internal logic.

## Storage Layout changes

LSP6 - KeyManager storage layout


b00ste (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -857,8 +881,8 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         bool isAllowedAddress;
         bool isAllowedFunction;
 
-        for (uint256 ii = 0; ii < allowedCallsLength; ii += 29) {
-
+        uint256 ii = 0;
```
skimaharvey, line 884: I believe it is cheaper if we don't initialize it. 
Also might want to add it back to the for loop for readability

skimaharvey (approved):
Looks very good to me. Much clearer 🚀 
Fuzzed `verifyAllowedERC725YDataKeys(...)` and `verifyAllowedERC725YSingleKey(...)` and did not have any issues

YamenMerhi (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42: Changed, deployment cost increase by +420 gas, but runtime cost does not change.
I guess the trade off for readability here is better over the small deployment cost improvement.

frozeman (commented):
Nice work!

PR #404, opened 2022-12-17
Lsp2 utils/update generate array element key at index

skimaharvey:
## What this PR introduce? 

- [x] Change index param to `uint128` for the  `generateArrayElementKeyAtIndex(...)` function
- [x] Add checks to prevent cast errors when casting bytes down


contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -70,13 +82,19 @@ library LSP10Utils {
             // If the storage is already initiated
         } else if (encodedArrayLength.length == 32) {
             uint256 oldArrayLength = uint256(bytes32(encodedArrayLength));
+            uint256 newArrayLength = oldArrayLength + 1;
+
+            if (newArrayLength > type(uint64).max) {
```
skimaharvey, line 87: The reason here I am comparing to `uint64` is because on line 102 we cast down to bytes8 and to avoid overriding keys

b00ste (approved):

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -70,13 +82,19 @@ library LSP10Utils {
...
+            uint128 oldArrayLength128 = uint128(oldArrayLength);
 
             keys[0] = _LSP10_VAULTS_ARRAY_KEY;
-            values[0] = bytes.concat(bytes32(oldArrayLength + 1));
+            values[0] = bytes.concat(bytes32(newArrayLength));
 
             keys[1] = LSP2Utils.generateArrayElementKeyAtIndex(
                 _LSP10_VAULTS_ARRAY_KEY,
-                oldArrayLength
+                oldArrayLength128
```
CJ42, line 97: since we are using the variable only once, we can inline to avoid keeping track of it on the stack.

```suggestion
                uint128(oldArrayLength128)
```

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -70,13 +82,19 @@ library LSP10Utils {
             // If the storage is already initiated
         } else if (encodedArrayLength.length == 32) {
             uint256 oldArrayLength = uint256(bytes32(encodedArrayLength));
+            uint256 newArrayLength = oldArrayLength + 1;
+
+            if (newArrayLength > type(uint64).max) {
+                revert VaultIndexSuperiorToUint64(newArrayLength);
+            }
+            uint128 oldArrayLength128 = uint128(oldArrayLength);
```
CJ42, line 90: this variable can be removed and inlined. See comment below.

CJ42 (approved):

YamenMerhi (approved):

CJ42 (changes requested):
@skimaharvey you need to add the latest changes from `develop` so that the last CI job runs.

CJ42 (approved):

PR #433, opened 2023-01-11
refactor!:  [WP-I16] Revert instead of return when no extension exist in LSP17

YamenMerhi:
### Breaking Changes 🚫 
- Change the LSP17 logic to revert when there is no extension for the function selector being called 
- If the first 4 bytes are plain 4 zero bytes, the call is allowed and it just return, this check allow sending graffiti to not be checked against an extension.

###  Periphery
- Move the check of `msg.data < 4 || msg.sig == bytes4(0)` to the LSP0 and LSP9 instead of having them as part of the Core LSP17 implementation. 

### New Behavior  
- LSP0 & LSP9 now revert if the function selector being called with doesn't have an extension
- If the call to LSP0 & LSP9 is prepended with 4 bytes(0), the call is allowed and we just return.


skimaharvey (approved):

b00ste (approved):

constants.js
```diff
@@ -455,6 +455,13 @@ const Errors = {
 			message: 'LSP14: Cannot transfer ownership to address(this)',
 		},
 	},
+	LSP17ContractExtension: {
+		'0x3d746f74': {
+			error: 'NoExtensionForFunctionSelector(bytes4)',
```
CJ42, line 460: maybe we could rename it to?

```suggestion
			error: 'NoExtensionFoundForFunctionSelector(bytes4)',
```

contracts/LSP17ContractExtension/LSP17Errors.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+/**
+ * @dev reverts when there is no extension for the function selector being called with
+ */
+error NoExtensionForFunctionSelector(bytes4 functionSelector);
```
CJ42, line 7: As suggested above in the `constants.js`

```suggestion
error NoExtensionFoundForFunctionSelector(bytes4 functionSelector);
```

contracts/LSP17ContractExtension/LSP17Extendable.sol
```diff
@@ -51,26 +54,23 @@ abstract contract LSP17Extendable is ERC165 {
     function _getExtension(bytes4 functionSelector) internal view virtual returns (address);
 
     /**
-     * Forwards the call to an extension mapped to a function selector. If the address of
-     * the extension is address(0), then return.
+     * @dev Forwards the call to an extension mapped to a function selector. If the address of
```
CJ42, line 57: Suggested comment change

```suggestion
     * @dev Forwards the call to an extension mapped to a function selector. If no extension address 
     * is mapped to the function selector (address(0)), then revert.
```

tests/LSP16UniversalFactory/LSP16UniversalFactory.test.ts
```diff
@@ -512,7 +512,7 @@ describe("UniversalFactory contract", () => {
         await context.universalFactory.deployCreate2Init(
           UPBytecode,
           salt,
-          "0xaabbccdd",
+          "0x00000000aabbccdd",
```
CJ42, line 515: We could add an extra comment here to describe that it is just some random data sent along.

```suggestion
            "0x00000000aabbccdd", // send some random data along prepended with `
0x00000000`
```

tests/LSP17ContractExtension/LSP17Extendable.behaviour.ts
```diff
@@ -723,9 +709,82 @@ export const shouldBehaveLikeLSP17 = (
...
+            context.accounts[0].sendTransaction({
+              to: context.contract.address,
+              data: "0x01",
+            })
+          ).to.not.be.reverted;
+        });
+      });
+
+      describe("when calling with a payload preprended with 4 bytes of 0", () => {
+        describe("when the payload is equal to 4 bytes of zeros", () => {
```
CJ42, line 722:
```suggestion
        describe("when the payload is `0x00000000`", () => {
```

tests/LSP17ContractExtension/LSP17Extendable.behaviour.ts
```diff
@@ -723,9 +709,82 @@ export const shouldBehaveLikeLSP17 = (
...
+              await expect(
+                context.accounts[0].sendTransaction({
+                  to: context.contract.address,
+                  data: "0x00000000",
+                })
+              ).to.not.be.reverted;
+            });
+          });
+        });
+        describe("when the payload is larger than 4 bytes of zeros", () => {
```
CJ42, line 748: Suggested improvement in the description of the test

```suggestion
        describe("when the payload is `0x00000000` + some random data ('graffiti')", () => {
```

CJ42 (changes requested):
The changes looks good. Added some review comments

CJ42 (approved):

PR #435, opened 2023-01-11
[WP-M2] LSP-2: `LSP2Utils.isCompactBytesArray(bytes)` does not support zero-length elements (`[..., 0x, ...]`)

CJ42:

Based on the findings below, the changes detailed below are introduced.

image

image

### Modify logic of `isCompactBytesArray` in LSP2Utils

In `LSP2Utils`, modify the function `isCompactBytesArray(...)` to allow empty length elements like `0x00`.

For instance, a compact bytes array that contain entries that are zero bytes should be allowed, and pass the `isCompactBytesArray` check.

### Exceptions in Key Manager

Make two exceptions in the Key Manager to not allow empty `0x00` elements in the compact bytes array for `AllowedCalls` or `AllowedERC725YDataKeys`.

As a result, 2 x custom functions are introduced:

- `isCompactBytesArrayOfAllowedCalls`: only allow elements of length 28.
- `isCompactBytesArrayOfAllowedERC725YDataKeys`: only allow elements of length 0 or less than 32 bytes.


CJ42:
Closed. Replaced by #438

PR #438, opened 2023-01-13
fix: [WP-M2] LSP-2: `LSP2Utils.isCompactBytesArray(bytes)` does not support zero-length elements (`[..., 0x, ...]`)

CJ42:

Based on the findings below, the changes detailed below are introduced.

image

image

### Modify logic of `isCompactBytesArray` in LSP2Utils

In `LSP2Utils`, modify the function `isCompactBytesArray(...)` to allow empty length elements like `0x00`.

For instance, a compact bytes array that contain entries that are zero bytes should be allowed, and pass the `isCompactBytesArray` check.

### Exceptions in Key Manager

Make two exceptions in the Key Manager to not allow empty `0x00` elements in the compact bytes array for `AllowedCalls` or `AllowedERC725YDataKeys`.

As a result, 2 x custom functions are introduced:

- `isCompactBytesArrayOfAllowedCalls`: only allow elements of length 28.
- `isCompactBytesArrayOfAllowedERC725YDataKeys`: only allow elements of length 0 or less than 32 bytes.


contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -81,6 +81,66 @@ library LSP6Utils {
...
+     */
+    function isCompactBytesArrayOfAllowedCalls(bytes memory allowedCallsCompacted)
+        internal
+        pure
+        returns (bool)
+    {
+        uint256 pointer;
+
+        while (pointer < allowedCallsCompacted.length) {
+            if (pointer + 1 >= allowedCallsCompacted.length) return false;
```
skimaharvey, line 98: I think here should be `>`

contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -81,6 +81,66 @@ library LSP6Utils {
...
+     * @param allowedERC725YDataKeysCompacted a compact bytes array of ERC725Y Data Keys (full bytes32 data keys or bytesN prefix) to check.
+     * @return true if the value passed is a valid compact bytes array of ERC725Y Data Keys, false otherwise.
+     */
+    function isCompactBytesArrayOfAllowedERC725YDataKeys(
+        bytes memory allowedERC725YDataKeysCompacted
+    ) internal pure returns (bool) {
+        uint256 pointer;
+
+        while (pointer < allowedERC725YDataKeysCompacted.length) {
+            if (pointer + 1 >= allowedERC725YDataKeysCompacted.length) return false;
```
skimaharvey, line 127: same comment as above

tests/LSP2ERC725YJSONSchema/LSP2UtilsLibrary.test.ts
```diff
@@ -192,14 +192,73 @@ describe("LSP2Utils", () => {
...
+          "0x112233",
+        ]);
+
+        // replace the first length byte of 0xaabbccddee with an invalid length value
+        data = String(data).replace(/05/g, "10");
+        const result = await lsp2Utils.isCompactBytesArray(data);
+        expect(result).to.be.false;
+      });
+
+      it("should return false if the byte length of the last element is invalid and points 'too far'", async () => {
```
skimaharvey, line 260: nice new tests for isCompactBytesArray

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -884,7 +904,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         bytes memory allowedCalls = ERC725Y(_target).getAllowedCallsFor(from);
         uint256 allowedCallsLength = allowedCalls.length;
 
-        if (allowedCallsLength == 0 || !LSP2Utils.isCompactBytesArray(allowedCalls)) {
+        if (allowedCallsLength == 0 || !LSP6Utils.isCompactBytesArrayOfAllowedCalls(allowedCalls)) {
```
skimaharvey, line 907: hmm really not keen on this second check. In most cases, CBA will be valid so it will penalized most people where we will loop over the CBA twice... Invalid CBA wont make it easier for someone that does not have the necessary allowed calls anw so I think we should remove `LSP6Utils.isCompactBytesArrayOfAllowedCalls(allowedCalls)`

skimaharvey (commented):
Really nice new tests on the CBA. Also like how you separated the logic between ERC725YKeys and AllowedCalls

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: Also very unliked to have an invalid CBA because we check it when setting it. Really should remove this check

YamenMerhi (commented):
As @skimaharvey mentioned, let's do the check only when the data is getting set for now. Having it run on set and on execution is overkill

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42: yes discuss with @frozeman @YamenMerhi and @b00ste and come to a decision.
I agree on removing it as it is a waste of gas for a very rare edge case

skimaharvey (approved):

b00ste (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
b00ste: Yep, let's remove it for now. If we find a better way of doing it we could add it later

YamenMerhi (approved):

PR #441, opened 2023-01-16
refactor: make some internal functions virtual

b00ste:


skimaharvey (approved):

skimaharvey:
> Looks good to me, but not sure if `_generateSalt` should be mark as `virtual` as seems to be quite an important function for the logic of the UniversalFactory.
> 
> Let's see with @YamenMerhi

I think it is fine to override this function as well

b00ste:
#443

PR #447, opened 2023-01-18
refactor: refactor LSP0 `isValidSignature(..)` to low level call 

YamenMerhi:
- Remove the need to check for supportsInterface in `isValidSignature(..)` .
- Handle the call of `isValidSignature(..)` in a way that the check will not revert in case the function do not exist or the function return an unintended value. Instead just return failure value
- Unify the naming, use `isValidSignature.selector` instead of `InterfaceId`. 
- Use `code.length > 0` instead of `isContract(..)`
- Adjust the LSP0 tests to do the intended behavior of checking the signature (remove todo)
- Add tests for malicious behavior

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -162,11 +162,11 @@ abstract contract LSP0ERC725AccountCore is
         virtual
         returns (bytes4 magicValue)
     {
-        address owner_ = owner();
+        address _owner = owner();
 
         // If owner is a contract
-        if (owner_.code.length > 0) {
-            (bool success, bytes memory result) = owner_.staticcall(
+        if (_owner.code.length > 0) {
+            (bool success, bytes memory result) = _owner.staticcall(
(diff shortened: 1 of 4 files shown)


b00ste (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -161,18 +162,25 @@ abstract contract LSP0ERC725AccountCore is
...
-        address _owner = owner();
-        // if OWNER is a contract
-        if (_owner.isContract()) {
-            return
-                _owner.supportsERC165InterfaceUnchecked(_INTERFACEID_ERC1271)
-                    ? IERC1271(_owner).isValidSignature(dataHash, signature)
-                    : _ERC1271_FAILVALUE;
-            // if OWNER is a key
-        } else {
+        address owner_ = owner();
```
skimaharvey, line 165: I think appending with `_` is usually associated to params?

skimaharvey (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
CJ42: it's usually for `constructor` params

PR #449, opened 2023-01-18
refactor: add check for the payload length in `LSP6KeyManagerCore`

b00ste:

In this PR the following changes are made:
- A new error added, 'InvalidPayload(bytes)'.
- A check added in both internal function, `` and `` in order to verify that the payload is not smaller than 4 bytes.

### DIFF
--- a/tests/LSP6KeyManager/tests/PermissionCall.test.ts
+++ b/tests/LSP6KeyManager/tests/PermissionCall.test.ts
@@ -76,26 +76,6 @@ export const shouldBehaveLikePermissionCall = (
   });
 
   describe("when interacting via `execute(...)`", () => {
-    describe("when the payload is smaller than 4 bytes", () => {
-      it("should revert when using `execute(..)` with a payload smaller than 4 bytes", async () => {
-        await expect(context.keyManager["execute(bytes)"]("0xaabbcc"))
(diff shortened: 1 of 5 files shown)


tests/LSP6KeyManager/tests/PermissionCall.test.ts
```diff
@@ -76,6 +76,26 @@ export const shouldBehaveLikePermissionCall = (
   });
 
   describe("when interacting via `execute(...)`", () => {
+    describe("when the payload is smaller than 4 bytes", () => {
```
CJ42, line 79: Could we put this test in the file `OtherScenarios.test.ts` for the Key Manager test suite?

I had already started to draft one test long time ago and left it there skipped. 

Maybe we can also re-arrange this test to ensure this new custom error is returned with an empty payload? :)

https://github.com/lukso-network/lsp-smart-contracts/blob/5f4b9a7b2e224f1536be8d5164f58b57016cafcd/tests/LSP6KeyManager/tests/OtherScenarios.test.ts#L55-L57

CJ42 (changes requested):

skimaharvey (approved):

YamenMerhi (approved):

PR #452, opened 2023-01-19
refactor: remove isCBACheck

skimaharvey:
# What does this PR introduce

In `verifyAllowedCalls` we are looping over the allowedCalls array twice, once to make sure it is a valid allowedCalls and once to verify if the caller has the allowedCall permission.
Since we are checking the validity of the compact bytes array when setting the allowedCalls, in the vast majority of the case this will just be a waste of gas that can get expensive when the allowedCalls compact byte array get big. 

- [x] Remove `isCompactBytesArray` check
- [x] Update tests
- [x] Add new check within the loop to avoid `out_of_bonds` error

### DIFF
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -932,6 +932,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         bool isAllowedFunction;
 
         for (uint256 ii; ii < allowedCallsLength; ii += 30) {
+            if (ii + 28 > allowedCallsLength) {
+                revert InvalidEncodedAllowedCalls(allowedCalls);
+            }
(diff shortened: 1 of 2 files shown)


contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -932,6 +932,9 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         bool isAllowedFunction;
 
         for (uint256 ii; ii < allowedCallsLength; ii += 30) {
+            if (ii + 28 > allowedCallsLength) {
```
CJ42, line 935:
```suggestion
            if (ii + 28 >= allowedCallsLength) {
```

CJ42 (changes requested):

b00ste (approved):

CJ42 (approved):

PR #453, opened 2023-01-20
test: add fuzzing test to LSP16

skimaharvey:

Fuzzing tests for LSP16: 

- [x] LSP16 should not keep ethers 
- [x] `calculateAddress` should calculate the right address 
- [x] `calculateProxyAddress` should calculate the right address  
- [x] `_generateSalt` should be unique  

While checking for the above properties, we are also checking the LSP16 works properly and that contracts are being deployed

### DIFF
--- a/contracts/Mocks/NonPayableFallback.sol
+++ b/contracts/Mocks/NonPayableFallback.sol
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+contract NonPayableFallback {
+    fallback() external {}
+}
--- a/contracts/Mocks/NonPayableFallback.sol
+++ b/contracts/Mocks/NonPayableFallback.sol
@@ -2,5 +2,6 @@
 pragma solidity ^0.8.0;
 
 contract NonPayableFallback {
+    // solhint-disable-next-line payable-fallback
     fallback() external {}
 }
--- a/tests/foundry/LSP16UniversalFactory/LSP16UniversalProfile.t.sol
+++ b/tests/foundry/LSP16UniversalFactory/LSP16UniversalProfile.t.sol
@@ -129,7 +129,10 @@ contract LSP16UniversalProfileTest is Test {
         require(address(lsp16).balance == 0, "LSP16 should not have any balance");
     }
 
-    function testDeployCreate2InitShouldNotKeepValueWithUPInit(uint128 valueForInitializer) public {
+    function testDeployCreate2InitShouldNotKeepValueWithUPInit(
+        uint128 valueForInitializer,
+        bytes4 initilializerBytes
+    ) public {
         vm.deal(address(this), valueForInitializer);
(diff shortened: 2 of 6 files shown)


contracts/Mocks/NonPayableFallback.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+contract NonPayableFallback {
+    fallback() external {}
```
YamenMerhi, line 5: Solhint warning

tests/foundry/LSP16UniversalFactory/LSP16UniversalProfile.t.sol
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    function testCalculateAddressShouldReturnCorrectUPAddressWithDeployCreate2Init(
+        bytes32 providedSalt,
+        uint256 valueForInitializer
+    ) public {
+        vm.deal(address(this), valueForInitializer);
+        assert(address(this).balance == valueForInitializer);
+
+        bytes memory initializeCallData = hex"aabbcc";
```
YamenMerhi, line 254: Can't we pass this as a param and put `if ( bytes4(initializeCallData) != bytes4(0) || initializeCallData.length > 4 ) return` ? something like this. If not, its already good, its just a slight improvement

YamenMerhi (approved):

tests/foundry/LSP16UniversalFactory/LSP16UniversalProfile.t.sol
skimaharvey: The thing is that I am afraid we will "loose" a lot of tests if the seeds decides to send lots of initializeCallData of lengthes > 4. 
What I probably do is something like `abi.encodePacked(bytes4(initializeCallData))`

tests/foundry/LSP16UniversalFactory/LSP16UniversalProfile.t.sol
```diff
@@ -0,0 +1,378 @@
...
+
+    uint256 public testCounter;
+
+    function setUp() public {
+        lsp16 = new LSP16Mock();
+
+        nonPayableFallbackContract = new NonPayableFallback();
+        fallbackInitializer = new FallbackInitializer();
+        lsp0Init = new LSP0ERC725AccountInit();
+        lsp0 = new LSP0ERC725Account(address(20));
```
CJ42, line 38: what does the `address(20)` do here? Curious as never saw this before

CJ42 (approved):

tests/foundry/LSP16UniversalFactory/LSP16UniversalProfile.t.sol
skimaharvey: I dont know why I picked 20. Just used a random address as owner of LSP0ERC725Account 😅

PR #455, opened 2023-01-20
fix: [WP-I6] add `onERC721Received` check for the `safeTransferFrom` functions of `LSP8CompatibleERC721` contracts

CJ42:

## ⚠️ BREAKING CHANGES

In `LSP8CompatibleERC721` and `LSP8CompatibleERC721Init`:

- [x] make the `allowNonLSP1Recipient` parameter set to `true` by default for all transfer functions (`transfer` from LSP8, `transferFrom` and `safeTransferFrom` from ERC721).

- [x] for the `safeTransferFrom` function, add an extra call to check that the token receiver MUST implement the `onERC721Received` function and return the magic value (`onERC721Received(address,address,uint256,bytes)` function selector).

## 🧪 Tests

- [x] add tests for the following scenarios below.

- when `to` is an EOA = ✅ pass  
- when `to`  is a contract that supports + implement LSP1 
    - but does not implement `onERC721Received` = ❌ fail and revert  
    - if it also implements `onERC721Received`, but the function `onERC721Received` reverts = ❌ fails and revert  
    - if it also implements `onERC721Received` , but the function `onERC721Received`  does not return the right bytes value = ❌ fails and revert  
    - if it also implements `onERC721Received` = ✅ pass  

- when `to` is a contract that does not supports LSP1
    - but does not implement `onERC721Received` = ❌ fail and revert  
    - if it also implements `onERC721Received` , but the function `onERC721Received` reverts = ❌ fails and revert  
    - if it also implements `onERC721Received` , but the function `onERC721Received` does not return the right bytes value = ❌ fails and revert  
    - if it also implements `onERC721Received` = ✅ pass

### DIFF
--- a/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
@@ -285,6 +285,7 @@ abstract contract LSP8CompatibleERC721 is
                 if (reason.length == 0) {
                     revert("LSP8CompatibleERC721: transfer to non ERC721Receiver implementer");
                 } else {
+                    // solhint-disable no-inline-assembly
                     /// @solidity memory-safe-assembly
                     assembly {
                         revert(add(32, reason), mload(reason))
--- a/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
@@ -3,10 +3,12 @@ pragma solidity ^0.8.12;
 
 // interfaces
 import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import {ILSP8CompatibleERC721} from "./ILSP8CompatibleERC721.sol";
 import {ILSP8IdentifiableDigitalAsset} from "../ILSP8IdentifiableDigitalAsset.sol";
 
 // libraries
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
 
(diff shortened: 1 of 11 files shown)


tests/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.test.ts
```diff
@@ -78,7 +78,7 @@ describe("LSP8CompatibleERC721", () => {
     });
   });
 
-  describe("when using LSP8 contract with proxy", () => {
+  describe.skip("when using LSP8 contract with proxy", () => {
```
YamenMerhi, line 81: `.skip`

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721InitAbstract.sol
```diff
@@ -234,6 +256,42 @@ abstract contract LSP8CompatibleERC721InitAbstract is
...
+            try IERC721Receiver(to).onERC721Received(msg.sender, from, tokenId, data) returns (
+                bytes4 retval
+            ) {
+                return retval == IERC721Receiver.onERC721Received.selector;
+            } catch (bytes memory reason) {
+                if (reason.length == 0) {
+                    revert("LSP8CompatibleERC721: transfer to non ERC721Receiver implementer");
+                } else {
+                    /// @solidity memory-safe-assembly
+                    assembly {
```
YamenMerhi, line 285: linter error for assembly

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
```diff
@@ -244,6 +260,42 @@ abstract contract LSP8CompatibleERC721 is
...
+            try IERC721Receiver(to).onERC721Received(msg.sender, from, tokenId, data) returns (
+                bytes4 retval
+            ) {
+                return retval == IERC721Receiver.onERC721Received.selector;
+            } catch (bytes memory reason) {
+                if (reason.length == 0) {
+                    revert("LSP8CompatibleERC721: transfer to non ERC721Receiver implementer");
+                } else {
+                    /// @solidity memory-safe-assembly
+                    assembly {
```
YamenMerhi, line 289: linter error for assembly

YamenMerhi (changes requested):
First round of review

skimaharvey (approved):

b00ste (approved):

YamenMerhi (approved):
LGTM 🚀 What would be a good additional test is to test whether using `transferFrom` or `safeTransferFrom` when `onERC721Received(..)` pass and LSP1 revert, the whole call will revert even if `allowNonLSP1Recipient` is true.

PR #461, opened 2023-01-27
refactor: separate the LSP6 core contract in logic modules

b00ste:

- [x] Move set data permissions verification logic in `LSP6SetDataLogicModule.sol`.
- [x] Move execute permissions verification logic in `LSP6ExecuteLogicModule.sol`.
- [x] Move ownership transfer permissions verification logic in `LSP6OwnershipLogicModule.sol`.
- [x] Move `getPermissionName(...)` and `requirePermissions(...)` to the `LSP6Utils.sol` for re-usability in modules and core contract.

Side notes:

1. Unfortunately, for avoiding having `requirePermissions(...)` in each module, I moved it to `LSP6Utils.sol`. This function reverts with custom error if the required permission is not present.

> Cons:
> - Library method which reverts.
>
> Pros:
> - Useful function which can be re-used by other developers who want to create a Key Manager contract.

2. The target is passed to the modules as a method parameter.

> Cons:
> - This might confuse some developers regarding the requirement of LSP6 to have a single target contract stored in the storage of the contract.
>
> Pros:
> - This allows for re-usability of the modules in both single-target Key Manager and multi-target Key Managers.


YamenMerhi:
Could you try to merge develop, then clone RV internal repo, and run the foundry tests again the new splitted contracts ?

contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -196,4 +199,48 @@ library LSP6Utils {
...
     }
+
+    /**
+     * @dev returns the name of the permission as a string
+     */
+    function getPermissionName(bytes32 permission)
+        internal
+        pure
+        returns (string memory errorMessage)
+    {
```
CJ42, line 210: this is a good idea to put this function here in the `LSP6Utils`. It is useful, "implementation agnostic" and can be reused 👌

contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -196,4 +199,48 @@ library LSP6Utils {
...
+     * @dev revert if `controller`'s `addressPermissions` doesn't contain `permissionsRequired`
+     * @param controller the caller address
+     * @param addressPermissions the caller's permissions BitArray
+     * @param permissionRequired the required permission
+     */
+    function requirePermissions(
+        address controller,
+        bytes32 addressPermissions,
+        bytes32 permissionRequired
+    ) internal pure {
```
CJ42, line 240: I am not in favour of having this function in `LSP6Utils`. Because this function does not add any additional functionality to the `library`.  It simply re-uses the same `hasPermission` function from the same library + revert.

I think reverting in a library is ok when it is related to some properties that fail but should be true. For instance taking from OpenZeppelin:
- `Clones.sol`: if cloning a contract fails and return the `address(0)`: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/Clones.sol#L35
- `Create2.sol`: if deploying a contract at a predetermined address but not being able to fund it because of insufficient balance: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol#L31
- `Strings.sol`: when converting a number to hex, but the number is `0`: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Strings.sol#L68
- `ECDSA.sol`: when some of the properties of the signature (`v.` `r`, `s`, length, etc...) are invalid: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol#L23-L33

In this case, we force the implementation to revert with the custom error `NotAuthorised`. **But it should be up to the implementation contract really to implement how they want to handle the case if an address does not have some permissions**. For instance, a custom implementation of a Key Manager could:
- revert with a reason string.
- revert with a custom error different than this one (a custom error related to their own protocol).
- not revert but simply return `false` if the controller does not have the permissions, or `true` if it has them.

, but it should be up to the implementation contract (the custom Key Manager being built by the developer) to implement how to handle how they want trevert the way they want (wit how to revert should be up to the implementation

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -313,732 +319,21 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
                 (bytes32[], bytes[])
             );
 
-            _verifyCanSetData(from, permissions, inputKeys, inputValues);
+            _verifyCanSetData(_target, from, permissions, inputKeys, inputValues);
```
YamenMerhi, line 322: Maybe `LSP6SetDataLogicModule. _verifyCanSetData` will make it easier to know where the function exist

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -313,732 +319,21 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
 
             // ERC725X.execute(uint256,address,uint256,bytes)
         } else if (erc725Function == EXECUTE_SELECTOR) {
-            _verifyCanExecute(from, permissions, payload);
+            _verifyCanExecute(_target, from, permissions, payload);
```
YamenMerhi, line 326: Same here `LSP6ExecuteLogicModule. _verifyCanExecute ` will make it easier to know where the function exists

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -313,732 +319,21 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
...
         } else if (
             erc725Function == ILSP14Ownable2Step.transferOwnership.selector ||
             erc725Function == ILSP14Ownable2Step.acceptOwnership.selector
         ) {
-            _requirePermissions(from, permissions, _PERMISSION_CHANGEOWNER);
+            _verifyOwnershipPermissions(from, permissions);
```
YamenMerhi, line 331: here

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -304,7 +310,7 @@ abstract contract LSP6KeyManagerCore is ERC165, ILSP6KeyManager {
         if (erc725Function == SETDATA_SELECTOR) {
             (bytes32 inputKey, bytes memory inputValue) = abi.decode(payload[4:], (bytes32, bytes));
 
-            _verifyCanSetData(from, permissions, inputKey, inputValue);
+            _verifyCanSetData(_target, from, permissions, inputKey, inputValue);
```
YamenMerhi, line 313: and here

contracts/LSP6KeyManager/LSP6LogicModules/LSP6OwnershipLogicModule.sol
```diff
@@ -0,0 +1,21 @@
...
+
+// constants
+import {_PERMISSION_CHANGEOWNER} from "../LSP6Constants.sol";
+
+abstract contract LSP6OwnershipLogicModule {
+    function _verifyOwnershipPermissions(address controllerAddress, bytes32 controllerPermissions)
+        internal
+        pure
+    {
+        LSP6Utils.requirePermissions(
```
CJ42, line 15: I am not totally convinced about the usefulness of this module 🤔  but ok why not

CJ42 (changes requested):
added requested changes

contracts/LSP6KeyManager/LSP6LogicModules/LSP6OwnershipLogicModule.sol
b00ste: I made it so we can have a very basic example for a module. Basically, to show that you don't have to make complex modules like the ones for `execute(..)` or `setData(..)`

b00ste:
Screenshot 2023-02-14 at 13 51 54

YamenMerhi (approved):

skimaharvey (approved):

CJ42 (approved):

PR #465, opened 2023-02-05
refactor: add custom errors to LSP7 and LSP8 to prevent owner as operator

skimaharvey:

Currently we were returning when trying to set the token as operator.  
This could be a problem because the transaction would pass but nothing would happen to the smart contract storage. 
For this reason, this PR introduces the `LSP7TokenOwnerCannotBeOperator` and `LSP8TokenOwnerCannotBeOperator` that will be triggered when trying to authorize or revoke owner as operator. 

- [x] Create and implement LSP7TokenOwnerCannotBeOperator error 
- [x] Create and implement LSP8TokenOwnerCannotBeOperator error 
- [x] replace the `caller` variable from LSP8 by `msg.sender` to avoid pushing it to stack


CJ42 (approved):

CJ42:
let's add the new errors selectors in the `constants.js` file: https://github.com/lukso-network/lsp-smart-contracts/blob/58d40d16e00766570c2c043c5696281900c280b5/constants.js#L402-L435

YamenMerhi (approved):

PR #473, opened 2023-02-19
refactor: Support Extension for `bytes4(0)` sig in LSP0-LSP9

YamenMerhi:

- Changing the logic of handling the `bytes4(0)` sig when called to LSP0 and LSP9.

### Behavior Before
Whenever the contract was called with any data that starts with `bytes4(0)` it would emit `ValueReceived` event if native token was received and then return (ending execution).

### Behavior After
Whenever the contract was called with any data that starts with `bytes4(0)` it would emit `ValueReceived` event if native token was received and check if there is extension set for `bytes4(0)`, if there is an extension set then forward the call to the extension, if not then return (ending execution).

The only difference between `bytes4(0)` and the other selectors is that in case the contract was called with a function selector that doesn't have an extension set it would usually revert, but when the contract is called with `bytes4(0)` the call will still succeed and return even if there is no extension set.

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -41,6 +41,9 @@ import {_INTERFACEID_LSP14} from "../LSP14Ownable2Step/LSP14Constants.sol";
 
 import {_LSP17_EXTENSION_PREFIX} from "../LSP17ContractExtension/LSP17Constants.sol";
 
+// errors
+import "../LSP17ContractExtension/LSP17Errors.sol";
+
 /**
  * @title Core Implementation of ERC725Account
  * @author Fabian Vogelsteller <fabian@lukso.network>, Jean Cavallera (CJ42), Yamen Merhi (YamenMerhi)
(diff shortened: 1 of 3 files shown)


b00ste (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -68,36 +71,12 @@ abstract contract LSP0ERC725AccountCore is
     /**
      * @dev Emits an event when receiving native tokens
      *
-     * Executed when receiving native tokens with empty calldata.
+     * Executed when receiving a call with empty calldata.
```
CJ42, line 74: Maybe we could re-phrase it as bullet points with the cases when this function would run?
To add more precision. Since this function will run when:
- transferring LYX with no calldata
- doing an empty call.

```suggestion
     * Executed:
     *     - when receiving some native tokens without any additional data.
     *     - on empty calls to the contract.
```

skimaharvey (approved):

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -74,12 +75,93 @@ contract LSP9VaultCore is
     /**
      * @dev Emits an event when receiving native tokens
      *
-     * Executed when receiving native tokens with empty calldata.
+     * Executed when receiving a call with empty calldata.
```
CJ42, line 78: same comment as in LSP0

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -111,20 +90,98 @@ abstract contract LSP0ERC725AccountCore is
...
+     * @dev Forwards the call to an extension mapped to a function selector. If no extension address
+     * is mapped to the function selector (address(0)), then revert.
+     *
+     * The bytes4(0) msg.sig is an exception, the function won't revert if there is no extension found
+     * mapped to bytes4(0), but will execute the call to the extension in case it existed.
+     *
+     * The call to the extension is appended with bytes20 (msg.sender) and bytes32 (msg.value).
+     * Returns the return value on success and revert in case of failure.
+     *
+     * As the function uses assembly {return()/revert()} to terminate the call, it cannot be
```
CJ42, line 116: Suggest changing _As_ --> _Because_ or _Since_
```suggestion
     * Because the function uses assembly {return()/revert()} to terminate the call, it cannot be
```

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -111,20 +90,98 @@ abstract contract LSP0ERC725AccountCore is
...
+     * called before other codes in fallback().
+     *
+     * Otherwise, the codes after _fallbackLSP17Extendable() may never be reached.
```
CJ42, line 119: Nice explanation 👌

CJ42 (approved):
added just a suggestion for the Natspec docs on the `receive()` function.

PR #476, opened 2023-02-21
feat!: Add batchCalls function in LSP0 and LSP9

YamenMerhi:
- Add `batchCalls` function in LSP0 and LSP9, to allow executing different function in the same call. 

## ⚠️ BREAKING CHANGES

- Change in interfaceId from `0x66767497` to `0x0f15a0af`
- Change in interfaceId from `0x7050cee9` to `0x19331ad1`


b00ste (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -185,6 +186,33 @@ abstract contract LSP0ERC725AccountCore is
         }
     }
 
+    /**
+     * @dev Receives and executes a batch of function calls on this contract.
+     */
+    function batchCalls(bytes[] calldata data) external returns (bytes[] memory results) {
```
CJ42, line 192: do we want `external` or `public` here? Maybe it would be useful to make it `public` so that if someone want to write an internal function that needs to do a lot of things 

(e.g: 1. make an external, 2. update the storage, etc... )

They could still call the other functions separately, but I wonder.

Also need to think about the implications of calling this function via ERC725X.execute(DELEGATE_CALL, ...).

To be discussed

contracts/Mocks/ERC165Interfaces.sol
```diff
@@ -112,6 +117,7 @@ contract CalculateLSPInterfaces {
     function calculateInterfaceLSP9() public pure returns (bytes4) {
         // prettier-ignore
         bytes4 interfaceId =
+            LSP9VaultCore.batchCalls.selector ^
```
CJ42, line 120: If we include the function selector as part of the interfaceId, then maybe we should create an interface file for LSP0 and LSP9?

> (not sure in the case of LSP17 since there is no interface for it and it is the fallback function.

```solidity
// SPDX-License-Identifier: CC0-1.0
pragma solidity ^0.8.0;

import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
import {IERC725X} from "@erc725/smart-contracts/contracts/interfaces/IERC725X.sol";
import {IERC725Y} from "@erc725/smart-contracts/contracts/interfaces/IERC725Y.sol";
import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
import {ILSP1UniversalReceiver} from "../LSP1UniversalReceiver/ILSP1UniversalReceiver.sol";
import {ILSP14Ownable2Step} from "../LSP14Ownable2Step/ILSP14Ownable2Step.sol";

interface ILSP0ERC725Account is
    IERC165,
    IERC725X,
    IERC725Y,
    IERC1271,
    ILSP1UniversalReceiver,
    ILSP14Ownable2Step
    /* is ILSP17 */
{
    // / @notice Explain to an end user what this does
    // / @dev Explain to a developer any extra details
    // / @param Documents a parameter just like in doxygen (must be followed by parameter name)
    // / @return Documents the return variables of a contract’s function state variable
    // / @inheritdoc	Copies all missing tags from the base function (must be followed by the contract name)
    function batchCalls(bytes[] calldata data) external returns (bytes[] memory results);
}

```

CJ42 (commented):
added first round of review

skimaharvey (approved):
Cool feature 🚀

CJ42 (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
CJ42: ok we mark it as `public`

contracts/Mocks/ERC165Interfaces.sol
CJ42: ok to create an interface for this.

PR #479, opened 2023-02-21
test: add tests for edge cases with LSP1 returned values (e.g: `LSP1: asset sent is not registered`, ...)

CJ42:

## 🧪 Tests

Add more tests in LSP1 for the following edge cases:

- [x] check the returned value of LSP1 when a non-vault contract call the `universalReceiver(...)` function with a LSP9 related typeId.
- [x] check the returned value of LSP5 for cases of `LSP1: asset sent is not registered`, with LSP7 and LSP8.
- [x]  check the returned value of LSP5 for cases of `LSP1: asset sent is not registered`, with LSP7 and LSP8.

Code coverage currently report these returned values for not being covered by our test suite:

image


contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
```diff
@@ -70,92 +70,83 @@ contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiver {
...
+                notifier.code.length > 0 &&
+                !notifier.supportsERC165InterfaceUnchecked(_INTERFACEID_LSP9)
+            ) {
+                return "LSP1: not an LSP9Vault ownership transfer";
             }
-        }
 
-        bytes32 notifierMapKey = LSP2Utils.generateMappingKey(mapPrefix, bytes20(notifier));
-        bytes memory notifierMapValue = IERC725Y(msg.sender).getData(notifierMapKey);
+            (bytes32[] memory dataKeys, bytes[] memory dataValues) = isReceiving
```
YamenMerhi, line 85: I would suggest either reverting this change or refactoring in a way to have this logic:
```js
            // if the map value is already set, then do nothing
            if (bytes12(notifierMapValue) != bytes12(0))
                return "LSP1: asset received is already registered";
                
            // if there is no map value for the asset/vault to remove, then do nothing
            if (bytes12(notifierMapValue) == bytes12(0))
                return "LSP1: asset sent is not registered";
```

Because currently if a vault is not registered and you are trying to transfer it away, it will revert. Because we are not checking if the asset is not registered. Plus to check if the vault is already registered to avoid setting the same data again for the same vault in another element in the array.

YamenMerhi (changes requested):
Will continue reviewing after discussiong/ doing action on the requested change

b00ste (approved):

YamenMerhi (approved):

PR #481, opened 2023-02-22
refactor!: change `CHANGEPERMISSIONS` to `EDITPERMISSIONS`

b00ste:
## ⚠️ BREAKING CHANGE!

Changing the name of the `CHANGEPERMISSIONS` permission to `EDITPERMISSIONS`.


YamenMerhi (changes requested):
We should also change the describe string to match the new name of the permission, at least do a global search and replace.
Check here ⬇️ 
Screen Shot 2023-02-22 at 3 30 56 PM

YamenMerhi (approved):

skimaharvey (approved):

CJ42 (approved):

PR #482, opened 2023-02-24
refactor!: change LSP5/6/10 Array length from `uint256` to `uint128`

CJ42:

## ⚠️ BREAKING CHANGE

- Change the value stored for the Array length from a `uint256` to a `uint128` under the following data keys:

- `LSP5ReceivedAssets[]`
- `LSP10ReceivedAssets[]`
- `AddressPermissions[]`

This follows the breaking change from LSP2: https://github.com/lukso-network/LIPs/pull/180


b00ste (approved):

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -120,13 +120,13 @@ library LSP10Utils {
         IERC725Y account = IERC725Y(sender);
 
         // Updating the number of the received vaults
-        uint256 oldArrayLength = uint256(bytes32(account.getData(_LSP10_VAULTS_ARRAY_KEY)));
+        uint128 oldArrayLength = uint128(bytes16(account.getData(_LSP10_VAULTS_ARRAY_KEY)));
```
skimaharvey, line 123: shouldnt we check that result from `account.getData(_LSP10_VAULTS_ARRAY_KEY)` is not > uint128. Or we can safely assume that it wont be the case ?

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -120,13 +120,13 @@ library LSP10Utils {
...
 
         if (oldArrayLength > type(uint128).max) {
```
skimaharvey, line 125: not sure this check is still relevant since we are casting it above?

contracts/LSP5ReceivedAssets/LSP5Utils.sol
```diff
@@ -80,31 +80,29 @@ library LSP5Utils {
             values[2] = bytes.concat(interfaceID, bytes8(0));
 
             // If the storage is already initiated
-        } else if (encodedArrayLength.length == 32) {
-            uint256 oldArrayLength = uint256(bytes32(encodedArrayLength));
+        } else if (encodedArrayLength.length == 16) {
+            uint128 oldArrayLength = uint128(bytes16(encodedArrayLength));
 
             if (oldArrayLength + 1 >= type(uint64).max) {
                 revert ReceivedAssetsIndexSuperiorToUint64(oldArrayLength);
```
skimaharvey, line 87: might want to remove this check since we changed index from uint64 to uint128?

contracts/LSP5ReceivedAssets/LSP5Utils.sol
```diff
@@ -80,31 +80,29 @@ library LSP5Utils {
             values[2] = bytes.concat(interfaceID, bytes8(0));
 
             // If the storage is already initiated
-        } else if (encodedArrayLength.length == 32) {
-            uint256 oldArrayLength = uint256(bytes32(encodedArrayLength));
+        } else if (encodedArrayLength.length == 16) {
+            uint128 oldArrayLength = uint128(bytes16(encodedArrayLength));
```
skimaharvey, line 84: same comment as above

skimaharvey (approved):
overall looks good to me. Not sure if we should check that the returned value from getData is not > uint128.

contracts/LSP5ReceivedAssets/LSP5Utils.sol
CJ42: @skimaharvey yes we should remove this check. I am planning to change this in a new PR, to change the 2nd tuple values from `bytes8` to `uint128`

https://app.clickup.com/t/8677bq5hj

contracts/LSP10ReceivedVaults/LSP10Utils.sol
CJ42: I don't think it is relevant anymore no because of the changes, but maybe we should have a sanity check to ensure that the values fetched from the Array length are always 16 bytes long (a `uint128`).

The problem is also that if any sanity check is implemented here that makes the code revert, anyone setting something different than 16 bytes under the Array length (for LSP5[]/LSP10[]) would make external calls revert when receiving assets.

So the question is:

- do we add the requirement that it MUST be **exactly** 16 bytes?
- do we add the requirement that it MUST be **at least** 16 bytes?
- do we not add any requirements here? 

This requirement is irrelevant for AddressPermissions[] in LSP6, as we consider this data key optional.
But for LSP5/10, it might be important because the current codebase rely on the value stored under the ArrayLength of LSP5[] + LSP10[] to generate new indexed (when adding/removing assets or vaults).

@YamenMerhi @skimaharvey @b00ste

contracts/LSP10ReceivedVaults/LSP10Utils.sol
CJ42: I don't think it is relevant anymore no because of the changes, but _maybe we should have a sanity check_ to ensure that the values fetched from the Array length are always 16 bytes long (a `uint128`).

The problem is also that if any sanity check is implemented here that makes the code revert, anyone setting something different than 16 bytes under the Array length (for LSP5[]/LSP10[]) would make external calls revert when receiving assets.

So the question is:

- do we add the requirement that it MUST be **exactly** 16 bytes?
- do we add the requirement that it MUST be **at least** 16 bytes?
- do we not add any requirements here? 

This requirement is irrelevant for AddressPermissions[] in LSP6, as we consider this data key optional.
But for LSP5/10, it might be important because the current codebase rely on the value stored under the ArrayLength of LSP5[] + LSP10[] to generate new indexed (when adding/removing assets or vaults).

@YamenMerhi @skimaharvey @b00ste

YamenMerhi: I would say a check make sense because we don't want to override assets

YamenMerhi (approved):

YamenMerhi (approved):

PR #485, opened 2023-02-25
perf: reduce LSP1Delegate deployment cost by 8,678 gas

CJ42:

## 🚀 Perfs

Reduce deployment cost of LSP1DelegateUP by 8.6k gas with the following refactorings:

- Combine if statements with `&` when checking if the map key is related to a vaul.
- use explicit `return` when calling the Key Manager, instead of assigning the return value of the external call (made to the Key Manager) to the `result` variable.
- add local variable `isMapValueSet` to check if the map value is `bytes12(0)` or not, to improve readability.


skimaharvey (approved):

skimaharvey:
Screenshot 2023-02-25 at 19 07 51
since you are returning, do we need to keep in in the stack?

YamenMerhi (approved):
Not sure how this refactoring is increasing the gas cost, but we should really stop favoring deployment cost over runtime cost even if it was a big decrease in deployment and a small increase in runtime, because as we know this contract will only be deployed once and will run twice on each token transfer. 

LGTM 🚀  Let's merge and keep that in mind for our next PRs!

CJ42:
> Screenshot 2023-02-25 at 19 07 51
> 
> since you are returning, do we need to keep in in the stack?

We could remove the variable name, but maybe it might be useful for web3.js/ethers.js if the `universalReceiver(...)` function from LSP1Delegate is called directly via staticcall for instance. The returned object on a dApp would contain the `result` property.

I have removed them from `_whenSending` + `_whenReceiving`, since it's now returned directly and these are `internal` function.

https://github.com/lukso-network/lsp-smart-contracts/pull/485/commits/f58677f7e945fab6be0ffd33c2996f8777e4afc7

@YamenMerhi @skimaharvey

PR #486, opened 2023-02-28
refactor!: replace tuple value for LSP5/10 from `bytes8` -> `uint128`

CJ42:

## :warning: BREAKING CHANGE

- [x] Change the value stored under the LSP5/10 tuples from `bytes4,bytes8)` to --> `(bytes4,uint128)`

- [x] add checks to test when the value under the `LSP5ReceivedAssets[]` and `LSP10Vaults[]` array is the max possible index and introduce custom error to handle this case. 

This is to prevent a `Panic(uint256)` error type with error code `0x11` due to arithmetic overflow with `max(uint128) + 1`

This PR follows the changes introduced in the LSP5 and LSP10 specs. 
See this PR in the LIP repository for reference: https://github.com/lukso-network/LIPs/pull/173

### DIFF
--- a/constants.ts
+++ b/constants.ts
@@ -326,11 +326,6 @@ export const Errors = {
       error: "InvalidLSP5ReceivedAssetsArrayLength(bytes,uint256)",
       message: "LSP5: Invalid value for `LSP5ReceivedAssets[]` (array length)",
     },
-    "0x1c458e39": {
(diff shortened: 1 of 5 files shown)


CJ42:
not forget to remove the custom errors that are not used anymore from the `constants.ts`

YamenMerhi (approved):
LGTM

b00ste (approved):

PR #490, opened 2023-03-06
build: add Types to constant values

CJ42:

Add the Typescript type alongside each values defined inside `constants.ts`.

Co-authored-by: Dominik Zborowski <dominik.zborowski@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ contract MyAccount is LSP0ERC725Account {
 
 You can find more infos on how to deploy the contracts via hardhat in the [DEPLOYMENT](./DEPLOYMENT.md) page.
 
-## Available Constants
+## Available Constants & Types
 
 The [`@lukso/lsp-smart-contracts` npm package](https://www.npmjs.com/package/@lukso/lsp-smart-contracts) contains useful constants such as Interface IDs or ERC725Y Data Keys related to the LSP Standards. You can import and access them as follow:
 
@@ -85,6 +85,23 @@ It also includes constant values [Array data keys](https://github.com/lukso-netw
     length: '0x6460ee3c0aac563ccbf76d6e1d07bada78e3a9514e6382b736ed3f478ab7b90b',
     index: '0x6460ee3c0aac563ccbf76d6e1d07bada',
 },
+```
+
+### Typescript types
+
+The typescript types for the constant values are also available.
+
+```ts
+import {
+  Erc165InterfaceId,
+  erc1271Values,
+  ERC725XOperationTypes,
+  LSP2ArrayKey,
+  SupportedStandardsDataKey,
+  LSP6Permissions,
+  LSP1Hooks,
+  LSPCustomErrors,
+} from "@lukso/lsp-smart-contracts/module/constant.d.ts";
 ```
 
 ## Audits
--- a/README.md
+++ b/README.md
@@ -87,20 +87,15 @@ It also includes constant values [Array data keys](https://github.com/lukso-netw
 },
 ```
 
-### Typescript types
+### Typescript types
 
-The typescript types for the constant values are also available.
+The following additional typescript types are also available.
 
 ```ts
 import {
-  Erc165InterfaceId,
-  erc1271Values,
-  ERC725XOperationTypes,
   LSP2ArrayKey,
-  SupportedStandardsDataKey,
-  LSP6Permissions,
-  LSP1Hooks,
-  LSPCustomErrors,
+  LSPSupportedStandard,
+  LSPErrorInfo,
 } from "@lukso/lsp-smart-contracts/constant";
 ```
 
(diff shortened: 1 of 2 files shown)


constants.ts
```diff
@@ -197,19 +215,23 @@ export const ERC725YDataKeys = {
...
+export const LSP6_VERSION: number = 6;
 
 /**
  * @dev `bytes32` hex value for all the LSP6 permissions excluding REENTRANCY, DELEGATECALL and SUPER_DELEGATECALL for security (these should be set manually)
  */
-export const ALL_PERMISSIONS =
+export const ALL_PERMISSIONS: string =
   "0x00000000000000000000000000000000000000000000000000000000003f3f7f";
 
+export type LSP6Permissions = {
```
dzbo, line 226: I would add `PermissionName` type as it's useful in some cases:

```ts
type PermissionName =
    | 'CHANGEOWNER'
    | 'CHANGEPERMISSIONS'
    | 'ADDPERMISSIONS'
    | 'SETDATA'
...
```

then the `LSP6Permissions` would be:

```ts
export type LSP6Permissions = {
  [key in PermissionName]: string;
};
```

README.md
```diff
@@ -87,6 +87,23 @@ It also includes constant values [Array data keys](https://github.com/lukso-netw
...
+import {
+  Erc165InterfaceId,
+  erc1271Values,
+  ERC725XOperationTypes,
+  LSP2ArrayKey,
+  SupportedStandardsDataKey,
+  LSP6Permissions,
+  LSP1Hooks,
+  LSPCustomErrors,
+} from "@lukso/lsp-smart-contracts/module/constant.d.ts";
```
Esanim, line 104: constant -> constants, right?
I think you don't need `.d.ts` in the import here, but needs to be checked

constants.ts
```diff
@@ -64,12 +73,21 @@ export const OPERATION_TYPES = {
...
+  };
+};
+
 /**
  * @dev list of ERC725Y keys from the LSP standards.
  * Can be used to detect if a contract implements a specific LSP Metadata standard
  * and contain a set of pre-defined ERC725Y Data Keys.
  */
-export const SupportedStandards = {
+export const SupportedStandards: SupportedStandardsDataKey = {
```
richtera, line 90: But this will hide the LSP2ArrayKey types again. I don't think you can have typesdefs on the root items.

constants.ts
```diff
@@ -37,23 +39,30 @@ export const INTERFACE_IDS = {
...
+  MAGIC_VALUE: "0x1626ba7e";
+  FAIL_VALUE: "0xffffffff";
+};
+
 /**
  * @dev values returned by the `isValidSignature` function of the ERC1271 standard.
  * Can be used to check if a signature is valid or not.
  */
-export const ERC1271_VALUES = {
+export const ERC1271_VALUES: erc1271Values = {
```
richtera, line 51: This type is not necessary since the definition is already of that type.

constants.ts
```diff
@@ -6,12 +6,14 @@
...
 
+export type Erc165InterfaceId = { [key: string]: string };
+
 /**
  * @dev ERC165 interface IDs for the LSP interface standards + some backward compatible ERC token standards.
  * These `bytes4` values can be used to detect if a contract implements a specific interface
  * with `supportsInterface(interfaceId)`.
  */
-export const INTERFACE_IDS = {
+export const INTERFACE_IDS: Erc165InterfaceId = {
```
richtera, line 16: This type is not necessary also.

constants.ts
richtera: If you want you can do `export type ERC1271Values = typeof ERC1271_VALUES`

constants.ts
richtera: It would also be better to use typeof

richtera (changes requested):
I think defining the types at the object level will undo the 'as' typings of the children.

constants.ts
```diff
@@ -73,15 +76,15 @@ export const SupportedStandards = {
...
   LSP4DigitalAsset: {
     key: "0xeafec4d89fa9619884b60000a4d96624a38f7ac2d8d9a604ecf07c12c77e480c",
     value: "0xa4d96624",
-  },
+  } as LSPSupportedStandard,
   LSP9Vault: {
     key: "0xeafec4d89fa9619884b600007c0334a14085fefa8b51ae5a40895018882bdb90",
     value: "0x7c0334a1",
-  },
+  } as LSPSupportedStandard,
```
Esanim, line 87: are those 3 `as LSPSupportedStandard` even necessary for anything?

Esanim (approved):

constants.ts
richtera: They cause the same typedef to be used for each of the objects. Otherwise the key, value type pairs will be distinct.
So `typeof SupportedStandards.LSP3UniversalProfile` won't be identical to `typeof SupportedStandards.LSP4DigitalAsset` even though they will both have a key and value property of type string.

richtera (approved):
LGTM. I made a follow up comment.

dzbo (approved):

PR #491, opened 2023-03-06
refactor: remove internal functions and add first check for contract creation permission

CJ42:

in the Key Manager, remove the internal function that extract the permission depending on the operation type.

Simply inline the first check for contract creation and simplify the logic for checking for the permission `CALL` or `STATICCALL`


b00ste (approved):

YamenMerhi (approved):
LGTM

PR #495, opened 2023-03-08
perf: avoid preparing packed data twice on LSP7/8 token transfers

CJ42:

## 🚀 Perfs

In LSP7Core and LSP8Core, on each token transfer, the `packedData` being sent alongside the call to the `universalReceiver` function is generated twice.
- once when notifying token sender.
- once when notifying the token recipient.

Reduce the gas cost of token and NFT transfer by -379 gas by preparing the `lsp1Data` to be sent only once and pass it to the internal function `_notifyTokenSender` and `_notifyTokenReceiver`

### DIFF
--- a/contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol
+++ b/contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol
@@ -226,7 +226,8 @@ abstract contract LSP7DigitalAssetCore is ILSP7DigitalAsset {
 
         emit Transfer(operator, address(0), to, amount, allowNonLSP1Recipient, data);
 
-        _notifyTokenReceiver(address(0), to, amount, allowNonLSP1Recipient, data);
+        bytes memory lsp1Data = abi.encodePacked(address(0), to, amount, data);
+        _notifyTokenReceiver(to, allowNonLSP1Recipient, lsp1Data);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


b00ste (approved):

YamenMerhi (commented):
Looks good to me, but i am confused by the naming and the imports.
Why one is represented by an internal function and one imported from a library.
Why one is named tryNotifyUniversalReceiver and the other notifyTokenReceiver ?

CJ42:
@YamenMerhi because we use the LSP1 library function as a replacement of the `_notifyTokenSender` (only for notifying the token sender).

We cannot use the LSP1 library function for the token receiver, because the logic is different.
If the receiver does not implement LSP1, we have to check the `bool allowNonLSP1Recipient` and revert accordingly.

image

image

CJ42:
This function `_notifyTokenSender`:

image

Was replaced by the LSP1Utils function `tryNotifyUniversalReceiver`

image

We cannot implement the LSP1Utils function for the `_notifyTokenRecipient` because it has to revert if the recipient does not support the LSP1 interface.

We could potentially have a work around this by making the LSP1Utils library function return a `bool`.

image

But in LSP14, LSP0 and LSP9, we would have to ignore this return value. Not sure if this has some security implications. What do you think @skimaharvey @b00ste @YamenMerhi ?

YamenMerhi (approved):

PR #497, opened 2023-03-10
refactor: split execute verification function per operation in KeyManager

CJ42:

## ♻️ Refactor

Improve readability and logic of function `_verifyCanExecute` with internal functions that offer a better code clarity.
Split the checks on multiple functions depending on the operation type.

⚠️ This refactoring does not check anymore for the permission `TRANSFERVALUE` when using operation type `STATICCALL`, since if `value` param in `ERC725X.execute(...)`, ERC725X will revert as not accepting value transfer with operation type STATICCALL.

## 🧪 Tests

Add more tests for the permission `STATICCALL` to ensure:
- call revert with error `ERC725X_MsgValueDisallowedInStaticCall` when sending value alongside the call (**necessary since `TRANSFERVALUE` permission is not checked anymore in the Key Manager when operation type is STATICCALL. So we need to ensure it reverts at ERC725X level**
- add more tests with permission SUPER_STATICCALL or SUPER_CALL + STATICCALL + some allowed addresses.
- add more tests when deploying + funding a contract at the same time (permission DEPLOY + TRANSFER VALUE / SUPER TRANSFER VALUE)


YamenMerhi (changes requested):
Let's add tests for DEPLOY X TRANSFERVALUE, (discussed)

YamenMerhi (approved):

b00ste (approved):

PR #502, opened 2023-03-14
chore: rename function parameters to mixedCase

CJ42:

- Rename some function parameters starting with underscore `_` as spotted by Slither.
- rename LSP6Utils library function `generatePermissionsKeysForController` to --> `generatePermissionsKeysForNewController` for better readability + add Natspec comment above function.

### DIFF
--- a/contracts/LSP6KeyManager/LSP6Utils.sol
+++ b/contracts/LSP6KeyManager/LSP6Utils.sol
@@ -188,7 +188,7 @@ library LSP6Utils {
      *  - values[1] = the address of the controller
      *  - values[2] = the `permissions` passed as param
      */
-    function generatePermissionsKeysForNewController(
+    function generateNewPermissionsKeys(
         IERC725Y account,
         address controller,
         bytes32 permissions
--- a/contracts/LSP6KeyManager/LSP6Utils.sol
+++ b/contracts/LSP6KeyManager/LSP6Utils.sol
@@ -162,26 +162,41 @@ library LSP6Utils {
     /**
      * @dev combine multiple permissions into a single bytes32
      * Make sure that the sum of the values of the input array is less than 2^256-1 to avoid overflow.
-     * @param _permissions the array of permissions to combine
+     * @param permissions the array of permissions to combine
      * @return a bytes32 containing the combined permissions
      */
-    function combinePermissions(bytes32[] memory _permissions) internal pure returns (bytes32) {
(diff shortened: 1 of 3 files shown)


contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -162,26 +162,41 @@ library LSP6Utils {
...
+     *  - keys[0] = `AddressPermissions[]` (array length)
+     *  - keys[1] = `AddressPermissions[index]` (where to store the controller address)
+     *  - keys[2] = `AddressPermissions:Permissions:<controller>`
+     *
+     * @return values : an array of 3 x data values containing:
+     *  - values[0] = the new array length of `AddressPermissions[]`
+     *  - values[1] = the address of the controller
+     *  - values[2] = the `permissions` passed as param
+     */
+    function generatePermissionsKeysForNewController(
```
JeneaVranceanu, line 191: I'd rename it to `generatePermissionsKeysForAddress` or even shorter `generatePermissionsKeys` as it appears this function is capable of handling any kind of address (not only controllers) and any bit array of permissions.

CJ42: The term "controller" here is to be specific and translate _"an address with some permissions"._
I think "controller" gives more context. "Address" is very generic and it is why we changed we also need to make it more clear in the LSP6 Standard page on docs.lukso.tech.

image

In particular what needs to be clear in the name of the function is that it is for **adding a new controller**. See this line below the `+ 1`

image

CJ42: could also be `generateNewControllerPermissionsKeys`.

@YamenMerhi wdyt?

YamenMerhi: I don't mind both

YamenMerhi (approved):

CJ42:
@JeneaVranceanu I went with the name `generateNewPermissionsKeys` for the function, so to take into account your review comment (shorter + simple name) and add the notion that it is for a new controller :)

contracts/LSP6KeyManager/LSP6Utils.sol
JeneaVranceanu: Oh, right. The `arrayLength + 1` is indeed making a huge difference here.
Looks like I didn't pay enough attention last time.

PR #503, opened 2023-03-14
refactor: initialize memory vars before `if` checks in LSP1

CJ42:

Slither report the following when running against `LSP1UniversalReceiverDelegateUP`

Screenshot 2023-03-14 at 13 53 09

although this seems to be a false positive (as we are here dealing with `memory` variables, not `storage` ❌ ), refactor the code to allocate some memory for the array of `dataKeys[]` and `dataValues[]` first before to jump in the `if` checks.

This PR fixes the issue and remove the error reported by Slither.


YamenMerhi (approved):
Approving because of the gas cost optimization

b00ste (approved):

skimaharvey (approved):
Nice one

PR #505, opened 2023-03-15
fix!: require `SUPER_TRANSFERVALUE` permission for deploying contracts with value 

YamenMerhi:

- ⛔️ Require `SUPER_TRANSFERVALUE` permission for deploying contracts with value

### DIFF
--- a/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
+++ b/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
@@ -102,11 +102,9 @@ abstract contract LSP6ExecuteModule {
     ) internal view virtual {
         _requirePermissions(controller, permissions, _PERMISSION_DEPLOY);
 
-        bool hasSuperTransferValue = permissions.hasPermission(_PERMISSION_SUPER_TRANSFERVALUE);
-
         // CHECK if we are funding the contract
-        if (isFundingContract && !hasSuperTransferValue) {
-            _requirePermissions(controller, permissions, _PERMISSION_TRANSFERVALUE);
+        if (isFundingContract) {
+            _requirePermissions(controller, permissions, _PERMISSION_SUPER_TRANSFERVALUE);
         }
     }
 
--- a/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
+++ b/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
@@ -102,9 +102,11 @@ abstract contract LSP6ExecuteModule {
     ) internal view virtual {
         _requirePermissions(controller, permissions, _PERMISSION_DEPLOY);
 
+        bool hasSuperTransferValue = permissions.hasPermission(_PERMISSION_SUPER_TRANSFERVALUE);
+
         // CHECK if we are funding the contract
-        if (isFundingContract) {
-            _requirePermissions(controller, permissions, _PERMISSION_SUPER_TRANSFERVALUE);
+        if (isFundingContract && !hasSuperTransferValue) {
+            revert NotAuthorised(controller, "SUPER_TRANSFERVALUE");
         }
     }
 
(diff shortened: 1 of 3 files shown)


contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
```diff
@@ -102,11 +102,9 @@ abstract contract LSP6ExecuteModule {
     ) internal view virtual {
         _requirePermissions(controller, permissions, _PERMISSION_DEPLOY);
 
-        bool hasSuperTransferValue = permissions.hasPermission(_PERMISSION_SUPER_TRANSFERVALUE);
-
         // CHECK if we are funding the contract
-        if (isFundingContract && !hasSuperTransferValue) {
-            _requirePermissions(controller, permissions, _PERMISSION_TRANSFERVALUE);
+        if (isFundingContract) {
```
CJ42, line 106: I think we can simplify here by keeping the extra conditional check that you deleted '&& !hasSuperTransferValue'.

And revert immediately with 'revert NotAuthorised(...)' inside the if block.

contracts/LSP6KeyManager/LSP6Utils.sol
```diff
@@ -223,6 +223,7 @@ library LSP6Utils {
         if (permission == _PERMISSION_STATICCALL) return "STATICCALL";
         if (permission == _PERMISSION_DELEGATECALL) return "DELEGATECALL";
         if (permission == _PERMISSION_DEPLOY) return "DEPLOY";
+        if (permission == _PERMISSION_SUPER_TRANSFERVALUE) return "SUPER_TRANSFERVALUE";
```
CJ42, line 226: If we do the change I suggested above (keep the conditional check from before), we can remove this newly added line.

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
CJ42: We could also revert with 'NotAuthorised(controller, "FUND ON DEPLOYMENT")' 

Or something like that to be specific and explicit.

tests/LSP6KeyManager/tests/PermissionDeploy.test.ts
```diff
@@ -515,36 +496,16 @@ export const shouldBehaveLikePermissionDeploy = (
...
-      const newUp = new UniversalProfile__factory(context.accounts[0]).attach(
-        preComputedAddress
-      );
-
-      expect(await newUp.owner()).to.equal(initialUpOwner);
-
-      // check that the newly deployed contract (UP) has beedn funded with the correct balance
-      expect(await provider.getBalance(preComputedAddress)).to.equal(
-        fundingAmount
-      );
```
CJ42, line 547: We need to add at least one of the following test to ensure it works and deploy + fund the newly created contract when the caller has:
- DEPLOY + SUPER_TRANSFERVALUE --> deploy via CREATE
- DEPLOY + SUPER_TRANSFERVALUE --> deploy via CREATE2

Can you re-add the same tests from before where we were deploying + funding another UP (using a UP). These were good tests and a very creative use case imo 🙂💜

CJ42 (changes requested):
Added review comments

tests/LSP6KeyManager/tests/PermissionDeploy.test.ts
YamenMerhi: These tests are already existing, added by you [here](https://github.com/lukso-network/lsp-smart-contracts/pull/505/files#diff-d4697674dc7cb0aaa901316c52580dee047e932aab70e5b515ebf6aa5b5f5ec2L551-L729)

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
YamenMerhi: I think it's good to revert with the permission needed,  "FUND ON DEPLOYMENT" doesn't point that you should have `SUPER_TRANSFERVALUE` not `TRANSFERVALUE`

YamenMerhi: I think it's simpler and cheaper to keep it the way it is, because its already pointing to what it does:
- If you are funding the contract ==> require Super_TransferValue permission.
Is the same
- If you are funding the contract ==> you don't have the Super_TransferValue permission revert with ..

tests/LSP6KeyManager/tests/PermissionDeploy.test.ts
CJ42: all good, I had forgotten

CJ42 (approved):

PR #527, opened 2023-03-24
refactor: do not revert when LSP5/LSP10 map data is corrupted

b00ste:

TBD

### DIFF
--- a/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
@@ -42,6 +42,7 @@ import "../LSP1Errors.sol";
  */
 contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiver {
     using ERC165Checker for address;
+    error LengthSmallerThan20Bytes(bytes value);
 
     /**
      * @inheritdoc ILSP1UniversalReceiver
@@ -86,6 +87,7 @@ contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiver {
             return _whenReceiving(typeId, notifier, notifierMapKey, interfaceID);
         } else {
             if (!isMapValueSet) return "LSP1: asset sent is not registered";
+            if (notifierMapValue.length < 20) return "LSP1: asset data corrupted";
 
             return _whenSending(typeId, notifier, notifierMapKey, notifierMapValue);
         }
--- a/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault/LSP1UniversalReceiverDelegateVault.sol
+++ b/contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault/LSP1UniversalReceiverDelegateVault.sol
@@ -84,6 +84,8 @@ contract LSP1UniversalReceiverDelegateVault is ERC165, ILSP1UniversalReceiver {
             // if it's a token transfer (LSP7/LSP8)
             uint256 balance = ILSP7DigitalAsset(notifier).balanceOf(msg.sender);
             if (balance != 0) return "LSP1: full balance is not sent";
+            // if the returned data of lsp5 map is not encoded properly
+            if (notifierMapValue.length < 20) return "LSP1: asset data corrupted";
 
             (dataKeys, dataValues) = LSP5Utils.generateSentAssetKeys(
                 msg.sender,
--- a/tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
+++ b/tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
@@ -1101,8 +1101,8 @@ export const shouldBehaveLikeLSP1Delegate = (
     });
   });
 
-  describe("testing values set under `LSP5ReceivedAssets[]`", () => {
-    let context;
+  describe.only("testing values set under `LSP5ReceivedAssets[]`", () => {
+    let context: LSP1DelegateTestContext;
     let token: LSP7Tester;
 
     before(async () => {
@@ -1239,6 +1239,101 @@ export const shouldBehaveLikeLSP1Delegate = (
       });
     });
 
+    describe("when the Map value of LSP5Map is less than 20 bytes", () => {
+      before(async () => {
+        await token
+          .connect(context.accounts.owner1)
+          .mint(context.accounts.owner1.address, 100, true, "0x");
+
+        await token
+          .connect(context.accounts.owner1)
+          .transfer(
+            context.accounts.owner1.address,
+            context.universalProfile1.address,
+            10,
+            false,
+            "0x"
+          );
+
+        await context.universalProfile1
+          .connect(context.accounts.owner1)
+          ["setData(bytes32,bytes)"](
+            ERC725YDataKeys.LSP5.LSP5ReceivedAssetsMap +
+              token.address.substring(2),
+            "0xcafecafecafecafe"
+          );
+      });
+
+      it("it should revert silently", async () => {
+        const tokenTrasferCalldata = token.interface.encodeFunctionData(
+          "transfer",
+          [
+            context.universalProfile1.address,
+            context.accounts.owner1.address,
+            5,
+            true,
+            "0x",
+          ]
+        );
+
+        expect(
+          await context.universalProfile1
+            .connect(context.accounts.owner1)
+            ["execute(uint256,address,uint256,bytes)"](
+              0,
+              token.address,
+              0,
+              tokenTrasferCalldata
+            )
+        ).to.not.be.reverted;
+      });
+
+      it("it should emit UniversalReceiver event", async () => {
+        const tokenTrasferCalldata = token.interface.encodeFunctionData(
+          "transfer",
+          [
+            context.universalProfile1.address,
+            context.accounts.owner1.address,
+            5,
+            true,
+            "0x",
+          ]
+        );
+
+        const tokensSentBytes32Value = "0x" + "0".repeat(63) + 5;
+
+        const tokenTransferData = (
+          context.universalProfile1.address +
+          context.accounts.owner1.address.substring(2) +
+          tokensSentBytes32Value.substring(2)
+        ).toLocaleLowerCase();
+
+        const lsp1ReturnedData = ethers.utils.defaultAbiCoder.encode(
+          ["string", "bytes"],
+          ["LSP1: asset data corrupted", "0x"]
+        );
+
+        await expect(
+          context.universalProfile1
+            .connect(context.accounts.owner1)
+            ["execute(uint256,address,uint256,bytes)"](
+              0,
+              token.address,
+              0,
+              tokenTrasferCalldata
+            )
+        )
+          .to.emit(context.universalProfile1, "UniversalReceiver")
+          .withArgs(
+            token.address,
+            0,
+            LSP1_TYPE_IDS.LSP7Tokens_SenderNotification,
+            tokenTransferData,
+            lsp1ReturnedData
+          );
+      });
+    });
+
     afterEach(async () => {
       // cleanup adn reset the `LSP5ReceivedAssets[]` length value to 0
       // set the `LSP5ReceivedAssets[]` length value to the max(uint128)`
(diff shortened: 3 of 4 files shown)


contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -130,6 +130,14 @@ library LSP10Utils {
             index
         );
 
+        // If index is bigger than the array length, out of bounds
```
CJ42, line 133: Since we have this check below

image

Why don't we move this entire new block of code in the `else` statement below?

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault/LSP1UniversalReceiverDelegateVault.sol
```diff
@@ -84,13 +84,17 @@ contract LSP1UniversalReceiverDelegateVault is ERC165, ILSP1UniversalReceiver {
             // if it's a token transfer (LSP7/LSP8)
             uint256 balance = ILSP7DigitalAsset(notifier).balanceOf(msg.sender);
             if (balance != 0) return "LSP1: full balance is not sent";
+            // if the returned data of lsp5 map is not encoded properly
```
CJ42, line 87:
```suggestion
            // if the value under the `LSP5ReceivedAssetMap:<asset-address>` 
            // is not a valid tuple as `(bytes4,uint128)`
```

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault/LSP1UniversalReceiverDelegateVault.sol
```diff
@@ -84,13 +84,17 @@ contract LSP1UniversalReceiverDelegateVault is ERC165, ILSP1UniversalReceiver {
...
+            if (notifierMapValue.length < 20) return "LSP1: asset data corrupted";
 
             (dataKeys, dataValues) = LSP5Utils.generateSentAssetKeys(
                 msg.sender,
                 notifierMapKey,
                 notifierMapValue
             );
 
+            if (dataKeys.length == 0 && dataValues.length == 0) return "LSP1: asset data corrupted";
```
CJ42, line 96: I would suggest to add a comment here to explain what is going on.

contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol
```diff
@@ -42,6 +42,7 @@ import "../LSP1Errors.sol";
  */
 contract LSP1UniversalReceiverDelegateUP is ERC165, ILSP1UniversalReceiver {
     using ERC165Checker for address;
+    error LengthSmallerThan20Bytes(bytes value);
```
CJ42, line 45: is this custom error used somewhere?

If yes, I suggest to rename it and put it in `LSP1Errors.sol`

```suggestion
    error MapValueSmallerThan20Bytes(bytes value);
```

> Note: ideally we should have two custom errors: one for LSP5, one for LSP10 to give better context.

CJ42 (commented):
added 1st round of review

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+
+        await context.universalProfile1
+          .connect(context.accounts.owner1)
+          ["setData(bytes32,bytes)"](
+            ERC725YDataKeys.LSP5.LSP5ReceivedAssetsMap +
+              token.address.substring(2),
+            "0xcafecafecafecafe"
+          );
+
+        expect(
```
CJ42, line 1278: maybe we can put all of this in a separate `it("should have setup the LS5 data keys (included corrupted one correctly")` block? It might be easier to debug.

or maybe we can add a description string for the `beforeEach("setup LSP5 data keys (including corrupted one)")`

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+            true,
+            "0x",
+          ]
+        );
+
+        expect(
+          await context.universalProfile1
+            .connect(context.accounts.owner1)
+            ["execute(uint256,address,uint256,bytes)"](
+              0,
```
CJ42, line 1312:
```suggestion
              OPERATIONS_TYPE.CALL,
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+        const lsp1ReturnedData = ethers.utils.defaultAbiCoder.encode(
+          ["string", "bytes"],
+          ["LSP1: asset data corrupted", "0x"]
+        );
+
+        await expect(
+          context.universalProfile1
+            .connect(context.accounts.owner1)
+            ["execute(uint256,address,uint256,bytes)"](
+              0,
```
CJ42, line 1356: same as above

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+            lsp1ReturnedData
+          );
+      });
+
+      it("shouldn't de-register the asset", async () => {
+        const balance = await token.balanceOf(
+          context.universalProfile1.address
+        );
+
+        const tokenTrasferCalldata = token.interface.encodeFunctionData(
```
CJ42, line 1377: typo 
```suggestion
        const tokenTransferCalldata = token.interface.encodeFunctionData(
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+          ["execute(uint256,address,uint256,bytes)"](
+            0,
+            token.address,
+            0,
+            tokenTrasferCalldata
+          );
+
+        expect(
+          await context.universalProfile1["getData(bytes32[])"]([
+            ERC725YDataKeys.LSP5["LSP5ReceivedAssets[]"].length,
```
CJ42, line 1399: maybe you can put these data keys in a `const` array at the top? So that you don't have to re-write them again to avoid potential human typo?

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+          ])
+        ).to.deep.equal([
+          "0x" + "00".repeat(15) + "01",
+          token.address.toLowerCase(),
+          "0xcafecafecafecafe",
+        ]);
+      });
+    });
+
+    describe("when the Map value of LSP5Map is bigger than 20 bytes, (20 valid bytes + extra data)", () => {
```
CJ42, line 1412:
```suggestion
    describe("when the Map value of LSP5Map is bigger than 20 bytes, (valid `(byte4,uint128)` tuple  + extra bytes)", () => {
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+            true,
+            "0x",
+          ]
+        );
+
+        expect(
+          await context.universalProfile1
+            .connect(context.accounts.owner1)
+            ["execute(uint256,address,uint256,bytes)"](
+              0,
```
CJ42, line 1460: same comment as above

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+        const tokensSentBytes32Value = ethers.utils.hexZeroPad(
+          balance.toHexString(),
+          32
+        );
+
+        const tokenTransferData = (
+          context.universalProfile1.address +
+          context.accounts.owner1.address.substring(2) +
+          tokensSentBytes32Value.substring(2)
+        ).toLocaleLowerCase();
```
CJ42, line 1493: Why `Locale`? haha

```suggestion
        ).toLowerCase();
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+            balance,
+            true,
+            "0x",
+          ]
+        );
+
+        await context.universalProfile1
+          .connect(context.accounts.owner1)
+          ["execute(uint256,address,uint256,bytes)"](
+            0,
```
CJ42, line 1539: same as above, use `OPERATIONS_TYPE` here to improve readability of the test

```suggestion
            OPERATIONS_TYPE.CALL,
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+            ERC725YDataKeys.LSP5["LSP5ReceivedAssets[]"].length,
+            ERC725YDataKeys.LSP5["LSP5ReceivedAssets[]"].index + "0".repeat(32),
+            ERC725YDataKeys.LSP5.LSP5ReceivedAssetsMap +
+              token.address.substring(2),
+          ])
+        ).to.deep.equal(["0x" + "00".repeat(16), "0x", "0x"]);
+      });
+    });
+
+    describe("when the Map value of LSP5Map is equal to 20 invalid bytes", () => {
```
CJ42, line 1556: What should be the description exactly? I wonder if it should be "20 random bytes" or "20 invalid bytes".


```suggestion
    describe("when the Map value of LSP5ReceivedAssetsMap is 20 random bytes", () => {
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+        );
+
+        const tokenTransferData = (
+          context.universalProfile1.address +
+          context.accounts.owner1.address.substring(2) +
+          tokensSentBytes32Value.substring(2)
+        ).toLocaleLowerCase();
+
+        const lsp1ReturnedData = ethers.utils.defaultAbiCoder.encode(
+          ["string", "bytes"],
```
CJ42, line 1655: This line is great. Did not think about encoding as `string`. That makes the test super readable 👌

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+          .connect(context.accounts.owner1)
+          ["execute(uint256,address,uint256,bytes)"](
+            0,
+            vault.address,
+            0,
+            vaultTrasferCalldata
+          );
+
+        await expect(
+          await vault.connect(context.accounts.owner1).acceptOwnership()
```
CJ42:
```suggestion
          vault.connect(context.accounts.owner1).acceptOwnership()
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,845 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+      // start with a fresh empty context
+      context = await buildContext();
+
+      vault = await new LSP9Vault__factory(context.accounts.random).deploy(
+        context.accounts.owner1.address
+      );
+    });
+
+    describe("when the Map value of LSP10Map is less than 20 bytes", () => {
+      beforeEach(async () => {
```
CJ42, line 1734: Let's try not do `beforeEach` for the Vault.

We can simply:
1. save the transaction for accepting ownership in a variable `acceptOwnershipTx`
2. use this variable to test it against the different `it()` tests.

CJ42 (commented):
added review comments for the tests

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,839 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+        const tokensSentBytes32Value = ethers.utils.hexZeroPad(
+          balance.toHexString(),
+          32
+        );
+
+        const tokenTransferData = (
+          context.universalProfile1.address +
+          context.accounts.owner1.address.substring(2) +
+          tokensSentBytes32Value.substring(2)
+        ).toLocaleLowerCase();
```
CJ42, line 1345:
```suggestion
        ).toLowerCase();
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
CJ42: See screenshot below:

image

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault.behaviour.ts
```diff
@@ -805,6 +805,557 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+          "0x" + "00".repeat(15) + "01",
+          token.address.toLowerCase(),
+          "0xcafecafecafecafe",
+        ]);
+      });
+
+      it("it should pass", async () => {
+        const balance = await token.balanceOf(context.lsp9Vault1.address);
+
+        const tokenTrasferCalldata = token.interface.encodeFunctionData(
```
CJ42, line 865:
```suggestion
        const tokenTransferCalldata = token.interface.encodeFunctionData(
```

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateVault.behaviour.ts
```diff
@@ -805,6 +805,557 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+            balance,
+            true,
+            "0x",
+          ]
+        );
+
+        const vaultTokenTransferCalldata =
+          context.lsp9Vault1.interface.encodeFunctionData(
+            "execute(uint256,address,uint256,bytes)",
+            [0, token.address, 0, tokenTrasferCalldata]
```
CJ42, line 967:
```suggestion
            [OPERATIONS_TYPE.CALL, token.address, 0, tokenTrasferCalldata]
```

CJ42 (commented):
final comments for the tests

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
```diff
@@ -1239,18 +1261,839 @@ export const shouldBehaveLikeLSP1Delegate = (
...
+              token.address.substring(2),
+          ])
+        ).to.deep.equal([
+          "0x" + "00".repeat(15) + "01",
+          token.address.toLowerCase(),
+          "0xcafecafecafecafecafecafecafecafecafecafe",
+        ]);
+      });
+
+      afterEach(async () => {
```
CJ42, line 1584: if we apply the suggestion above (replacing the `beforeEach` by `before`), we can remove this `afterEach` block.

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -130,6 +130,11 @@ library LSP10Utils {
             index
         );
 
+        // If the address at index doesn't match the notifier address, corrupted data
+        bytes20 addressAtIndex = bytes20(account.getData(vaultInArrayKey));
+        bytes20 notifier = bytes20(vaultMapKey << 96);
+        if (notifier != addressAtIndex) return (keys, values);
+
         if (index == newArrayLength) {
```
CJ42, line 138: shouldn't we add a comment here to better explain? Like:

```
// if we are removing the last entry in the array
```

contracts/LSP10ReceivedVaults/LSP10Utils.sol
```diff
@@ -151,7 +156,7 @@ library LSP10Utils {
 
             // Swapping last element in ArrayKey with the element in ArrayKey to remove || {Swap and pop} method;
             // check https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L80
-        } else {
+        } else if (index < newArrayLength) {
```
CJ42, line 159: add comment:

```
// if we are removing another entry within the array (the first or another element)
```

contracts/LSP10ReceivedVaults/LSP10Utils.sol
CJ42: @b00ste please ignore this comment. @YamenMerhi did it in his PR.

tests/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP.behaviour.ts
b00ste: I don't think we should have a separate test because we are just making sure that we are constructing the right context. I would keep the expect inside the before, if it fails it is displayed properly that it fails in the `before` so it is really easy to debug

YamenMerhi (approved):

CJ42 (approved):

PR #557, opened 2023-04-03
feat: Allow calling `ERC725X.execute(uint256[],address[],uint256[],bytes[])` through the LSP6

b00ste:

This PR introduces the path `ERC725X.execute(uint256[],address[],uint256[],bytes[])` to LSP6KeyManager.

## 🚀 Feature

Previously calling `ERC725X.execute(uint256[],address[],uint256[],bytes[])` would revert with the custom error `InvalidERC725Function`. Now the call goes to the `target` contract. 

### PR Checklist

- [ ] Wrote Tests
- [ ] Wrote Documentation
- [x] Ran `npm run linter`
- [x] Ran `npm run prettier`
- [x] Ran `npm run build`
- [x] Ran `npm run test`

### DIFF
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -158,3 +158,9 @@ error CannotSendValueToSetData();
  * @dev reverts when calling the KeyManager through execute(..)
  */
 error CallingKeyManagerNotAllowed();
+
+/**
+ * @dev reverts when the parameters of `ERC725X.execute(uint256[],address[],uint256[],bytes[])`
+ * have diferent lengths
+ */
+error ERC725XBatchExecuteArrayParamsLengthMissmatch();
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -158,9 +158,3 @@ error CannotSendValueToSetData();
  * @dev reverts when calling the KeyManager through execute(..)
  */
 error CallingKeyManagerNotAllowed();
-
-/**
- * @dev reverts when the parameters of `ERC725X.execute(uint256[],address[],uint256[],bytes[])`
- * have diferent lengths
- */
-error ERC725XBatchExecuteArrayParamsLengthMissmatch();
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -43,7 +43,8 @@ import {
 import {
     SETDATA_SELECTOR,
     SETDATA_ARRAY_SELECTOR,
-    EXECUTE_SELECTOR
+    EXECUTE_SELECTOR,
+    EXECUTE_ARRAY_SELECTOR
 } from "@erc725/smart-contracts/contracts/constants.sol";
 import {
     _INTERFACEID_ERC1271,
@@ -394,6 +395,10 @@ abstract contract LSP6KeyManagerCore is
             // ERC725X.execute(uint256,address,uint256,bytes)
         } else if (erc725Function == EXECUTE_SELECTOR) {
             LSP6ExecuteModule._verifyCanExecute(_target, from, permissions, payload);
+
+            // ERC725X.execute(uint256,address,uint256,bytes)
+        } else if (erc725Function == EXECUTE_ARRAY_SELECTOR) {
+            LSP6ExecuteModule._verifyCanBatchExecute(_target, from, permissions, payload);
         } else if (
             erc725Function == ILSP14Ownable2Step.transferOwnership.selector ||
             erc725Function == ILSP14Ownable2Step.acceptOwnership.selector
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -36,7 +36,8 @@ import {
     InvalidERC725Function,
     NotAuthorised,
     CallerIsNotTheTarget,
-    CannotSendValueToSetData
+    CannotSendValueToSetData,
+    ERC725XBatchExecuteArrayParamsLengthMissmatch
 } from "./LSP6Errors.sol";
 
 // constants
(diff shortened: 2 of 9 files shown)


b00ste:
LSP6KeyManager deployment cost 3298129 => 3724578

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
```diff
@@ -35,7 +35,8 @@ import {
     OPERATION_1_CREATE,
     OPERATION_2_CREATE2,
     OPERATION_3_STATICCALL,
-    OPERATION_4_DELEGATECALL
+    OPERATION_4_DELEGATECALL,
+    EXECUTE_SELECTOR
```
CJ42, line 39: This can be removed once we have refactored to use the internal functions in a for loop

CJ42 (changes requested):
The reason there is a 400k gas difference is because there 5 or 6 x internal functions that are duplicated to use the "memory" keyword (as currently it is calldata).

We need to:
1) use "memory" for the internal functions instead of "calldata", (but keep "calldata" for the
function implementations of the public interface).

2) refactor the code to not have to pass "bytes" to the internal functions but instead the decoded parameters. We should not have to reconstruct a "singleExecute" payload. This is very inefficient especially in a loop.

To implement this batch feature, the current internal functions of the LSP6 Execute module (that are currently used to verify a single payload) need a bit of refactoring.

contracts/LSP6KeyManager/LSP6Errors.sol
```diff
@@ -158,3 +158,9 @@ error CannotSendValueToSetData();
  * @dev reverts when calling the KeyManager through execute(..)
  */
 error CallingKeyManagerNotAllowed();
+
+/**
+ * @dev reverts when the parameters of `ERC725X.execute(uint256[],address[],uint256[],bytes[])`
+ * have diferent lengths
+ */
+error ERC725XBatchExecuteArrayParamsLengthMissmatch();
```
CJ42, line 166: Let's import and reuse the one from the @erc725 repo, like for the setData([],[]) path.
No point duplicating this definition, as there will be 2 x error selectors for exactly the same thing.

image

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
```diff
@@ -59,29 +59,18 @@ abstract contract LSP6ExecuteModule {
...
      * @param permissions the permissions of the controller address.
-     * @param payload the ABI encoded payload `controlledContract.execute(...)`.
      */
     function _verifyCanExecute(
         address controlledContract,
         address controller,
         bytes32 permissions,
-        bytes calldata payload
+        uint256 operationType,
+        address callee,
```
CJ42, line 68: I think "to" was a better name. callee makes sense but is a very specific Smart Contract technical term.
Only smart contract devs will understand this term.
Let's revert to the previous name.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -393,7 +395,55 @@ abstract contract LSP6KeyManagerCore is
...
+            );
+
+            // ERC725X.execute(uint256,address,uint256,bytes)
+        } else if (erc725Function == EXECUTE_ARRAY_SELECTOR) {
+            (
+                uint256[] memory operationTypes,
+                address[] memory callees,
+                uint256[] memory values,
+                bytes[] memory datas
+            ) = abi.decode(payload[4:], (uint256[], address[], uint256[], bytes[]));
```
CJ42, line 428: Hmmm this is more tricky that I thought and I did not see it coming.

Because now with this refactoring, we can see that we are not ensuring anymore that for every single payload in a batch ERC725X.execute([],[],[],[]), the offset of the fourth "data" parameter is always 0x...80.

image

We should ensure this is as well, not only for the single ERC725X.execute(uint256,address,ui t256,bytes).

Maybe we could try an other alternative approach.

_alternative 1:_ by looping through calldata slices.
The starting offset on every iterations could be easy to calculate. The ending offset could be a bit more harder but feasible.

_alternative 2:_

Or we could combine both: abi-decoding the four arrays but having a pre-check as well on the calldata at the start of each iterations.

CJ42: Actually maybe for bytes[] the offset check is not necessary.

But since there are 4 x arrays now (for each params list), there are probably more offsets to be checked for safety.

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
```diff
@@ -179,7 +192,7 @@ abstract contract LSP6ExecuteModule {
         // + 32 (`data` length)
         // --------------------
         // = 164 bytes in total
-        bool isCallDataPresent = payload.length > 164;
+        bool isCallDataPresent = data.length > 0;
```
CJ42, line 195: this read better this way 👌🏻

contracts/Mocks/KeyManager/KeyManagerInternalsTester.sol
```diff
@@ -29,7 +29,12 @@ contract KeyManagerInternalTester is LSP6KeyManager {
     }
 
     function verifyAllowedCall(address _sender, bytes calldata _payload) public view {
-        super._verifyAllowedCall(_target, _sender, _payload);
+        (uint256 operationType, address callee, uint256 value, bytes memory data) = abi.decode(
```
CJ42, line 32: ⚠️❌ I am blocking this PR as this kind of diff should NEVER happen.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -393,7 +395,55 @@ abstract contract LSP6KeyManagerCore is
...
-            LSP6ExecuteModule._verifyCanExecute(_target, from, permissions, payload);
+            // CHECK the offset of `data` is not pointing to the previous parameters
+            if (
+                bytes32(payload[100:132]) !=
+                0x0000000000000000000000000000000000000000000000000000000000000080
+            ) {
+                revert InvalidPayload(payload);
+            }
+
+            (uint256 operationType, address callee, uint256 value, bytes memory data) = abi.decode(
```
CJ42, line 406: since abi.decode reverts in Solidity 0.8.0, it would be good to add more unit tests for this internal function to be aware the different kind of behaviours and payloads that would make abi-decode revert or not.

We need to make more research and investigations as we always brought up with @skimaharvey and not go straight in with these code changes.

> we should also do the same for setData that also use abi.decode. Have more unit tests for the internal functions that verify setData payloads.

CJ42 (changes requested):
This new feature needs more investigation and research. Blocking for now ❌🔎

CJ42:
For reference, on Ethereum, this increase in deployment cost would result in the following:

before: 3,298,128
after: 3,420,298

+ 122,170
 
x 50 (gas price) = 6,108,500 gwei = 0.0061085 ETH

1 ETH = 1,770 EUR
1 ETH = 1,940 USD

0.0061085 = 10.81 EUR
0.0061085 = 11.85 USD

CJ42:
Closing this PR as we need more investigations for potential bugs for this path + review the gas increase (both deployment + runtime)

PR #562, opened 2023-04-14
fix: [TRST-M-5] `renounceOwnership()` will accidentally be one-step in a certain time period

b00ste:

This PR introduces a fix for a bug that occurs on new EVM blockchains, when someone intends to use LSP14.

## 🐛 Bug

In the current Implementation if the `block.number` is smaller than 400, then `renounceOwnership(..)` starts with the second step, bypassing the first step.

Fixes #???? 

### PR Checklist

- [x] Wrote Tests
- [ ] Wrote Documentation
- [x] Ran `npm run linter`
- [x] Ran `npm run prettier`
- [x] Ran `npm run build`
- [x] Ran `npm run test`


tests/LSP14Ownable2Step/LSP14Ownable2Step.behaviour.ts
```diff
@@ -609,4 +609,37 @@ export const shouldBehaveLikeLSP14 = (
...
     });
   });
+
+  describe.only("should be able to have 2-step `renoucneOwnership()` on a fresh blockchain", () => {
+    before(async () => {
+      context = await buildContext();
+
+      // Skip 1 block
+      // Simulate real blockchain
+      // You would never be able to deploy a contract and initiate `renounceOwnership(..)`
```
CJ42, line 619: Good test and comments. Just suggested some rewording


```suggestion
      // in a real blockchain environment, you would not likely be able to deploy a contract 
      // and initiate `renounceOwnership(..)` in the block 0
```

tests/LSP14Ownable2Step/LSP14Ownable2Step.behaviour.ts
```diff
@@ -609,4 +609,37 @@ export const shouldBehaveLikeLSP14 = (
...
+    it("should instantiate the renounceOwnership process correctly", async () => {
+      const renounceOwnershipTx = await context.contract
+        .connect(context.deployParams.owner)
+        .renounceOwnership();
+
+      await renounceOwnershipTx.wait();
+
+      const _renounceOwnershipStartedAtAfter = await provider.getStorageAt(
+        context.contract.address,
+        2
```
CJ42, line 633: Are we sure this state variable is at this storage slot?
Maybe we use this in the other tests, but this has changed since we have added more changes?

I would suggest to get the slot number where this state variable is defined programmatically, rather than a hardcoded literal.
By looking through the build-info of the contract.
There might be some content in the JSON file (either the AST, or something else) that contains the storage layout where we can grab this info.

This is required to ensure the correctness of the test.
I am happy to help and assist with this.

tests/LSP14Ownable2Step/LSP14Ownable2Step.behaviour.ts
```diff
@@ -609,4 +609,37 @@ export const shouldBehaveLikeLSP14 = (
...
+      );
+
+      expect(
+        ethers.BigNumber.from(_renounceOwnershipStartedAtAfter).toNumber()
```
CJ42, line 637: Is the `toNumber()` needed? Does the test fail if you remove it?

I think chai with Hardhat handle comparisons between regular numbers and big numbers.

Try to remove it and see if the tests still pass:

```suggestion
        ethers.BigNumber.from(_renounceOwnershipStartedAtAfter)
```

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
```diff
@@ -152,7 +152,11 @@ abstract contract LSP14Ownable2Step is ILSP14Ownable2Step, OwnableUnset {
         uint256 confirmationPeriodEnd = confirmationPeriodStart +
             RENOUNCE_OWNERSHIP_CONFIRMATION_PERIOD;
 
-        if (currentBlock > confirmationPeriodEnd) {
+        bool isNewNetwork = currentBlock <
+            RENOUNCE_OWNERSHIP_CONFIRMATION_DELAY + RENOUNCE_OWNERSHIP_CONFIRMATION_PERIOD &&
```
CJ42, line 156: Let's be VERY careful here!

Order of precedence of operators is very odd in Solidity.

https://docs.soliditylang.org/en/v0.8.19/types.html#order-of-precedence-of-operators

Let's put groups to compare in parentheses + define the a local variable.

> NB: `renounceOwnershipCanStartAt` could be better named maybe? 

```suggestion
            uint256 renounceOwnershipCanStartAt = RENOUNCE_OWNERSHIP_CONFIRMATION_DELAY + RENOUNCE_OWNERSHIP_CONFIRMATION_PERIOD;
             bool isNewNetwork = (currentBlock <
            renounceOwnershipCanStartAt) && (_renounceOwnershipStartedAt == 0)
```

CJ42: I am also not sure about this fix.
@YamenMerhi @skimaharvey we can look at this later

CJ42 (changes requested):

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
CJ42: I really don't get it at all. Why overcomplicating with this complex local variable when the recommendation is very straight forward?

image

CJ42 (approved):

YamenMerhi (approved):

PR #573, opened 2023-04-26
refactor: [TRST-M-1 LSP20] LSP20 fix accepted magic value, 4 bytes followed by 28 zero bytes

b00ste:

## ♻️ Refactor

Change the values accepted by LSP20  from magic value right padded by any 28+ bytes to magic value padded by 28 zero bytes and after that any bytes.

Fixes #???? 

### PR Checklist

- [ ] Wrote Tests
- [ ] Wrote Documentation
- [x] Ran `npm run linter`
- [x] Ran `npm run prettier`
- [x] Ran `npm run build`
- [x] Ran `npm run test`


contracts/LSP20CallVerification/LSP20CallVerification.sol
```diff
@@ -49,12 +47,25 @@ abstract contract LSP20CallVerification {
...
+        // check if the returned data contains at least 32 bytes, potentially an abi encoded bytes4 value
+        if (returnedData.length < 32) revert LSP20InvalidMagicValue(postCall, returnedData);
 
-        if (
-            returnedData.length < 32 ||
-            bytes4(abi.decode(returnedData, (bytes32))) != ILSP20.lsp20VerifyCallResult.selector
-        ) revert LSP20InvalidMagicValue(true, returnedData);
+        // check if the returned data has in the first 32 bytes an abi encoded bytes4 value
+        if (bytes28(bytes32(returnedData) << 32) != bytes28(0))
+            revert LSP20InvalidMagicValue(postCall, returnedData);
```
skimaharvey, line 68: cant we do it in one check?
```
if (returnedData.length < 32 || bytes28(bytes32(returnedData) << 32) != bytes28(0)) revert LSP20InvalidMagicValue(postCall, returnedData);
```

skimaharvey: and 
```
if (returnedData.length < 32 || bytes32(returnedData) << 32 != bytes32(0) revert LSP20InvalidMagicValue(postCall, returnedData);
```

b00ste: yeah, but you have the same checks in both functions, I figured it would be better to have them in an internal function to not repeat the code

skimaharvey: i think the internal function is fine but i think we could still to one `if` with the two conditions.
not sure what @YamenMerhi @CJ42 think

skimaharvey (approved):
LGTM

YamenMerhi (approved):

contracts/LSP20CallVerification/LSP20CallVerification.sol
```diff
@@ -49,12 +47,23 @@ abstract contract LSP20CallVerification {
...
+        bool success,
+        bytes memory returnedData
+    ) internal pure {
+        if (!success) _revert(postCall, returnedData);
 
-        if (
-            returnedData.length < 32 ||
-            bytes4(abi.decode(returnedData, (bytes32))) != ILSP20.lsp20VerifyCallResult.selector
-        ) revert LSP20InvalidMagicValue(true, returnedData);
+        // check if the returned data contains at least 32 bytes, potentially an abi encoded bytes4 value
```
CJ42, line 63: why "potentially"?

contracts/LSP20CallVerification/LSP20CallVerification.sol
CJ42: one check like this looks good to me

tests/LSP20CallVerification/LSP20CallVerification.behaviour.ts
```diff
@@ -521,6 +521,43 @@ export const shouldBehaveLikeLSP20 = (
         });
       });
 
+      describe("that implements verifyCall that returns a valid magicValue with additional data after the first 32 bytes", () => {
+        let firstCallReturnMagicValueContract: FakeContract;
+        let newUniversalProfile: UniversalProfile;
+
+        before(async () => {
+          firstCallReturnMagicValueContract = await smock.fake(
```
CJ42, line 529: nice 👌

contracts/LSP20CallVerification/LSP20CallVerification.sol
b00ste: Because when you check for length it is not 100% an abi encoded value, we check that it's abi encoded in the second check, if the first 4 byte are anything an the next 28 bytes are zero bytes

PR #576, opened 2023-04-27
fix: [TRST-H-1] Disallow turning reentrancy bool to false in reentrantCalls

YamenMerhi:
## 🐛 Bug

Screen Shot 2023-04-27 at 5 37 15 PM

### Explanation

The bug did not enable an attacker to execute an unauthorized action, but rather it permitted the execution of an action that the attacker was authorized to perform in a reentrant call without requiring Reentrancy Permission. This was possible by chaining the call with previous execute calls that deactivated the reentrancy guard.

### Fix 

The problem was fixed by making the function always tell if the call is a normal call or a reentrant call. If a reentrant call is detected, the function will not turn off the reentrancy guard/require a post-execution.

### PR Checklist

- [x] Wrote Tests
- [ ] Wrote Documentation **N/A**
- [X] Ran `npm run linter` (solhint)
- [X] Ran `npm run format` (prettier)
- [X] Ran `npm run build`
- [X] Ran `npm run test`

### DIFF
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -230,14 +230,14 @@ abstract contract LSP6KeyManagerCore is
             isSetData = true;
         }
 
-        bool reentrantCall = _nonReentrantBefore(isSetData, caller);
+        bool isReentrantCall = _nonReentrantBefore(isSetData, caller);
 
         _verifyPermissions(caller, msgValue, data);
         emit VerifiedCall(caller, msgValue, bytes4(data));
 
         // if it's a setData call, do not invoke the `lsp20VerifyCallResult(..)` function
         return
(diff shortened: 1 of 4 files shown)


contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -230,14 +230,14 @@ abstract contract LSP6KeyManagerCore is
             isSetData = true;
         }
 
-        _nonReentrantBefore(isSetData, caller);
+        bool reentrantCall = _nonReentrantBefore(isSetData, caller);
```
b00ste, line 233: Maybe like this would be better for readability

```suggestion
        bool isReentrantCall = _nonReentrantBefore(isSetData, caller);
```

b00ste (approved):

skimaharvey (approved):

PR #578, opened 2023-05-02
chore: LSP14 remove line that will never be executed

b00ste:

## ♻️ Refactor

Removing the line `delete pendingOwner;` which is executed in the second step of the `renounceOwnership()`. This code is being executed twice, once in the first step of `renounceOwnership()` and the second time in the second phase of `renounceOwnership()`.

Because the processes of transferring/renouncing ownership cancel each other it makes it impossible to set a new `pendingOwner` and then do the second phase of renouncing ownership, as setting a new `pendingOwner` is made through transferring ownership which will cancel the process of renouncing ownerhip

Fixes #???? 

### PR Checklist

- [ ] Wrote Tests
- [ ] Wrote Documentation
- [x] Ran `npm run linter` (solhint)
- [x] Ran `npm run build`
- [x] Ran `npm run test`


CJ42 (approved):
I think there will always be the edge case with delegatecall between a renounce / transfer ownership (to override the storage).
But apart from that, lgtm.

YamenMerhi (approved):

PR #583, opened 2023-05-03
chore: [TRST-R-5] return `FAILVALUE` when the signature is invalid

YamenMerhi:

## ♻️ Chore

- Avoid reverts and return `FAILVALUE` when the signature is invalid in LSP0 and LSP6

### PR Checklist

- [X] Wrote Tests
- [ ] Wrote Documentation
- [X] Ran `npm run linter` (solhint)
- [X] Ran `npm run format` (prettier)
- [X] Ran `npm run build`
- [X] Ran `npm run test`

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -699,10 +699,18 @@ abstract contract LSP0ERC725AccountCore is
         }
         // If owner is an EOA
         else {
-            return
-                _owner == ECDSA.recover(dataHash, signature)
-                    ? _ERC1271_MAGICVALUE
-                    : _ERC1271_FAILVALUE;
+            // if isValidSignature fail, the error is catched in returnedError
+            (address recoveredAddress, ECDSA.RecoverError returnedError) = ECDSA.tryRecover(
+                dataHash,
+                signature
+            );
+
+            // if recovering throws an error, return the fail value
+            if (returnedError != ECDSA.RecoverError.NoError) return _ERC1271_FAILVALUE;
+
+            // if recovering is successfull, return the magic value if the recovered address
+            // matches the address of the owner, otherwise return fail value
+            return recoveredAddress == _owner ? _ERC1271_MAGICVALUE : _ERC1271_FAILVALUE;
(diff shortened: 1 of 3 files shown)


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -699,10 +699,18 @@ abstract contract LSP0ERC725AccountCore is
...
+            // if isValidSignature fail, the error is catched in returnedError
+            (address recoveredAddress, ECDSA.RecoverError returnedError) = ECDSA.tryRecover(
+                dataHash,
+                signature
+            );
+
+            // if recovering throws an error, return the fail value
+            if (returnedError != ECDSA.RecoverError.NoError) return _ERC1271_FAILVALUE;
+
+            // if recovering is successfull, return the magic value if the recovered address
```
CJ42, line 711: suggested changes for the comment

```suggestion
            // if recovering is successful and the recovered address matches the owner's address,
            // return the ERC1271 magic value. Otherwise, return the ERC1271 fail value
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -114,8 +114,16 @@ abstract contract LSP6KeyManagerCore is
...
+        // if isValidSignature fail, the error is catched in returnedError
+        (address recoveredAddress, ECDSA.RecoverError returnedError) = ECDSA.tryRecover(
+            dataHash,
+            signature
+        );
+
+        // if recovering throws an error, return the fail value
+        if (returnedError != ECDSA.RecoverError.NoError) return _ERC1271_FAILVALUE;
 
+        // if the address recovered has SIGN permission return the magicValue, failValue otherwise
```
CJ42, line 126: Not sure if needed but maybe we can specify it is the magic / fail value of ERC1271 for precision.
Also suggesting to write these as plain text (`magic value`), not like the variable names (`magicValue`)

```suggestion
        // if the address recovered has SIGN permission return the ERC1271 magic value, otherwise the fail value
```

CJ42 (approved):

b00ste (approved):

PR #601, opened 2023-06-02
fix: Allow direct import of all files in dist

richtera:

Allow direct import of constants.es.js and constants.cjs.js in the dist folder.
This will cause an error in typescript because of constants.d.ts not being found,
but hardhat uses the node loader and is not capable of finding the right file to load.
It uses ts-node which decides to pick the .es.js file, but that has "import" statements
which are not compatible with the nodejs compiler. I.e. it assumes modules
use require/module instead of import/export.


skimaharvey (approved):

CJ42 (approved):

richtera:
Rebased so that only the single line change shows I  PR. I had rebased it to main by mistake

PR #609, opened 2023-06-07
fix: variable shadowing with `transferOwnership(_pendingOwner)` and `_pendingOwner` state variable

CJ42:

## 🐛 Bug

Slither reports the following variable shadowing in `LSP0ERC725AccountCore.sol`.

image

Despite that the `_pendingOwner` variable is defined as `private` inside `LSP14`, rename the variable so that the parameter name of the function `transferOwnership` does not clash the name with the state variable `_pendingOwner`.

### PR Checklist

- [x] Wrote Tests
- [x] Wrote Documentation
- [x] Ran `npm run linter` (solhint)
- [x] Ran `npm run build`
- [x] Ran `npm run test`

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -480,17 +480,20 @@ abstract contract LSP0ERC725AccountCore is
     /**
      * @notice Achieves the goal of LSP14Ownable2Step by implementing a 2-step ownership transfer process.
      *
-     * @dev Sets the pending owner address as an address that should call {acceptOwnership} in order to complete the ownership transfer of the account.
-     * Notifies the pending owner via LSP1Standard by calling {universalReceiver()} on the pending owner if it's an address that supports LSP1.
+     * @dev Sets the address of the `pendingNewOwner` as a pending owner that should call {`acceptOwnership()`} in order to complete
+     * the ownership transfer to become the new {`owner()`} of the account.
      *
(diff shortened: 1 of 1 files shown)


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -505,7 +505,7 @@ abstract contract LSP0ERC725AccountCore is
      *
      * - pending owner cannot accept ownership in the same tx via the LSP1 hook.
      */
-    function transferOwnership(address _pendingOwner)
+    function transferOwnership(address newOwner)
```
YamenMerhi, line 508: **Minor Point:**
I wouldn't say `newOwner` is exactly accurate because he's not the new owner even after the completion of the tx.
I would keep `pendingOwner` but change the underscore location so it doesn't shadow the variable.

YamenMerhi (approved):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
CJ42: How about `newPendingOwner`? or `pendingNewOwner`?

@YamenMerhi @skimaharvey @b00ste wdyt?

CJ42: Actually, we cannot use `pendingOwner` as a param because it shadows the name of the getter function `pendingOwner()`.

So I will go with `pendingNewOwner`.

b00ste (approved):

PR #611, opened 2023-06-08
chore: add NatSpec guidelines to `CONTRIBUTING.md`

b00ste:
## 📄 Documentation

This PR adds guidelines to how to properly write NatSpec for each `function`, `error` and `event` in the Solidity Contracts.

### PR Checklist

- [ ] Wrote Tests
- [ ] Wrote Documentation
- [ ] Ran `npm run linter` (solhint)
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`


YamenMerhi (approved):

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
 <!-- ![alt text](https://docs.github.com/assets/images/help/pull_requests/select-squash-and-merge-from-drop-down-menu.png) -->
 <img src="https://docs.github.com/assets/images/help/pull_requests/select-squash-and-merge-from-drop-down-menu.png" alt="drawing" style="width:600px;"/>
 
+## Solidity NatSpec
```
CJ42, line 55: I would suggest a more explicit subtitle

```suggestion
## Solidity Code Comments
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+When changing or adding NatSpec comments to the any function, error or event in any contract make sure to adhere to the following guidelines:
```
CJ42, line 57:
```suggestion
When changing or adding NatSpec comments to any `function`, `error` or `event` in any contract make sure to adhere to the following guidelines:
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+1. `@dev` and `@notice` tags can both contain text description and two types of lists, bullet points or numbered lists. Make sure that those tags always start with text description first, not with lists.
```
CJ42, line 59:
```suggestion
1. `@dev` and `@notice` tags can both contain text descriptions and two types of lists: bullet points or numbered lists. Make sure that those tags always start with text descriptions first, not with lists.
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+ *
+ * 1. First bullet point.
+ *   - First orderd element.
+ *   - Second orderd element.
+ *
+ * 2. Second bullet point.
+ *   - First orderd element.
+ *   - Second orderd element.
+ */
+```
```
CJ42, line 99: I think it would also be quite good to show a concrete example of what **not to do**. Like in the Solidity docs.

https://docs.soliditylang.org/en/v0.8.19/style-guide.html#blank-lines

image

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+2. `@param` tag is mandatory if the function, error or event has any parameters.
+
+3. `@return` tag is mandatory if the fucntion has any return value.
```
CJ42, line 103:
```suggestion
3. `@return` tag is mandatory if the function has any return value.
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+4. Make sure to use custom tags for different kinds of information:
```
CJ42, line 105:
```suggestion
4. Make sure to use one of the custom tags below to document any additional informations of different kinds:
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+- Use `@custom:requirements` for all the requirements to use a function.
+- Use `@custom:events` for all the emited events during the function execution.
+- Use `@custom:warning` for any warnings someone must be aware of when usinf a function.
```
CJ42, line 109: Since warning tags can be generic and added to not only functions, I would keep it generic.

```suggestion
- Use `@custom:warning` for any warnings someone must be aware of.
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+- Use `@custom:danger` for any dangers one needs to be aware when using the function.
+- Use `@custom:hint` for any tips you might have for someone that wants to use the function.
+- Use `@custom:info` for any extra informartion.
```
CJ42, line 112:
```suggestion
- Use `@custom:info` for any extra information.
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+The custom tags should span across a single line or multiple lines using `-`.
```
CJ42, line 114: Suggested change.

```suggestion
If there is only a single comment for the `@custom` tag, it can be written on the same line as the `@custom` tag. 
If there is multiple comments for the `@custom` tag, it should span across multiple lines using `-` as bullet points.
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+
+```solidity
+/**
+ * @custom:events
+ * - {ValueReceived} event when someone tranfers native tokens to the contract.
+ * - {Executes} event when the function is executed without any issues.
+ */
+```
+
+5. If you want add an internal link to an function, error or event from the current contract, use `{}`.
```
CJ42, line 134:
```suggestion
5. If you want to add an internal link to a `function`, `error` or `event` defined the current contract, use `{}`.
```

CONTRIBUTING.md
```diff
@@ -52,6 +52,91 @@ When merging a branch to `develop` PRs should be squashed into one conventional
...
+   E.g.: `{pendingOwner}`, `{pendingOwner()}` etc.
+
+6. If you want to add an external link to anything, use `[]`.
```
CJ42, line 137: I think it is not clear here about external links. We should probably add an extra note that the Markdown file where the documentation will be included can include footer with the external links for these references. (But explain that these footer links **MUST NOT** be added to the generated docs!)

CJ42 (commented):
lgtm! 🚀 

I am also thinking if we should mention for Natspec comments that we favour Natspec comments written as `/** */` and not as triple slashes `///`.

lmk wyt but not a big deal if it's not mentioned.

CJ42 (approved):

PR #644, opened 2023-07-31
Add virtual

CallumGrindle:

 ♻️ Refactor
Add missing virtual keywords

### PR Checklist

- [ ] Wrote Tests
- [x] Ran `npm run build`
- [x] Ran `npm run test`

### DIFF
--- a/contracts/LSP20CallVerification/LSP20CallVerification.sol
+++ b/contracts/LSP20CallVerification/LSP20CallVerification.sol
@@ -70,7 +70,7 @@ abstract contract LSP20CallVerification {
         bool postCall,
         bool success,
         bytes memory returnedData
-    ) internal pure {
+    ) internal pure virtual {
         if (!success) _revert(postCall, returnedData);
 
         // check if the returned data contains at least 32 bytes, potentially an abi encoded bytes4 value
@@ -81,7 +81,10 @@ abstract contract LSP20CallVerification {
         ) revert LSP20InvalidMagicValue(postCall, returnedData);
     }
 
-    function _revert(bool postCall, bytes memory returnedData) internal pure {
+    function _revert(
+        bool postCall,
+        bytes memory returnedData
+    ) internal pure virtual {
         // Look for revert reason and bubble it up if present
         if (returnedData.length > 0) {
             // The easiest way to bubble the revert reason is using memory via assembly
--- a/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
+++ b/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
@@ -364,7 +364,7 @@ abstract contract LSP6ExecuteModule {
     function _isAllowedAddress(
         bytes memory allowedCall,
         address to
-    ) internal pure returns (bool) {
+    ) internal pure virtual returns (bool) {
         // <offset> = 4 bytes x 8 bits = 32 bits
         //
         // <offset>v----------------address---------------v
--- a/contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
+++ b/contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
@@ -438,7 +438,7 @@ abstract contract LSP6SetDataModule {
         bytes32 dataKey,
         bytes memory dataValue,
         bool hasBothAddControllerAndEditPermissions
-    ) internal view returns (bytes32) {
+    ) internal view virtual returns (bytes32) {
         if (!LSP6Utils.isCompactBytesArrayOfAllowedERC725YDataKeys(dataValue)) {
             revert InvalidEncodedAllowedERC725YDataKeys(
                 dataValue,
--- a/contracts/LSP7DigitalAsset/extensions/LSP7Burnable.sol
+++ b/contracts/LSP7DigitalAsset/extensions/LSP7Burnable.sol
@@ -14,7 +14,11 @@ abstract contract LSP7Burnable is LSP7DigitalAsset {
      *
      * See internal _burn function for more details
      */
-    function burn(address from, uint256 amount, bytes memory data) public {
+    function burn(
+        address from,
+        uint256 amount,
+        bytes memory data
+    ) public virtual {
         _burn(from, amount, data);
     }
 }
--- a/contracts/LSP9Vault/LSP9VaultCore.sol
+++ b/contracts/LSP9Vault/LSP9VaultCore.sol
@@ -233,7 +233,7 @@ contract LSP9VaultCore is
      */
     function batchCalls(
         bytes[] calldata data
-    ) public returns (bytes[] memory results) {
+    ) public virtual returns (bytes[] memory results) {
         results = new bytes[](data.length);
         for (uint256 i; i < data.length; ) {
             (bool success, bytes memory result) = address(this).delegatecall(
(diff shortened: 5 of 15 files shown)


CJ42:
@CallumGrindle can you add yourself in the [`.allcontributorsrc`](https://github.com/lukso-network/lsp-smart-contracts/blob/develop/.all-contributorsrc) :)

See here for how to: https://github.com/lukso-network/lsp-smart-contracts/blob/develop/CONTRIBUTING.md#adding-contributors

contracts/LSP20CallVerification/LSP20CallVerification.sol
```diff
@@ -81,7 +81,10 @@ abstract contract LSP20CallVerification {
         ) revert LSP20InvalidMagicValue(postCall, returnedData);
     }
 
-    function _revert(bool postCall, bytes memory returnedData) internal pure {
+    function _revert(
+        bool postCall,
+        bytes memory returnedData
+    ) internal pure virtual {
```
CJ42, line 87: I think this function does not need to be marked as `virtual` and could even be changed from `internal` to `private`, as it is only used inside `LSP20CallVerification.sol`.

(below the other contracts are just a Mock contract + a Foundry test)

image

Or we could keep it as `internal` so that it can be used in the inheritance, but rename it to something more specific related to LSP20. The name is a bit too generic imo.

```
function _revertWithLSP20DefaultError(...)
```

Or something like that @YamenMerhi @skimaharvey

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
```diff
@@ -364,7 +364,7 @@ abstract contract LSP6ExecuteModule {
     function _isAllowedAddress(
         bytes memory allowedCall,
         address to
-    ) internal pure returns (bool) {
+    ) internal pure virtual returns (bool) {
```
CJ42, line 367: Hmm is this a good idea?

I wouldn't make these functions `virtual` as they are critical to verifying permissions. Also because other similar related functions are not marked as `virtual`.

image

If someone override this function like this, the verification of permissions for `AllowedCalls` would have a flow in a Key Manager implementation!

```solidity
    function _isAllowedAddress(
        bytes memory allowedCall,
        address to
    ) internal pure returns (bool) {
        return true;
    }
```

contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
```diff
@@ -438,7 +438,7 @@ abstract contract LSP6SetDataModule {
         bytes32 dataKey,
         bytes memory dataValue,
         bool hasBothAddControllerAndEditPermissions
-    ) internal view returns (bytes32) {
+    ) internal view virtual returns (bytes32) {
```
CJ42, line 441: Also questionable here, as is it the right thing to do here.

But maybe ok out of consistency because the other functions are marked as `virtual`. 👍 

image

contracts/LSP7DigitalAsset/extensions/LSP7Burnable.sol
```diff
@@ -14,7 +14,11 @@ abstract contract LSP7Burnable is LSP7DigitalAsset {
      *
      * See internal _burn function for more details
      */
-    function burn(address from, uint256 amount, bytes memory data) public {
+    function burn(
+        address from,
+        uint256 amount,
+        bytes memory data
+    ) public virtual {
```
CJ42, line 21: Yes make sense 👌

contracts/LSP9Vault/LSP9VaultCore.sol
```diff
@@ -366,7 +366,12 @@ contract LSP9VaultCore is
     /**
      * @dev Modifier restricting the call to the owner of the contract and the UniversalReceiverDelegate
      */
-    function _validateAndIdentifyCaller() internal view returns (bool isURD) {
+    function _validateAndIdentifyCaller()
```
CJ42, line 369: I would not mark this as `virtual`, as currently we want to enforce this behaviour in the Vault that only the owner or the LSP1 Delegate can reenter.

So this should not be allowed to be overriden.

@YamenMerhi @skimaharvey @b00ste

CJ42 (changes requested):

contracts/LSP9Vault/LSP9VaultCore.sol
CallumGrindle: the thinking was that we don't know how people will use these. Any of these functions could be overwritten to add extra functionality. Imo if someone extends a function to write a junk/broken implementation thats on them and we should let them

CJ42:
@CallumGrindle can you also mark the `target()` function as `virtual` please?

@skimaharvey had issues when overriding

image

https://github.com/lukso-network/lsp-smart-contracts/blob/b9e6c8967eb621feff343cc071aa044f0c644faf/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol#L85-L87

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
b00ste: Maybe someone wants a different way of checking for AllowedCalls 🤔

contracts/LSP20CallVerification/LSP20CallVerification.sol
YamenMerhi: Okay with renaming 👍

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
CJ42: Yes but then why not marking the other functions as virtual as well:

- [`_isAllowedCallType(allowedCall, requiredCallTypes)`](https://github.com/lukso-network/lsp-smart-contracts/blob/8f0cfb2c573c44702d3155375b2d935b043416b3/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol#L416-L419)
- `_isAllowedAddress(allowedCall, to)`
- `_isAllowedStandard(allowedCall, to)`
- `_isAllowedFunction(allowedCall, selector)`

We already have these functions marked as `virtual`:
- [`_verifyCanDeployContract`](https://github.com/lukso-network/lsp-smart-contracts/blob/8f0cfb2c573c44702d3155375b2d935b043416b3/contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol#L151C14-L155)
- `_verifyCanStaticCall`
- `_verifyCanCall`
- `_verifyAllowedCall`

Wdyt @b00ste @YamenMerhi @CallumGrindle @skimaharvey

contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
skimaharvey: i think we should all mark then virtual no?

skimaharvey (approved):

contracts/LSP6KeyManager/LSP6Modules/LSP6ExecuteModule.sol
CallumGrindle: Added to these now

b00ste (approved):

CJ42 (approved):

PR #645, opened 2023-08-02
refactor: (C4#123) add transfer ownership lock

YamenMerhi:

- Add a `_inTransferOwnership` state variable, which ensures that `acceptOwnership()` cannot be called while `transferOwnership()` is in execution, similar to a reentrancy guard.

## 🐛 Bug

The currentOwner == owner() check ensures that pendingNewOwner did not call acceptOwnership() in the universalReceiver() callback. However, a malicious contract can bypass this check by doing the following in its universalReceiver() function:

- Call acceptOwnership() to gain ownership of the LSP0 account.
- Do whatever he wants, such as transferring the account's entire LYX balance to himself.
- Call execute() to perform a delegate call that does either of the following:
      - Delegate call into a contract that self-destructs, which will destroy the account permanently.
Otherwise, use delegate call to overwrite _owner to the previous owner.

This defeats the entire purpose of a two-step ownership transfer, which should ensure that the LSP0 account cannot be lost in a single call if the owner accidentally calls transferOwnership() with the wrong address.

### PR Checklist

- [ ] Wrote Tests
- [x] Ran `npm run build`
- [x] Ran `npm run test`


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -624,6 +622,12 @@ abstract contract LSP0ERC725AccountCore is
     function acceptOwnership() public virtual override {
         address previousOwner = owner();
 
+        if (_inTransferOwnership) {
```
b00ste, line 625: What about creating a modifier in LSP14 and using it both in LSP0 and LSP9

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
```diff
@@ -53,6 +53,11 @@ abstract contract LSP14Ownable2Step is ILSP14Ownable2Step, OwnableUnset {
      */
     address private _pendingOwner;
 
+    /**
+     * @dev The boolean that indicates whether the contract is in an active ownership transfer phase
+     */
+    bool internal _inTransferOwnership;
```
b00ste, line 59: Maybe we should take openzeppelins approach and instead of using `bool`, we should use `uint256` and changing from `uint256 NOT_IN_OWNERSHIP_TRANSFER = 1` to `uint256 IN_OWNERSHIP_TRANSFER = 2`

b00ste (approved):
Can you please add tests for this behaviour?

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
YamenMerhi: This would add the need to have a constructor to set the state variable to its value, which we should avoid now because of `Core` architecture

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: in LSP0 we cannot use the modifier as it should be located inside a specific part of the code as we have normal execution by the owner and lsp20 path, but can work with normal lsp14, but is it worth it just to use it for lsp14 and lsp9 once ?

b00ste: I think it improves readability, and might be useful for people that want to implement LSP14 in their contracts. It might make it easier to avoid a potential bug

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -561,6 +561,9 @@ abstract contract LSP0ERC725AccountCore is
 
         // If the caller is the owner perform transferOwnership directly
         if (msg.sender == currentOwner) {
+            // set the transfer ownership lock
+            _inTransferOwnership = true;
```
CJ42, line 565: I am wondering if it is cheaper to use a `uint8` and set two potential value to `1` or `2` for optimisation (the same way as a reentrancy guard.
What is your opinion on that @skimaharvey @b00ste @YamenMerhi?

CJ42: See here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/54a235f8959ecffb1916cf5693ec9bbd695cbf71/contracts/security/ReentrancyGuard.sol#L23-L33

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/54a235f8959ecffb1916cf5693ec9bbd695cbf71/contracts/security/ReentrancyGuard.sol#L72-L73

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
CJ42: Agree @b00ste idea on adding a `modifier`. We can make the modifier `virtual`  too and `override` it in LSP0 to add the LSP20 logic inside it (based on if the caller is the `owner()` or not.

See this code snippet example for overriding the `modifier` syntax:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

contract Playground {
   
   modifier NotInTransferOwnership virtual {
       require(msg.sender != address(0));
       _;
   }

}

contract Example is Playground {

    modifier NotInTransferOwnership virtual override {
        // this is just as an example
        require(msg.sender != address(0));
        require(msg.sender != address(this));
        _;
    }

}

```

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
CJ42: Ah ok, I mentioned this above. But forgot about the `Core` architecture.
You can ignore my comment above :)

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
CJ42: You can ignore this comment because of this :)
https://github.com/lukso-network/lsp-smart-contracts/pull/645/files#r1281746603

contracts/LSP14Ownable2Step/LSP14Ownable2Step.sol
```diff
@@ -87,6 +94,12 @@ abstract contract LSP14Ownable2Step is ILSP14Ownable2Step, OwnableUnset {
     function acceptOwnership() public virtual {
         address previousOwner = owner();
 
+        if (_inTransferOwnership) {
+            revert(
+                "LSP14: newOwner MUST accept ownership in a separate transaction"
```
CJ42, line 99: I would suggest to take the opportunity of this PR to move from revert reason string to custom error.
This error string is quite long and a custom error could reduce the deployment cost of LSP14 and the UP in turns.

Suggesting:

```solidity
error LSP14MustAcceptOwnershipInSeparateTransaction
```

Or something like that.

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
```diff
@@ -0,0 +1,86 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import "forge-std/Test.sol";
+import "../../../contracts/LSP0ERC725Account/LSP0ERC725Account.sol";
+
+contract Implementation {
```
CJ42, line 7: Same here, why not having the implementation contract inheriting from `LSP0ERC725Account`? So that we don't have to re-create the storage layout like this?

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
```diff
@@ -0,0 +1,86 @@
...
+
+        // Call acceptOwnership() to become the owner
+        account.acceptOwnership();
+
+        // and then steal all ether in contract
+        // and set back the owner to the original value
+
+        // Transfer all LYX balance to this contract
+        account.execute(
+            0, // OPERATION_0_CALL
```
CJ42, line 42: Maybe we can import the `OPERATION_0_CALL` constant from `@erc725/smart-contracts` to improve readability here?

```solidity
import {OPERATION_0_CALL} from "@erc725/smart-contracts/contracts/constants.sol";

```

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
```diff
@@ -0,0 +1,86 @@
...
+            address(this),
+            10 ether,
+            ""
+        );
+
+        // Overwrite _owner with the previous owner using delegatecall
+        Implementation implementation = new Implementation();
+        account.execute(
+            4, // OPERATION_4_DELEGATECALL
```
CJ42, line 51: same suggestion as above, suggesting to import the constants from `@erc725/smart-contracts` to improve readability.

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
```diff
@@ -0,0 +1,86 @@
...
+        // Transfer all LYX balance to this contract
+        account.execute(
+            0, // OPERATION_0_CALL
+            address(this),
+            10 ether,
+            ""
+        );
+
+        // Overwrite _owner with the previous owner using delegatecall
+        Implementation implementation = new Implementation();
```
CJ42, line 49: I am confused, this is creating a new contract here?

Shouldn't you pass the `account` address instead?

image

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
CJ42, line 1: Shouldn't this test suite be inside the LSP14 folder? Since it is related to LSP14 standard?

Maybe we could make the test to use a basic `ERC725` contract, not necessarely a `LSP0ERC725Account`. So that we can test the behaviour of delegatecall without having the tests to tied to LSP0 contracts.

I think importing LSP0 here will also create an issue with the monorepo and a cyclic dependency, as:
LSP0 --> imports LSP14 and use LSP14 package in the monorepo.
(if we move this test suite in LSP14): LSP14 --> import LSP0 below line 5, so will require LSP14 package, which will create a dependency.

@kalote what do you think?

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
```diff
@@ -0,0 +1,86 @@
...
+        account.acceptOwnership();
+
+        // and then steal all ether in contract
+        // and set back the owner to the original value
+
+        // Transfer all LYX balance to this contract
+        account.execute(
+            0, // OPERATION_0_CALL
+            address(this),
+            10 ether,
```
CJ42, line 44: Suggest this change to match with the comments above.

```suggestion
            address(account).balance,
```

CJ42 (commented):
added review comments

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: Not in favor of a modifier as it is not that shared in the code and can be easily forgotten, but I'll implement it

tests/foundry/LSP0ERC725Account/TwoStepOwnership.sol
kalote: If this test is related to LSP14, then yes, it should probably be in LSP14 folder. Although, I don't see any LSP14 reference in the below contract.
For the monorepo, I haven't implemented the test suite enough to tell you exactly if this is going to be an issue. What I can tell though, is that the contract below is definitely not a unit test. Probably more an integration test. Which will be inside its own package, so I guess it should be ok.

CJ42 (approved):

PR #648, opened 2023-08-02
refactor: (C4#126) disallow setting keyManager for lsp20 extension

YamenMerhi:

- Disallow setting keyManager address as extension for lsp20 function selectors
 

## ♻️ Refactor

After setting the address of the KeyManager as extension for `lsp20VerifyCallResult` function.

As lsp20VerifyCallResult() is called through the LSP0 account, msg.sender == _target will be true, thereby calling _nonReentrantAfter() to reset _reentrancyStatus.

Therefore, after lsp20VerifyCallResult() is added as an extension, anyone can then call this before performing a reentrant call to LSP6KeyManager to bypass reentrancy checks.

The opposite case can happen when setting the KeyManager address as extension for `lsp20VerifyCall` function, as in this case the reentrancy status will be set to true, and hence each call will require reentrancy permission.

### PR Checklist

- [x] Wrote Tests
- [x] Ran `npm run build`
- [x] Ran `npm run test`


b00ste (approved):

tests/LSP6KeyManager/Admin/PermissionChangeAddExtensions.test.ts
```diff
@@ -193,6 +193,74 @@ export const shouldBehaveLikePermissionChangeOrAddExtensions = (
...
+            context.keyManager,
+            'KeyManagerCannotBeSetAsExtensionForLSP20Functions',
+          );
+        });
+
+        it('should revert when adding the KeyManager address as extensions for lsp20VerifyCallResult function', async () => {
+          const payloadParam = {
+            dataKey:
+              ERC725YDataKeys.LSP17.LSP17ExtensionPrefix +
+              'd3fc45d3' + // function selector of `lsp20VerifyCallResult(bytes32,bytes)`
```
CJ42, line 223: Maybe you can remove the comment, store this variable in a `const` and use Typechain?

```ts
const lsp20VerifyCallResultSelector = context.keyManager.interface.getSigHash("lsp20VerifyCallResult")
```

Same for the function selector of `lsp20VerifyCall(address,uint256,bytes)`

CJ42 (approved):

PR #649, opened 2023-08-02
feat: create LSP23MultiChainDeployment

skimaharvey:


b00ste:
I will leave this here, to be discussed later.

Split the `DeployedContracts(...)` & `DeployedERC1167Proxies(...)` event in 5 events:
- `DeployedContract(...)`
- `DeployedOwnerContract(...)`
- `DeployedERC1167Proxie(...)`
- `DeployedOwnerERC1167Proxie(...)`
- `PostDeployment(...)`

This allows for an easier use of events, eg. `DeployedContract` is used when the initial contract was successfully deployed but also before the owner contract was deployed.
The execution timeline is correct. Also allows for easier to read events, instead of a single event with 4 simple parameters and 2 structs, into events containing 1 simple param, 1 struct and the last event containing 2 simple params.

CJ42 (approved):

PR #659, opened 2023-08-08
fix: (C4 #97) add check for `length == 0` to prevent `mask` to allow any data key

CJ42:

### Overview

There is an insufficient length check when validating the allowed data keys that can be set by a caller. In particular, if a decoded length in the compact bytes array happens to be 0, the caller will be validated against any data key.

Although a decoded length is not possible when setting allowed data keys via the `LSP6SetDataModule` contract, there is no guarantee that data values before ownership transfer to an LSP6 contract are valid. This allows a scamming opportunity as malicious actors who set-up an ERC725 contract for another can create a backdoor that cannot be easily seen as the allowed data keys will simply have an extra `0x0000`.

When verifying allowed data keys that a caller can set, a length check is done on the first 2 bytes of a compact bytes array to ensure that the length does not exceed a data key's length.

```solidity
            if (length > 32)
                revert InvalidEncodedAllowedERC725YDataKeys(
                    allowedERC725YDataKeysCompacted,
                    "couldn't DECODE from storage"
                );
```

This length value is used to decide the bit mask when validating data keys.

```solidity
            mask =
                bytes32(
                    0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
                ) <<
                (8 * (32 - length));
```

In the case that `length == 0`, then `mask == bytes32(0)`.

Then the validation check for all data keys will pass as the validation will be reduced to whether or not `allowedKey & bytes32(0) == inputDataKey & bytes32(0)`, which is always true.

```solidity
                allowedKey := and(memoryAt, mask)
            }

            if (allowedKey == (inputDataKey & mask)) return;
```

This allows the caller to set data values for all data keys except those used for LSP1, LSP6, and LSP17 due to the initial checks in `LSP6SetDataModule._getPermissionRequiredToSetDataKey(...)`. In particular, they would be able to obstruct data values for the following established keys:

LSP3 keys detailing information about a universal profile
LSP5 keys detailing information about received assets
LSP10 keys detailing information about vault ownership

### Implemented Fixes

-  add a check in `LSP6SetDataModule` to revert if in the  extracted from the `AllowedERC725YDataKeys` read from the linked account storage, there is an entry in the CompactByteArray that is `0x0000`, meaning the `length` for the entry (= the allowed data key) is `0`.

- Add Foundry tests to test that the scenario described above reverts

### PR Checklist

- [x] Wrote Tests
- [x] Ran `npm run build`
- [x] Ran `npm run test`

### DIFF
--- a/tests/foundry/LSP6KeyManager/LSP6SetDataTest.t.sol
+++ b/tests/foundry/LSP6KeyManager/LSP6SetDataTest.t.sol
@@ -1,0 +1,230 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.4;
+
+import "forge-std/Test.sol";
+import "../../../contracts/UniversalProfile.sol";
+import "../../../contracts/LSP1UniversalReceiver/LSP1UniversalReceiverDelegateUP/LSP1UniversalReceiverDelegateUP.sol";
+import "../../../contracts/LSP6KeyManager/LSP6KeyManager.sol";
+import "../../../contracts/LSP14Ownable2Step/ILSP14Ownable2Step.sol";
+import "@erc725/smart-contracts/contracts/interfaces/IERC725Y.sol";
+
+import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
+import {
+    LSP2Utils
+} from "../../../contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol";
+import {LSP6Utils} from "../../../contracts/LSP6KeyManager/LSP6Utils.sol";
+
+import "../../../contracts/LSP1UniversalReceiver/LSP1Constants.sol";
+import "../../../contracts/LSP6KeyManager/LSP6Constants.sol";
+import "../../../contracts/LSP17ContractExtension/LSP17Constants.sol";
+
+contract LSP6SetDataTest is Test {
+    using BytesLib for bytes;
+
+    UniversalProfile universalProfile;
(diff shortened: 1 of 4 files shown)


Hugoo:
Task linked: [DEV-7697 (C4 #97) - Insufficient Length Check When Verifying Allowed Data Keys](https://app.clickup.com/t/2645698/DEV-7697)

tests/foundry/LSP6KeyManager/LSP6SetDataTest.t.sol
```diff
@@ -0,0 +1,230 @@
...
+        assert(universalProfile.owner() == address(keyManager));
+
+        // Verify malicious can set data for most data keys
+        bytes memory functionArgs = abi.encode(dataKey, dataValue);
+        bytes memory callData = abi.encodeWithSelector(
+            IERC725Y.setData.selector,
+            functionArgs
+        );
+
+        // CHECK it reverts when calling via the Key Manager
```
YamenMerhi, line 221: How you are testing for revert ? I don't see any `vm.expectRevert`

skimaharvey: because it using `testFail_` pattern. it is expecting a revert

YamenMerhi: Ah nice, didn't see it

skimaharvey (approved):
LGTM. Interesting though that with `vm.expectRevert` it is not working

PR #670, opened 2023-08-09
refactor: mark relay call signature as used

b00ste:

## ♻️ Refactor

After investigating the way signatures are stored in [`Safe.sol`](https://github.com/safe-global/safe-contracts/blob/d2dfe316ff24162159ee48b05dd6827cf95f06db/contracts/Safe.sol#L67-L68), it turns out that they have a [`signMessage(...)`](https://github.com/safe-global/safe-contracts/blob/d2dfe316ff24162159ee48b05dd6827cf95f06db/contracts/libraries/SignMessageLib.sol#L22) function which marks a signature as approved signature. That further is checked in [`isValidSignature(...)`](https://github.com/safe-global/safe-contracts/blob/d2dfe316ff24162159ee48b05dd6827cf95f06db/contracts/handler/CompatibilityFallbackHandler.sol#L34) which requires that a signature to be approved.

With this said, they don't quite have the same purpose as in our case. We need to store signatures in order to not allow double use while they store signatures that are approved. In our case we cannot use [`signMessage(...)`](https://github.com/safe-global/safe-contracts/blob/d2dfe316ff24162159ee48b05dd6827cf95f06db/contracts/libraries/SignMessageLib.sol#L22) as it will block gas-less execution and will require for someone to have gas in order to approve a signature. We need an approach that does not break the ux behind relay calls.

I took the liberty to implement it in the following way:
- Before executing a relay call, we check that the signature is marked as unused (0) in the `signedMessages` mapping.
- After executing a relay call successfully we mark the signature as used (1) in the `signedMessages` mapping.

That would 100% not allow any signature to be used twice.

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -40,7 +40,8 @@ import {
     InvalidERC725Function,
     CannotSendValueToSetData,
     RelayCallBeforeStartTime,
-    RelayCallExpired
+    RelayCallExpired,
+    CannotReplayTheRelayCall
 } from "./LSP6Errors.sol";
 
 import {
@@ -83,6 +84,9 @@ abstract contract LSP6KeyManagerCore is
     // Variables, methods and modifier used for ReentrancyGuard are taken from the link below and modified accordingly.
     // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.8/contracts/security/ReentrancyGuard.sol
     bool internal _reentrancyStatus;
+
+    // Mapping to keep track of all message hashes that have been approved by ALL REQUIRED owners
+    mapping(bytes32 => bool) public signedMessages;
 
     mapping(address => mapping(uint256 => uint256)) internal _nonceStore;
 
@@ -431,11 +435,18 @@ abstract contract LSP6KeyManagerCore is
         _verifyPermissions(signer, msgValue, payload);
         emit PermissionsVerified(signer, msgValue, bytes4(payload));
 
+        bytes32 encodedMessageHash = keccak256(encodedMessage);
+        if (signedMessages[encodedMessageHash]) {
+            revert CannotReplayTheRelayCall(encodedMessageHash);
+        }
+
         bytes memory result = _executePayload(msgValue, payload);
 
         if (!isReentrantCall && !isSetData) {
             _nonReentrantAfter();
         }
+
+        signedMessages[encodedMessageHash] = true;
 
         return result;
     }
--- a/contracts/LSP6KeyManager/LSP6Errors.sol
+++ b/contracts/LSP6KeyManager/LSP6Errors.sol
@@ -245,3 +245,10 @@ error RelayCallExpired();
  * @dev reverts when the address of the Key Manager is being set as extensions for lsp20 functions
  */
 error KeyManagerCannotBeSetAsExtensionForLSP20Functions();
+
+/**
+ * @dev reverts when trying to use the same arguments with a different signature.
+ * @notice Cannot use the same arguments twice! Encoded message: `encodedMessage`
+ * @param encodedMessageHash The hash of the [EIP-191] signed data.
+ */
+error CannotReplayTheRelayCall(bytes32 encodedMessageHash);


Hugoo:
Task linked: [DEV-7688 Add signed messages inside the mapping. (after use)](https://app.clickup.com/t/2645698/DEV-7688)

tests/Reentrancy/LSP6/LSP6Reentrancy.test.ts
```diff
@@ -182,8 +182,8 @@ export const shouldBehaveLikeLSP6ReentrancyScenarios = (
               .connect(relayer)
               .executeRelayCall(signature, nonce, validityTimestamps, executeRelayCallPayload),
           )
-            .to.be.revertedWithCustomError(context.keyManager, 'InvalidRelayNonce')
-            .withArgs(signer.address, nonce, signature);
+            .to.be.revertedWithCustomError(context.keyManager, 'CannotUseSignatureTwice')
```
skimaharvey, line 185: probably should rename line 140?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -347,6 +351,9 @@ abstract contract LSP6KeyManagerCore is
         uint256 msgValue,
         bytes calldata payload
     ) internal virtual returns (bytes memory) {
+        if (signedMessages[keccak256(signature)] == 1) {
```
skimaharvey, line 354: why use one over bool? is it really a gas optimisation?
Also:
- probably could store the `keccak256(signature)` into a stack variable as we do it again below
- initially i thought we were going to store the hash of the `keccak256(encodedMessage)` but I guess signature makes sense as well

YamenMerhi:
Few things I had in mind:
- Need to decide what is best for the visibility of the mapping (should it be public or internal)
- Why not store the signature itself instead of a hash, lowering the possibilities of a hash collision.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
b00ste: I took the safe implementation, my guess would be that it is indeed cheaper. 
For the first point, yes but is we do as Yamen suggested then it makes no sense to store the signature plain bytes to the stack.
For the second point, to me it makes more sense to check the signature as the signature is the thing that you don't want to re-play.

Wdyt @skimaharvey @CJ42? Should we keep the current mapping? Or change it to one from the below examples?
- `mapping (bytes32 => bool) signedMessages;`
- `mapping (bytes => uint256) signedMessages;`
- `mapping (bytes => bool) signedMessages;`

skimaharvey: would not store `bytes` but `bytes32` . I think first one is good `mapping (bytes32 => bool) signedMessages;`

skimaharvey: imo lets investigate signature malleability more before merging this PR https://github.com/kadenzipfel/smart-contract-vulnerabilities/blob/master/vulnerabilities/signature-malleability.md

skimaharvey: thats why we should may be actually store the hash of the `encodedMessage` message so we cant find the equivalent

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -397,12 +401,18 @@ abstract contract LSP6KeyManagerCore is
         _verifyPermissions(signer, msgValue, payload);
         emit VerifiedCall(signer, msgValue, bytes4(payload));
 
+        if (signedMessages[keccak256(encodedMessage)] == 1) {
```
skimaharvey, line 404: so you dont think its worth it saving the hashing in a stack variable?

skimaharvey: also not sure about using 1 vs bool

skimaharvey (approved):

skimaharvey:
LGTM.
- Just not sure we should not save the hash the `econdedMessage` in a stack variable. 
- Using bool instead of `1`

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42, line 404: Depending if we implement this or not, I recommended we cache the keccak256 hash in a local variable, instead of re-hashing again further down below.

Hashing is costly in gas, the more bytes there are to hash

CJ42 (changes requested):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
b00ste: it is, just got out of my mind

CJ42:
@b00ste @skimaharvey 

I suggest we put this PR as draft.
Also could you check how much gas difference there is for the deployment cost by comparing the gas benchmark of the Github actions?

(Looks like the gas diff for Foundry does not show this)

PR #674, opened 2023-08-11
fix: (C4 #139) refactor `_fallbackLSP17Extendable` function to enable to run code after it is called + prevent potential solc bug "storage write removal"

CJ42:

## 🐛 Bug

Currently, the function in LSP17 `_fallbackLSP17Extendable` does not allow to run any code logic after this function is called, because it uses the low level opcode `return()` in assembly. 
This restrict some potential customization through inheritance, when one want to run some custom code logic after this function is being called.

This PR refactor this function:
- in LSP17 to take a parameter.
- in LSP0 + LSP9 to move from the following `fallback` syntax:

```solidity
fallback() external payable {
    // ...
}
```

To the new following `fallback` syntax (available in all `0.8.x` versions of Solidity).

```solidity
fallback(bytes calldata data) external payable returns (bytes memory) {
   // ...
}
```

By removing assembly, , the function can be overriden to be called via `super._fallbackLSP17Extendable(...)` and run any code after.

> **Note:** this refactoring also helps to prevent potential storage removal bug from the Solidity compiler, if projects inherit the LSP17 contract and compile their contract with a version from 0.8.13 to 0.8.16 (where the bug is present). See Solidity blog for more infos: https://soliditylang.org/blog/2022/09/08/storage-write-removal-before-conditional-termination/

## 🧪 Tests

Created new test suite `LSP17Extendable.test.ts` to test the basic functionalities of LSP17 and this new refactoring.
Tests include checking that storage write works as expected both **before** and **after** the `_fallbackLSP17Extension()` function is called.

Also added tests to ensure different type of revert reason strings are bubble up.

### PR Checklist

- [x] Wrote Tests
- [x] Wrote & Generated Documentation (readme/natspec/dodoc)
- [x] Ran `npm run build`
- [x] Ran `npm run test`

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -64,6 +64,7 @@ jobs:
             "lsp9init",
             "lsp11",
             "lsp11init",
+            "lsp17",
             "lsp20",
             "lsp20init",
             "lsp23",
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -805,23 +805,18 @@ abstract contract LSP0ERC725AccountCore is
         if (extension == address(0))
             revert NoExtensionFoundForFunctionSelector(msg.sig);
 
-        bytes memory calldataWithCallerInfos = abi.encodePacked(
-            callData,
-            msg.sender,
-            msg.value
-        );
-
         (bool success, bytes memory result) = extension.call(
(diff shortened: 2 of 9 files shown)


Hugoo:
Task linked: [DEV-5243 (C4 #139 + related to #58) Refactor function _fallbackLSP17Extendable(...) to enable to run code after this function is called](https://app.clickup.com/t/2645698/DEV-5243)

YamenMerhi (approved):

skimaharvey (approved):
Pretty cool

PR #678, opened 2023-08-14
feat!: add first implementation of LSP25 Execute Relay Call + update LSP6 interface ID

CJ42:

### Change in LSP6 interface

The 3 x functions `getNonce(address,uint128)`, `executeRelayCall(...)` and `executeRelayCallBatch(...)` are part of their own standard LSP25 Execute Relay Call.

**This change the interface ID of LSP6 from `0x38bb3cdb` --> to `0x627ca5d3 ` **

`LSP6KeyManagerCore` now also supports the interface ID of LSP25: `0x5ac79908`

### Change in the version to sign for `executeRelayCall`

Since Relay Calls / Meta Transactions are now in their own standard LSP25, the version used to sign execute relay call has changed:

- before: number `6`
- **now: number `25` ✅ **

## 🚀 Feature

- Add `abstract` contract `LSP25MultiChannelNonce` that contains all the logic and handle the signature validation 

## ♻️ Refactor

- Add LSP25 in inheritance of `LSP6KeyManagerCore`
- add LSP25 interface ID in `supportsInterface(bytes4)` function of `LSP6KeyManagerCore`.

## 🧪 Tests

- [ ] More tests to be defined for the LSP25 standard, so that this implementation is not on its own without any unit tests.
- [ ] Add them in the CI and in `package.json` as `npm run test:lsp25`

## ⚡️ Performance

- [ ] Gas diff to be checked compared to before.

## 📄 Documentation

- [ ] Natspec comments to be reviewed for LSP25 vs LSP6 since the functions have been migrated out of LSP6. Also check if any LSP6 function or references are still old or outdated.
- Change in the README in the `import` section for the constants from `LSP6_VERSION` to `LSP25_VERSION`.

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/LSP25ExecuteRelayCall/LSP25Errors.sol
+++ b/contracts/LSP25ExecuteRelayCall/LSP25Errors.sol
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.4;
+
+/**
+ * @dev Reverts when the `signer` address retrieved from the `signature` has an invalid nonce: `invalidNonce`.
+ *
+ * @param signer The address of the signer
+ * @param invalidNonce The nonce retrieved for the `signer` address
+ * @param signature The signature used to retrieve the `signer` address
+ */
+error InvalidRelayNonce(address signer, uint256 invalidNonce, bytes signature);
+
+/**
+ * @dev Reverts when the start timestamp provided to {executeRelayCall} function is bigger than the current timestamp.
+ */
+error RelayCallBeforeStartTime();
+
+/**
+ * @dev Reverts when the period to execute the relay call has expired.
+ */
+error RelayCallExpired();
--- a/contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol
+++ b/contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.4;
+
+import {LSP25_VERSION} from "./LSP25Constants.sol";
+
+// libraries
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+// errors
+import {
+    InvalidRelayNonce,
+    RelayCallBeforeStartTime,
+    RelayCallExpired
+} from "./LSP25Errors.sol";
+
+abstract contract LSP25MultiChannelNonce {
+    using ECDSA for *;
+
+    // Mapping of signer -> channelId -> nonce in channel
+    mapping(address => mapping(uint256 => uint256)) internal _nonceStore;
+
+    function getNonce(
+        address from,
+        uint128 channelId
+    ) public view virtual returns (uint256) {
+        uint256 nonceInChannel = _nonceStore[from][channelId];
+        return (uint256(channelId) << 128) | nonceInChannel;
+    }
+
+    function _validateExecuteRelayCall(
+        bytes memory signature,
+        uint256 nonce,
+        uint256 validityTimestamps,
+        uint256 msgValue,
+        bytes calldata callData
+    ) internal returns (address signerAddress) {
+        bytes memory encodedMessage = abi.encodePacked(
+            LSP25_VERSION,
+            block.chainid,
+            nonce,
+            validityTimestamps,
+            msgValue,
+            callData
+        );
+
+        signerAddress = address(this)
+            .toDataWithIntendedValidatorHash(encodedMessage)
+            .recover(signature);
+
+        if (!_isValidNonce(signerAddress, nonce)) {
+            revert InvalidRelayNonce(signerAddress, nonce, signature);
+        }
+
+        // increase nonce after successful verification
+        _nonceStore[signerAddress][nonce >> 128]++;
+
+        if (validityTimestamps != 0) {
+            uint128 startingTimestamp = uint128(validityTimestamps >> 128);
+            uint128 endingTimestamp = uint128(validityTimestamps);
+
+            // solhint-disable not-rely-on-time
+            if (block.timestamp < startingTimestamp) {
+                revert RelayCallBeforeStartTime();
+            }
+            if (block.timestamp > endingTimestamp) {
+                revert RelayCallExpired();
+            }
+        }
+
+        return signerAddress;
+    }
+
+    /**
+     * @notice verify the nonce `_idx` for `_from` (obtained via `getNonce(...)`)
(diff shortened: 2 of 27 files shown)


CJ42:
## Gas cost changes

image

contracts/LSP25ExecuteRelayCall/LSP25Errors.sol
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @dev Reverts when the relay call is cannot yet bet executed.
+ * This mean that the starting timestamp provided to {executeRelayCall} function is bigger than the current timestamp.
+ */
+error RelayCallBeforeStartTime();
+
+/**
+ * @dev Reverts when the period to execute the relay call has expired.
+ */
+error RelayCallExpired();
```
b00ste, line 15: We need `@notice` tag here

contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        bytes32 eip191Hash = address(this).toDataWithIntendedValidatorHash(
+            lsp25EncodedMessage
+        );
+
+        return eip191Hash.recover(signature);
+    }
+
+    /**
+     * @notice Verifying if the current timestamp is within the date and time range provided by `validityTimestamps`.
```
b00ste, line 99: `@notice` tag not needed, as the function is internal. I think this information can be moved to `@dev`

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -70,9 +74,11 @@ abstract contract LSP6KeyManagerCore is
     ERC165,
     ILSP6KeyManager,
```
b00ste, line 75: I think we can simplify the `ILSP6KeyManager` to `ILSP6` If we import like this:

`import {ILSP6KeyManager as ILSP6} from "./ILSP6KeyManager.sol";`

dodoc/config.ts
```diff
@@ -34,6 +34,7 @@ export const dodocConfig = {
     'contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8Enumerable.sol',
     'contracts/LSP8IdentifiableDigitalAsset/presets/LSP8CompatibleERC721Mintable.sol',
     'contracts/LSP8IdentifiableDigitalAsset/presets/LSP8Mintable.sol',
+    'contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol',
```
b00ste, line 37: I am not sure this should be under tokens comment

b00ste (approved):
Some small notes, otherwise LGTM

richtera:
I was trying to patch this into the extension but I am seeing build errors:
```console
DocstringParsingError: Documentation tag @inheritdoc references inexistent contract "ILSP6KeyManager".
  --> contracts/LSP6KeyManager/LSP6KeyManagerCore.sol:93:5:
   |
93 |     /**
   |     ^ (Relevant source part starts here and spans across multiple lines).
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -385,18 +419,23 @@ abstract contract LSP6KeyManagerCore is
...
 
-        address signer = address(this)
-            .toDataWithIntendedValidatorHash(encodedMessage)
-            .recover(signature);
+        if (!_isValidNonce(signer, nonce)) {
+            revert InvalidRelayNonce(signer, nonce, signature);
+        }
+
+        // increase nonce after successful verification
+        _nonceStore[signer][nonce >> 128]++;
```
skimaharvey, line 436: would it be fine if in the `isValidNonce`
we do this `return (idx & mask) == _nonceStore[from][idx >> 128]++;`
to save a little on gas?

contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol
```diff
@@ -0,0 +1,143 @@
...
+        uint128 endingTimestamp = uint128(validityTimestamps);
+
+        // solhint-disable not-rely-on-time
+        if (block.timestamp < startingTimestamp) {
+            revert RelayCallBeforeStartTime();
+        }
+        if (block.timestamp > endingTimestamp) {
+            revert RelayCallExpired();
+        }
+    }
```
skimaharvey, line 119: spoke to @YamenMerhi earlier about that but I believe it would make more sense that we consider `uint128(0)` to be the `undefined` value.
Meaning if I do not want to mention a `startingTimestamp` or a `endingTimestamp` we should just leave this value empty. 
because right now if we want both to be empty we just leave them both empty (which makes sense) and it will return on line `107` 
however if I only want to have a `startingDate` I will need to give a value to the `endingDate`. A "random" value or `bytes(f)` but why? 
For UX reasons I think I would delibarately leave the values are initialized. More logic. Specially with our logic on line 107.

skimaharvey (approved):
Big change. Looks good to me.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42: I am ok with that, but then we might have to document that the function also increments the nonce, and not only check it. So it is not a `view` function anymore than can be used standalone.

@skimaharvey @YamenMerhi @b00ste wdyt?

contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol
CJ42: > Meaning if I do not want to mention a startingTimestamp or a endingTimestamp we should just leave this value empty.

Agree here, such value for a `validityTimestamp` should work:

image

```
// 64E69D80 -> valid from 24th August 2023
0x00000000000000000000000064E69D800000000000000000000000000000000
```

YamenMerhi (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: I would say it depends how big of a gas difference does it make?

CJ42: > I would say it depends how big of a gas difference does it make?

After testing, it reduce runtime cost by around 500 gas for the two cases mentioned above, and deployment cost of the Key Manager by 4k.

We decided to keep the function as it is to keep it generic.

PR #688, opened 2023-08-24
ci: replace github action to check if version changed.

CJ42:

## 🤖 CI

Replace Github action to check if version changed by: https://github.com/marketplace/actions/semver-utils
The current github action does not work with release candidates or pre-releases, where the version in `package.json` contains `rc`, for instance `v0.11.0-rc.2`.


richtera (approved):

.github/workflows/build-artifacts.yml
```diff
@@ -10,13 +10,34 @@ jobs:
   build-artifacts:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout main branch
+        uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.base.sha }}
           fetch-depth: 0
```
kalote, line 17: Do you need all history + branches here? It may be improved by reducing the amount of info you're fetching (e.g., fetch-depth: #_of_commits_you_need)

.github/workflows/build-artifacts.yml
```diff
@@ -10,13 +10,34 @@ jobs:
...
 
-      - name: Check if version changed
-        uses: EndBug/version-check@v2
+      - name: Grab version in `package.json` in main
+        run: node -p "'version=' + require('./package.json').version" >> $GITHUB_OUTPUT
```
kalote, line 20: can be simplified with `cat package.json | jq -r .version` if jq is installed

.github/workflows/build-artifacts.yml
CJ42: I don't need the branches, I just need the history of this PR where the Github action ran, to compare the `package.json`.

.github/workflows/build-artifacts.yml
CJ42: Ok I will try this out :)

.github/workflows/build-artifacts.yml
kalote: For me, you only need the last commit (e.g., `fetch-depth: 1` (which is the default)). This is not mandatory, just a quick improvement. Hope to you 👍

CJ42: Ok I will keep it like this in case do not edit too much in the CI. As I worry the commit that updated the version is further down in the commit history and could be missed.

kalote (approved):

PR #695, opened 2023-08-31
chore: [C4] Add first batch of Gas optimization

YamenMerhi:

- Add first batch of Gas optimization

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`


contracts/LSP25ExecuteRelayCall/LSP25MultiChannelNonce.sol
```diff
@@ -56,8 +56,7 @@ abstract contract LSP25MultiChannelNonce {
         address from,
         uint128 channelId
     ) internal view virtual returns (uint256 idx) {
-        uint256 nonceInChannel = _nonceStore[from][channelId];
```
CJ42, line 59: Hmm this seems a bit less readable to me. But not a big deal otherwise, I am happy with the changes.

What do you think @skimaharvey @b00ste @YamenMerhi ?

contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol
```diff
@@ -384,15 +384,17 @@ abstract contract LSP7DigitalAssetCore is ILSP7DigitalAsset {
                     amount
                 );
             }
-            _operatorAuthorizedAmount[from][operator] -= amount;
+            _operatorAuthorizedAmount[from][operator] =
```
CJ42, line 387: nice so this removes a `SLOAD` basically? 👌

CJ42 (approved):

contracts/LSP7DigitalAsset/LSP7DigitalAssetCore.sol
```diff
@@ -384,15 +384,17 @@ abstract contract LSP7DigitalAssetCore is ILSP7DigitalAsset {
...
+                amount;
         }
 
         _beforeTokenTransfer(from, address(0), amount);
 
         // tokens being burnt
         _existingTokens -= amount;
 
-        _tokenOwnerBalances[from] -= amount;
+        _tokenOwnerBalances[from] = balance - amount;
```
skimaharvey, line 397: im wondering if it is not dangerous particularly with reentrancies with the `_beforeTokenTransfer` hooks. Possibly `balance` will not be the same

skimaharvey: in our case `_beforeTokenTransfer` does not do anything but possibly in some implementation it will. We could then face some issues where the value of `balance` has changed. 
I think we should discard this change

b00ste (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -329,37 +331,43 @@ abstract contract LSP6KeyManagerCore is
             isSetData = true;
         }
 
+        address target_ = _target;
```
CJ42, line 334: Not a big fan of using postfix underscore (except `constructor` arguments). Would suggest a different variable name in camelCase to maintain Solidity style guide

```suggestion
        address targetContract = _target;
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -329,37 +331,43 @@ abstract contract LSP6KeyManagerCore is
...
+            );
 
-            _verifyPermissions(caller, msgValue, data);
+            _verifyPermissions(target_, caller, msgValue, data);
             emit PermissionsVerified(caller, msgValue, bytes4(data));
 
             // if it's a setData call, do not invoke the `lsp20VerifyCallResult(..)` function
             return
-                isSetData || isReentrantCall
+                isSetData || (reentrancyStatus == _ENTERED)
```
CJ42, line 349: I think because of the order of precedence of operators, `isSetData` should always be checked first.

See from the Solidity docs. But good idea to surround it in parentheses in case 👌 

image

image

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -396,14 +404,20 @@ abstract contract LSP6KeyManagerCore is
             isSetData = true;
         }
 
-        bool isReentrantCall = _nonReentrantBefore(isSetData, msg.sender);
+        address target_ = _target;
```
CJ42, line 407: same comment as above

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -481,10 +501,11 @@ abstract contract LSP6KeyManagerCore is
      * @return bytes The data returned by the call made to the linked {target} contract.
      */
     function _executePayload(
+        address target_,
```
CJ42, line 504: same comment as above. It feels inconsistent to have one parameter name that is prepended with a `_`. Suggest renaming to this to make it more explicit with the word _"Contract"_, as we are running a payload in a targeted smart contract

```suggestion
        address targetContract,
```

contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
```diff
@@ -222,9 +222,8 @@ abstract contract LSP8IdentifiableDigitalAssetCore is
         address caller,
         bytes32 tokenId
     ) internal view virtual returns (bool) {
-        address tokenOwner = tokenOwnerOf(tokenId);
-
-        return (caller == tokenOwner || _operators[tokenId].contains(caller));
+        return (caller == tokenOwnerOf(tokenId) ||
```
CJ42, line 225: We will need to increase the line width for Solidity files. This is hard to read (@skimaharvey )

CJ42 (changes requested):

CJ42 (approved):

CJ42:
Amazing! 🚀

PR #717, opened 2023-09-14
fix: add check for permission value length in `_getPermissionToSetControllerPermissions` and in `LSP6Utils` function `getPermissionsFor`

CJ42:

## 🐛 Bug

- **Important Bug Fix**: fix the check against permissions `ADD_CONTROLLER` vs `EDIT_PERMISSIONS`. If the permission value stored under the controller we are setting the permissions for is not exactly 32 bytes long, consider it as _"no permissions set"_, and **require the permission `ADD_CONTROLLER`**.

- Add check in function `getPermissionsFor` to ensure that the raw `bytes` value of a permission fetched from the ERC725Y storage are **exactly 32 bytes long**. If they raw bytes value is not 32 bytes long, consider this as no permissions check.

This is to make it consistent with the input validation below in `LSP6SetDataModule`.

image

## 📄 Documentation

- Add info Natspec tag above this function and update auto generated docs.

## 🧪 Tests

- Add extra tests for the new bug fix mentioned above, depending on the caller's permissions.

### DIFF
--- a/contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
+++ b/contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
@@ -431,12 +431,16 @@ abstract contract LSP6SetDataModule {
         address controlledContract,
         bytes32 inputPermissionDataKey
     ) internal view virtual returns (bytes32) {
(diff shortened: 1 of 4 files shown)


skimaharvey (approved):
Agreed 👍

b00ste (approved):

PR #725, opened 2023-09-20
feat: allow sending value when using `setData(..)` through the LSP6

b00ste:

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/LSP20CallVerification/LSP20CallVerification.sol
+++ b/contracts/LSP20CallVerification/LSP20CallVerification.sol
@@ -44,8 +44,9 @@ abstract contract LSP20CallVerification {
 
         bytes4 magicValue = abi.decode(returnedData, (bytes4));
 
-        if (bytes3(magicValue) != bytes3(ILSP20.lsp20VerifyCall.selector))
+        if (bytes3(magicValue) != bytes3(ILSP20.lsp20VerifyCall.selector)) {
             revert LSP20InvalidMagicValue(false, returnedData);
+        }
 
         return magicValue[3] == 0x01;
     }
--- a/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
+++ b/contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
@@ -46,8 +46,7 @@ import {
     InvalidPayload,
     InvalidRelayNonce,
     NoPermissionsSet,
-    InvalidERC725Function,
-    CannotSendValueToSetData
+    InvalidERC725Function
 } from "./LSP6Errors.sol";
 
 import {
@@ -335,7 +334,8 @@ abstract contract LSP6KeyManagerCore is
                 caller
             );
 
-            _verifyPermissions(targetContract, caller, msgValue, false, data);
+            _verifyPermissions(targetContract, caller, false, data);
+
             emit PermissionsVerified(caller, msgValue, bytes4(data));
 
             // if it's a setData call, do not invoke the `lsp20VerifyCallResult(..)` function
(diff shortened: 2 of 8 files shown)


contracts/LSP20CallVerification/LSP20CallVerification.sol
```diff
@@ -39,10 +39,11 @@ abstract contract LSP20CallVerification {
 
         bytes4 magicValue = abi.decode(returnedData, (bytes4));
 
-        if (bytes3(magicValue) != bytes3(ILSP20.lsp20VerifyCall.selector))
+        if (bytes3(magicValue) != bytes3(ILSP20.lsp20VerifyCall.selector)) {
             revert LSP20InvalidMagicValue(false, returnedData);
+        }
 
-        return magicValue[3] == 0x01 ? true : false;
+        return magicValue[3] == 0x01;
```
CJ42, line 46: Nice 👌

tests/LSP6KeyManager/Interactions/BatchExecute.test.ts
```diff
@@ -425,19 +425,16 @@ export const shouldBehaveLikeBatchExecute = (
       });
 
       describe('if specifying some value for each values[index]', () => {
-        it('should revert with Key Manager error `CannotSendValueToSetData` when sending value while setting data', async () => {
-          const amountToFund = ethers.utils.parseEther('2');
+        it('should pass when sending value while setting data', async () => {
+          const msgValues = [ethers.utils.parseEther('2'), ethers.utils.parseEther('2')];
+          const totalMsgValue = ethers.utils.parseEther('4');
```
CJ42, line 430: Could be done with `.reduce` like it is done below.

```js
const totalMsgValue = msgValues.reduce((accumulator, currentValue) =>
            accumulator.add(currentValue),
          );
```

CJ42 (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -528,13 +530,13 @@ abstract contract LSP6KeyManagerCore is
...
     /**
      * @dev Verify if the `from` address is allowed to execute the `payload` on the {target} contract linked to this Key Manager.
+     * @param targetContract The contract that is owned by the Key Manager
      * @param from Either the caller of {execute} or the signer of {executeRelayCall}.
      * @param payload The abi-encoded function call to execute on the {target} contract.
      */
     function _verifyPermissions(
         address targetContract,
         address from,
-        uint256 msgValue,
```
CJ42, line 537: We need to keep this parameter, as we also need to check for permissions.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -544,7 +546,6 @@ abstract contract LSP6KeyManagerCore is
 
         // ERC725Y.setData(bytes32,bytes)
         if (erc725Function == IERC725Y.setData.selector) {
-            if (msgValue != 0) revert CannotSendValueToSetData();
```
CJ42, line 547: Since we are removing the custom error, we need to check for permission `TRANSFERVALUE` in addition to check for the permission `SETDATA` below via `_verifyCanSetData(...)`.

CJ42 (changes requested):
If we allow to send native tokens while setting data, we need now to add the permission check of `TRANSFERVALUE` while setting data + tests for it.

I suggest we park this PR on the side and put it as draft for now until we have discussed and agreed on the decision with @frozeman

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: why? It is just founding the account when setting data? not transfering?

CJ42: Yes indeed. Got it wrong

skimaharvey (approved):
LGTM

YamenMerhi (approved):

PR #727, opened 2023-09-22
chore: use `Apache-2.0` license across all `.sol` files

CJ42:

Change the license used by all the `.sol` files to use `Apache-2.0`.


YamenMerhi (approved):
To be checked with @skimaharvey about LSP23 bytecode changes, LSP16 is already changed and have a commitHash where it should be deployed.

skimaharvey (approved):

PR #729, opened 2023-09-26
feat!: add callee params to lsp20 & EXECUTE_RELAY_CALL permission & mapping reentrancyStatus

skimaharvey:

## ⚠️ BREAKING CHANGES

- [x] Update `lsp20` interface with the `callee` params 
- [x] Update `lsp20` constant file 
- [x] Update `lsp20` mocks contracts 
- [x] Add `EXECUTE_RELAY_CALL` permission
- [x] Change `reentrancyStatus` from `uint8` to `mapping(address => uint8)`
- [x] Update `constant.ts` file


contracts/LSP20CallVerification/ILSP20CallVerifier.sol
```diff
@@ -12,11 +12,13 @@ interface ILSP20CallVerifier {
      * the function is allowed, concatened with a byte that determines if the lsp20VerifyCallResult function should
      * be called after the original function call. The byte that invoke the lsp20VerifyCallResult function is strictly `0x01`.
      *
+     * @param callee The address of the contract that implements the LSP20 interface
      * @param caller The address who called the function on the msg.sender
      * @param value The value sent by the caller to the function called on the msg.sender
      * @param receivedCalldata The receivedCalldata sent by the caller to the msg.sender
      */
     function lsp20VerifyCall(
+        address callee,
```
CJ42, line 21: would suggest changing the order to 1st `caller` and then `callee`. Sounds more natural to me.

b00ste (approved):

contracts/LSP20CallVerification/ILSP20CallVerifier.sol
skimaharvey: I personally like this order because in most cases the `callee` will be `address(this)`. Also feel like it is the most important contract so it should come first but no strong opinions

constants.ts
```diff
@@ -274,7 +274,7 @@ export const CALLTYPE = {
 /**
  * @dev `bytes32` hex value for all the LSP6 permissions excluding REENTRANCY, DELEGATECALL and SUPER_DELEGATECALL for security (these should be set manually)
  */
-export const ALL_PERMISSIONS = '0x00000000000000000000000000000000000000000000000000000000003f3f7f';
+export const ALL_PERMISSIONS = '0x00000000000000000000000000000000000000000000000000000000007f3f7f';
```
CJ42, line 277: I have an idea 💡 

Let's add a tests in the `shouldInitializeLikeLSP6` test suite to test that this constant literal value has the same result as `combinePermissions([...])`.

This way, we will always be sure we put the right value and we don't have the same problem we encountered @skimaharvey where we had to debug.
It will be like the ERC165 interface IDs checks.

contracts/LSP20CallVerification/ILSP20CallVerifier.sol
```diff
@@ -12,11 +12,13 @@ interface ILSP20CallVerifier {
      * the function is allowed, concatened with a byte that determines if the lsp20VerifyCallResult function should
      * be called after the original function call. The byte that invoke the lsp20VerifyCallResult function is strictly `0x01`.
      *
+     * @param callee The address of the contract that implements the LSP20 interface
```
CJ42, line 15: Which of the LSP20 interface? I guess the `LSP20CallVerifier`. Suggesting to precise this in the comment:

```suggestion
     * @param callee The address of the contract that implements the `LSP20CallVerifier` interface
```

image

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -94,7 +98,7 @@ abstract contract LSP6KeyManagerCore is
 
     // Variables, methods and modifier used for ReentrancyGuard are taken from the link below and modified accordingly.
     // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.8/contracts/security/ReentrancyGuard.sol
-    uint8 internal _reentrancyStatus;
+    mapping(address => uint8) internal _reentrancyStatus;
```
CJ42, line 101: Since we are now moving to a `mapping` for the reentrancy guard, the `uint8` is not needed anymore.
In the past, `uint8` was mainly to pack the reentrancy guard into the same storage slot as the `target_`.

image

So in the function to check for permission `REENTRANCY`, only one `SLOAD` would be necessary to check for reentrancy (reentrancy guard + target contract would be fetched all at once from the same storage slot).

However, since now it is a `mapping`, mappings variable has its own storage slot, and values for each data keys have their own dedicated storage slot computed as follow:

image

My assumption is that now, using a `uint8` inside the mapping actually add more gas, because it needs to convert from `uint256` to `uint8` after reading the value from storage. So I suggest we move the reentrancy guard values to a `uint256`, including these two constants:

```solidity
    uint256 internal constant _NOT_ENTERED = 1;
    uint256 internal constant _ENTERED = 2;
```

```suggestion
    mapping(address => uint256) internal _reentrancyStatus;
```

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -584,13 +595,22 @@ abstract contract LSP6KeyManagerCore is
         } else {
             revert InvalidERC725Function(erc725Function);
         }
+
+        if (isRelayedCall) {
```
CJ42, line 599: Maybe we could move this check at the top before checking for the function selector?

CJ42 (commented):
Added first round of review

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: good point

constants.ts
skimaharvey: sounds good may be we can create a ticket for that and do another PR?

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
skimaharvey: Done ✅

contracts/LSP20CallVerification/LSP20CallVerification.sol
```diff
@@ -33,6 +33,7 @@ abstract contract LSP20CallVerification {
         (bool success, bytes memory returnedData) = logicVerifier.call(
             abi.encodeWithSelector(
                 ILSP20.lsp20VerifyCall.selector,
+                address(this),
```
YamenMerhi, line 36: In order to make the LSP20 modules reusable, I would suggest to parameterize in `_verifyCall` all the lsp20 functions parameter, because not in all scenarios the address(this) will be the target or `msg.sender` is the one to verify permissions for.

But probably can be done in later PRs, just a note

YamenMerhi (approved):

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -589,8 +607,10 @@ abstract contract LSP6KeyManagerCore is
     /**
      * @dev Initialise _reentrancyStatus to _NOT_ENTERED.
      */
-    function _setupLSP6ReentrancyGuard() internal virtual {
-        _reentrancyStatus = 1;
+    function _setupLSP6ReentrancyGuard(
+        address targetContract
+    ) internal virtual {
+        _reentrancyStatus[targetContract] = 1;
```
CJ42, line 613: Let's use the constant values:

```suggestion
        _reentrancyStatus[targetContract] = _NOT_ENTERED;
```

CJ42: I am always bugged every time I am looking at this function. Why do we keep this function if the `mapping` / `reentrancyStatus` is now set to `internal` and accessible through the `constructor` / `initialize` function in the derived contract?

Also the reason why this `_setupLSP6ReentrancyGuard` function bugs me so much is because it does exactly the same thing as this function below:

image

I don't understand the reason why we are keeping it. It adds unecessary bytecode in the end.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -589,8 +607,10 @@ abstract contract LSP6KeyManagerCore is
     /**
      * @dev Initialise _reentrancyStatus to _NOT_ENTERED.
```
CJ42: if moving to boolean, let's not forget to change this comment too.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42: If we are moving to boolean, this will be set to `false`.

PR #743, opened 2023-10-11
Lsp6/singleton

skimaharvey:
# THIS PR IS PURELY FOR INTERNAL TESTING AND CODE WITHIN IT SHOULD NOT BE RE-USED

- [x] Create LSP6KeyManagerSingleton contract
- [x] Create LSP25 contract so that it works with singleton contract
- [x] Add Relay calls tests to LSP6KeyManagerSingleton


contracts/Factory4337.sol
```diff
@@ -0,0 +1,156 @@
...
+    function createAccount(
+        address implementationContract,
+        address keyManagerSingleton,
+        address mainController,
+        address extension4337,
+        uint256 salt
+    ) public returns (address) {
+        address universalProfileAddress = Clones.cloneDeterministic(
+            implementationContract,
+            keccak256(abi.encodePacked(mainController, salt))
```
skimaharvey, line 47: need to add the `keyManagerSingleton` to prevent squatting

CJ42:
Closing for now as not planning to implement yet, and as the branch suggests it is only for internal testing

PR #745, opened 2023-10-12
fix: bug in dodoc config, where function signature was wrongly generated 

b00ste:

## 🐛 Bug

In this PR a bug is fixed where function with unnamed parameters have the signature generated wrongly.

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`


dodoc/config.ts
```diff
@@ -201,11 +201,25 @@ const formatCode = (code: string, type: string) => {
     .trim();
 
   if (!formatedCode.endsWith('()')) {
+    const start = `${formatedCode.split('(')[0]}(`;
+    const end = ')';
+    const middle = formatedCode.replace(start, '').replace(end, '');
```
CJ42, line 206: typo in variable name

```suggestion
    const middle = formattedCode.replace(start, '').replace(end, '');
```

dodoc/config.ts
```diff
@@ -201,11 +201,25 @@ const formatCode = (code: string, type: string) => {
...
+
     formatedCode =
-      formatedCode
+      start +
+      middle
         .split(',')
-        .map((elem) => elem.trim().substring(0, elem.trim().indexOf(' ')))
-        .toString() + ')';
+        .map((elem) => {
```
CJ42, line 212: Maybe you can put this into a constant function to make it easier to read?

```ts
const removeParameters = (content: string) => {
    return content.split(',')
        .map((elem) => {
        const trimmedElem = elem.trim();

          if (trimmedElem.includes(' ')) {
            return trimmedElem.substring(0, elem.trim().indexOf(' '));
          } else {
            return trimmedElem;
          }
          });
}
```

CJ42 (approved):

CJ42 (approved):

YamenMerhi (approved):

PR #748, opened 2023-10-12
refactor: replace `BytesLib` in LSP8 with assembly block

b00ste:

## ♻️ Refactor

Add an assembly block in order to remove `BytesLib` usage from `LSP8CompatibleERC721.sol` & `LSP8CompatibleERC721InitAbstract.sol`

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
@@ -15,7 +15,6 @@ import {
 import {
     EnumerableSet
 } from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
-import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
 
 // modules
 import {
@@ -103,15 +102,25 @@ abstract contract LSP8CompatibleERC721 is
     ) public view virtual returns (string memory) {
         bytes memory data = _getData(_LSP4_METADATA_KEY);
 
-        // offset = bytes4(hashSig) + bytes32(contentHash) -> 4 + 32 = 36
-        uint256 offset = 36;
+        if (data.length <= 36) return "";
 
-        bytes memory uriBytes = BytesLib.slice(
-            data,
-            offset,
-            data.length - offset
-        );
-        return string(uriBytes);
+        uint256 requiredLength = data.length - 36;
+
+        string memory result = new string(requiredLength);
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            // set the actual value that will be returned
+            for {
+                let i := 0
+            } lt(i, requiredLength) {
+                i := add(i, 32)
+            } {
+                mstore(add(result, add(i, 32)), mload(add(data, add(i, 68))))
+            }
+        }
+
+        return result;
     }
 
     /**
--- a/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721InitAbstract.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721InitAbstract.sol
@@ -15,7 +15,6 @@ import {
 import {
     EnumerableSet
 } from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
-import {BytesLib} from "solidity-bytes-utils/contracts/BytesLib.sol";
 
 // modules
 import {
@@ -111,15 +110,25 @@ abstract contract LSP8CompatibleERC721InitAbstract is
     ) public view virtual returns (string memory) {
         bytes memory data = _getData(_LSP4_METADATA_KEY);
 
-        // offset = bytes4(hashSig) + bytes32(contentHash) -> 4 + 32 = 36
-        uint256 offset = 36;
+        if (data.length <= 36) return "";
 
-        bytes memory uriBytes = BytesLib.slice(
-            data,
-            offset,
-            data.length - offset
-        );
-        return string(uriBytes);
+        uint256 requiredLength = data.length - 36;
+
+        string memory result = new string(requiredLength);
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            // set the actual value that will be returned
+            for {
+                let i := 0
+            } lt(i, requiredLength) {
+                i := add(i, 32)
+            } {
+                mstore(add(result, add(i, 32)), mload(add(data, add(i, 68))))
+            }
+        }
+
+        return result;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
```diff
@@ -102,15 +102,24 @@ abstract contract LSP8CompatibleERC721 is
...
+                let i := 0
+            } lt(i, requiredLength) {
+                i := add(i, 0x20)
+            } {
+                mstore(
+                    add(uriBytes, add(i, 0x20)),
+                    mload(add(data, add(i, 68)))
+                )
+            }
+            mstore(0x40, add(uriBytes, add(0x20, requiredLength)))
```
CJ42, line 121: Need comment here about updating the free memory pointer.

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
```diff
@@ -102,15 +102,24 @@ abstract contract LSP8CompatibleERC721 is
         uint256 /* tokenId */
     ) public view virtual returns (string memory) {
         bytes memory data = _getData(_LSP4_METADATA_KEY);
-
-        // offset = bytes4(hashSig) + bytes32(contentHash) -> 4 + 32 = 36
-        uint256 offset = 36;
-
-        bytes memory uriBytes = BytesLib.slice(
-            data,
```
CJ42, line 110: The function was not even checking before if the length was less than 36 bytes long. If this was the case, BytesLib 
.slice library would have probably reverted.

 Ow with assembly, it will not revert and just return something else (random) from memory.

This can be tricky and dangerous. We should check and test these potential scenarios.

contracts/LSP8IdentifiableDigitalAsset/extensions/LSP8CompatibleERC721.sol
```diff
@@ -102,15 +102,24 @@ abstract contract LSP8CompatibleERC721 is
...
+        assembly {
+            uriBytes := mload(0x40)
+            mstore(uriBytes, requiredLength)
+            for {
+                let i := 0
+            } lt(i, requiredLength) {
+                i := add(i, 0x20)
+            } {
+                mstore(
+                    add(uriBytes, add(i, 0x20)),
```
CJ42, line 117: Suggest writing in decimal for readability (except the free memory pointer 0x40)

```suggestion
                    add(uriBytes, add(i, 32)),
```

CJ42 (changes requested):

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -473,4 +473,83 @@ library LSP2Utils {
...
+     * @dev Uses assembly to remove the first `n` bytes from `data`.
+     *
+     * @param data The data that needs to be trimmed.
+     * @param n Number of bytes that will be removed.
+     */
+    function removeFirstNBytes(
+        bytes memory data,
+        uint256 n
+    ) internal pure returns (bytes memory) {
+        assembly {
```
CJ42, line 487:
```suggestion
        // solhint-disable-next-line no-inline-assembly
        assembly {
```

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -473,4 +473,83 @@ library LSP2Utils {
...
+            // return if `data` length is smaller that `n`
+            if lt(mload(data), n) {
+                let res := mload(0x40)
```
CJ42, line 490: Please write variable in plain name. Especially in assembly. Assembly is already very hard to read. So local assembly variables have to be very clear on what they are.

```suggestion
                let result := mload(0x40)
```

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -473,4 +473,83 @@ library LSP2Utils {
...
+
+                mstore(0x40, add(res, 96))
+
+                return(add(res, 32), 64)
+            }
+
+            let requiredLength := sub(mload(data), n)
+
+            // create new variable at free memory pointer
+            let result := mload(0x40)
```
CJ42, line 516: Let's move fetching the free memory pointer at the top

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -473,4 +473,83 @@ library LSP2Utils {
...
+            if eq(mload(data), n) {
+                let res := mload(0x40)
+
+                mstore(res, 64)
+                mstore(add(res, 32), 32)
+                mstore(add(res, 64), 0)
+
+                mstore(0x40, add(res, 96))
+
+                return(add(res, 32), 64)
```
CJ42, line 510: Maybe using assembly `return` here might be tricky, as it will return the whole function call. This might not be ideal if this function or `tokenURI()` is used inside an other public or internal function.

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -473,4 +473,83 @@ library LSP2Utils {
...
+            }
+            // return if `data` length is equal to `n`
+            if eq(mload(data), n) {
```
CJ42, line 501: I think we can refactor these two blocks as:

```
if (data.length <= n) return data;
```

contracts/LSP2ERC725YJSONSchema/LSP2Utils.sol
```diff
@@ -473,4 +473,83 @@ library LSP2Utils {
...
         dataValues[4] = lastElementMapValue;
     }
+
+    /**
+     * @dev Uses assembly to remove the first `n` bytes from `data`.
+     *
+     * @param data The data that needs to be trimmed.
+     * @param n Number of bytes that will be removed.
+     */
+    function removeFirstNBytes(
```
CJ42, line 483: I would not recommend to put this into the `LSP2Utils` library. Because it is not related to LSP2 directly, and also because we do not want to encourage people to use assembly code.

CJ42 (changes requested):

CJ42:
@b00ste @skimaharvey @YamenMerhi If `BytesLib` is not used anywhere anymore, I suggest we remove it from the dependencies inside `package.json`, so that there are less things being installed when installing `@lukso/lsp-smart-contracts`.

https://github.com/lukso-network/lsp-smart-contracts/blob/e3034e12b2c1d78347145580dbcfe3e553d40716/package.json#L110

CJ42:
Since we are keeping the `BytesLib` in the Key Manager, this PR is not applicable anymore. We can make the changes in the future if we decide to optimise to not use `BytesLib` and remove it completely from the `package.json`.

PR #764, opened 2023-10-26
feat!: create `Version` contract and inherit it

b00ste:

## 🚀 Feature

This PR creates the `Version.sol` contract that allows to track the package version in which the contract was released.

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/Version.sol
+++ b/contracts/Version.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.4;
+
+abstract contract Version {
+    string internal constant _VERSION = "0.12.0";
+
+    /**
+     * @dev Get the version of the contract.
+     * @notice Contract version.
+     *
+     * @return The version of the the contract.
+     */
+    function version() public view virtual returns (string memory) {
+        return _VERSION;
+    }
+}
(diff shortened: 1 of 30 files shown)


Hugoo:
Task linked: [DEV-8832 Create `Version.sol` contract](https://app.clickup.com/t/2645698/DEV-8832)

YamenMerhi (approved):
Would rather have it language-independent being just a getter.
For now, it is important that it goes in LSP0 - LSP1DELEGATEUP - LSP6 - LSP7 - LSP8 for the launch and later decide whether it should go into submodules.

skimaharvey (approved):

contracts/Version.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.4;
+
+abstract contract Version {
+    string private constant _VERSION = "0.12.0";
```
CJ42, line 5: I would mark this constant `internal`, so that child contracts can access this state variable.
This is more efficient at runtime compared to querying the public method `version()` inside the logic of a function.

It will just add a literal in the runtime bytecode, instead of creating the assembly code to do `JUMP` to `version()`.

PR #777, opened 2023-10-31
fix: [M-03] allow setting 21 bytes under LSP17Extension data key in LSP6KeyManager

YamenMerhi:
## 🐛 Bug

- allow setting 21 bytes under LSP17Extension data key in LSP6KeyManager

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
+++ b/contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
@@ -322,7 +322,11 @@ abstract contract LSP6SetDataModule {
             // LSP17Extension:<bytes4>
         } else if (bytes12(inputDataKey) == _LSP17_EXTENSION_PREFIX) {
             // CHECK that `dataValue` contains exactly 20 bytes, which corresponds to an address for a LSP17 Extension
-            if (inputDataValue.length != 20 && inputDataValue.length != 0) {
+            if (
+                inputDataValue.length != 20 &&
+                inputDataValue.length != 21 &&
+                inputDataValue.length != 0
+            ) {
                 revert InvalidDataValuesForDataKeys(
                     inputDataKey,
                     inputDataValue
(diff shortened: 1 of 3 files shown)


CJ42 (approved):
Should we test it with the other restricted controllers as well?

contracts/LSP6KeyManager/LSP6Modules/LSP6SetDataModule.sol
```diff
@@ -322,7 +322,11 @@ abstract contract LSP6SetDataModule {
             // LSP17Extension:<bytes4>
         } else if (bytes12(inputDataKey) == _LSP17_EXTENSION_PREFIX) {
             // CHECK that `dataValue` contains exactly 20 bytes, which corresponds to an address for a LSP17 Extension
```
CJ42, line 324:
```suggestion
            // CHECK that `dataValue` contains exactly 20 or 21 bytes (if setting to forward value), which corresponds to an address for a LSP17 Extension
```

skimaharvey (approved):

b00ste (approved):

tests/LSP6KeyManager/Admin/PermissionChangeAddExtensions.test.ts
```diff
@@ -1133,5 +1133,57 @@ export const shouldBehaveLikePermissionChangeOrAddExtensions = (
         });
       });
     });
+
+    describe('when setting random bytes under the LSP17Extension data key ', () => {
```
CJ42, line 1137: These tests are a copy paste of LSP20.
We are not in a LSP20 context here, we are calling via LSP6.

CJ42 (changes requested):
We cannot write tests like that.
We need to check that validation happen also depending on the controller permission.

For instance, that regardless if controller has `ADDEXTENSION`, or `CHANGEEXTENSION`, or both, the error of invalid value being set occur.

CJ42 (approved):

PR #785, opened 2023-11-03
chore: bootstrap releases for path: . + add `Version.sol` in file to update version for

CJ42:
Configuring release-please for path: `.`

Also add `Version.sol` in the files to update the version number for.


skimaharvey (approved):

contracts/Version.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.4;
 
 abstract contract Version {
-    string internal constant _VERSION = "0.12.0";
+    string internal constant _VERSION = "0.12.0"; // x-release-please-version
```
richtera, line 5: It might be better to use the x-release-please-start-version and end-version block elements. It's very easy for someone to think (oh let me put this comment on the previous line which would break the integration)

.release-please-manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+  ".": "0.12.0"
```
richtera, line 2: We only need to commit the file initially the first time, this will automatically get updated using the release-please release chore PR.

b00ste (approved):

PR #786, opened 2023-11-03
chore: adjust Natspec comments + use mixedCase name for `modifer`

CJ42:


contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -636,9 +636,9 @@ abstract contract LSP6KeyManagerCore is
     }
 
     /**
-     * @dev Update the status from `_NON_ENTERED` to `_ENTERED` and checks if
-     * the status is `_ENTERED` in order to revert the call unless the caller has the REENTRANCY permission
-     * Used in the beginning of the `nonReentrant` modifier, before the method execution starts.
+     * @dev Update the status from `false` to `true` and checks if
+     * the status is `true` in order to revert the call unless the caller has the `REENTRANCY` permission
+     * Used in the beginning of the {`lsp20VerifyCall`}, {`_execute`} and {`_executeRelayCall`} functions, before the methods execution starts.
```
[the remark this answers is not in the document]
JeneaVranceanu: I'm suggesting this only because the steps I imagined from reading this documentation are different from the actual steps this function takes.

On top of that, I see that the `else` condition in the function can be changed to:
```solidity
} else if (!isSetData) {
    _reentrancyStatus[targetContract] = true;
}
```

As there's is no condition for `} else {` and the path is always the same.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
```diff
@@ -663,7 +663,7 @@ abstract contract LSP6KeyManagerCore is
 
     /**
      * @dev Resets the status to `false`
-     * Used in the end of the `nonReentrant` modifier after the method execution is terminated
+     * Used in the end of the {`lsp20VerifyCall`}, {`_execute`} and {`_executeRelayCall`} functions after the methods execution is terminated.
```
JeneaVranceanu, line 666: Shouldn't it be `... after the functions' execution is terminated`?
Not `methods`.

docs/contracts/LSP6KeyManager/LSP6KeyManager.md
```diff
@@ -1224,9 +1224,9 @@ function _nonReentrantBefore(
 ) internal nonpayable returns (bool reentrancyStatus);
 ```
 
-Update the status from `_NON_ENTERED` to `_ENTERED` and checks if
-the status is `_ENTERED` in order to revert the call unless the caller has the REENTRANCY permission
-Used in the beginning of the `nonReentrant` modifier, before the method execution starts.
+Update the status from `false` to `true` and checks if
+the status is `true` in order to revert the call unless the caller has the `REENTRANCY` permission
+Used in the beginning of the [`lsp20VerifyCall`](#`lsp20verifycall`), [`_execute`](#`_execute`) and [`_executeRelayCall`](#`_executerelaycall`) functions, before the methods execution starts.
```
JeneaVranceanu, line 1229: Suggestions written for `contracts/LSP6KeyManager/LSP6KeyManagerCore.sol` apply here as well.

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
CJ42: could do

docs/contracts/LSP6KeyManager/LSP6KeyManager.md
CJ42: @Jenea This Markdown document is auto-generated from the Natspec comment in the `.sol` file.
So updating the `.sol` file will update automatically this file.

For your info, so that you don't have to review the same twice for any future code review :)

contracts/LSP6KeyManager/LSP6KeyManagerCore.sol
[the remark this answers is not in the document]
CJ42: Thanks for pointing that out.
I have re-worded the Natspec comments.

For the `else if`, we discussed internally and prefer to keep it as it is for now for the sake of readability, to distinguish more clearly the condition check for the reentrancy status.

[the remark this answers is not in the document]
CJ42: @JeneaVranceanu I have re-worded the whole Natspec comment for a better readability.

YamenMerhi (approved):

b00ste (approved):

PR #794, opened 2023-11-06
docs: Improve natspec for LSP0

YamenMerhi:

## 📄 Documentation

docs: Improve natspec for LSP0

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -89,13 +89,6 @@ abstract contract LSP0ERC725AccountCore is
     using Address for address;
 
     /**
-
-     *
-     * @custom:events {UniversalReceiver} event when receiving native tokens.
-     */
-
-    /**
-     * @notice Invokes the `universalReceiver` function logic internally when receiving native tokens.
      * @dev Executed:
      * - When receiving some native tokens without any additional data.
      * - On empty calls to the contract.
@@ -799,7 +792,7 @@ abstract contract LSP0ERC725AccountCore is
      * If there is an extension for the function selector being called, it calls the extension with the
      * `CALL` opcode, passing the `msg.data` appended with the 20 bytes of the {msg.sender} and 32 bytes of the `msg.value`.
      *
-     * @custom:hint If you would like to forward the `msg.value` to the extension contract, you should store an additiona `0x01` byte after the address of the extension under the corresponding LSP17 data key.
+     * @custom:hint If you would like to forward the `msg.value` to the extension contract, you should store an additional `0x01` byte after the address of the extension under the corresponding LSP17 data key.
      */
     function _fallbackLSP17Extendable(
         bytes calldata callData
--- a/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
+++ b/contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
@@ -93,7 +93,9 @@ abstract contract LSP0ERC725AccountCore is
      * - When receiving some native tokens without any additional data.
      * - On empty calls to the contract.
      *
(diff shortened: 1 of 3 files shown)


contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -787,14 +799,7 @@ abstract contract LSP0ERC725AccountCore is
...
      *
-     * @custom:hint This function does not forward to the extension contract the `msg.value` received by the contract that inherits `LSP17Extendable`.
-     * If you would like to forward the `msg.value` to the extension contract, you can override the code of this internal function as follow:
-     *
-     * ```solidity
-     * (bool success, bytes memory result) = extension.call{value: msg.value}(
-     *     abi.encodePacked(callData, msg.sender, msg.value)
-     * );
-     * ```
+     * @custom:hint If you would like to forward the `msg.value` to the extension contract, you should store an additiona `0x01` byte after the address of the extension under the corresponding LSP17 data key.
```
CJ42, line 802:
```suggestion
     * @custom:hint If you would like to forward the `msg.value` to the extension contract, you should store an additional `0x01` byte after the address of the extension under the corresponding LSP17 data key.
```

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -89,11 +89,19 @@ abstract contract LSP0ERC725AccountCore is
     using Address for address;
 
     /**
+
+     *
+     * @custom:events {UniversalReceiver} event when receiving native tokens.
```
CJ42, line 94: Why is there this Natspec tag and comment block? It is already present below

image

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -89,11 +89,19 @@ abstract contract LSP0ERC725AccountCore is
...
+     */
+
+    /**
+     * @notice Invokes the `universalReceiver` function logic internally when receiving native tokens.
```
CJ42, line 98: This is not a `@notice` tag, but documentation for developers. Suggesting adding this under the `@dev` tag.

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
```diff
@@ -120,6 +128,8 @@ abstract contract LSP0ERC725AccountCore is
      *
      * 2. If the data sent to this function is of length less than 4 bytes (not a function selector), return.
      *
+     * Whenever the call is associated with native tokens, the function will delegate internally the handling of native tokens to the {universalReceiver} function passing `_TYPEID_LSP0_VALUE_RECEIVED` as typeId and the calldata as received data, except when the native token will be sent directly to the extension.
```
CJ42, line 131: Also worth mentioning when receiving less than 4 bytes of data.

CJ42: > except when the native token will be sent directly to the extension.

This last sentence makes it hard to understand. The only exception according to the code is when:
- you receive some native tokens
-  you receive some data that is more than 4 bytes.

In this case, the associated LSP17 extension is called.

CJ42 (changes requested):

contracts/LSP0ERC725Account/LSP0ERC725AccountCore.sol
YamenMerhi: This is generic saying all the time when native token is associated (even when data is less than 4 bytes), the universalReceiver is invoked, unless the native token is being forwarded to the extension, but happy to add a suggestion if you have one
> Whenever the call is associated with native tokens, the function will delegate internally the handling of native tokens to the {universalReceiver} function passing `_TYPEID_LSP0_VALUE_RECEIVED` as typeId and the calldata as received data, except when the native token will be sent directly to the extension.

CJ42 (approved):

PR #801, opened 2023-11-10
ci: run manifest command to update extra files on release

CJ42:


CJ42:
Putting this back as draft, as we might not want to update the `Version.sol` for the next release, since it will apply only to the tokens and not the other contracts that inherit `Version.sol`

skimaharvey (approved):

PR #802, opened 2023-11-16
feat: extract verification metadata

dzbo:
Move `verification` object to own type so it allow to import it by 3rd parties.

### DIFF
--- a/constants.ts
+++ b/constants.ts
@@ -124,7 +124,7 @@ export type MetadataVerification = {
   method: string;
   data: string;
   source?: string;
-}
+};
 
 export type ImageMetadata = {
   width: number;
--- a/constants.ts
+++ b/constants.ts
@@ -120,14 +120,16 @@ export type LSP4DigitalAssetMetadata = {
   icon: ImageMetadata[];
 };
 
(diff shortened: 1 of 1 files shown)


constants.ts
```diff
@@ -120,14 +120,16 @@ export type LSP4DigitalAssetMetadata = {
   icon: ImageMetadata[];
 };
 
+export type MetadataVerification = {
```
richtera, line 123: But we have this type already and it's called just Verification

richtera, line 123: But we have this type already and it's called just Verification

dzbo: Where is that type?

richtera: https://github.com/lukso-network/tools-lsp-factory/blob/a9780fd2b80cf00f8ab4c2aa4910a890360cbabb/src/lib/interfaces/metadata.ts#L36
You're right it's in another repo.

dzbo: Yeah, it should be moved here. I will rename it tho

richtera (approved):

CJ42 (approved):

PR #830, opened 2023-12-12
docs: add `LSP0ValueReceived` typeId in Natspec with anchor links + `_getNonce`

CJ42:
Keeping this PR as draft until we agree if we can make changes to LSP0 Natspec comments.


YamenMerhi:
I would wait for the monorepo, we already released 0.14.0 which needs to go live in production. If a new release was urgent, the change in LSP0 would make us deploy new version of UP which require new support of other dApps and services .

CJ42:
Closing in favour of #965

PR #846, opened 2024-01-10
fix: [L-09] restrict owner change in _beforeTokenTransfer

skimaharvey:

- [x] Restrict owner change in _beforeTokenTransfer

### DIFF
--- a/contracts/LSP8IdentifiableDigitalAsset/LSP8Errors.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/LSP8Errors.sol
@@ -103,3 +103,12 @@ error LSP8TokenIdsDataEmptyArray();
  * @notice Batch call failed.
  */
 error LSP8BatchCallFailed(uint256 callIndex);
+
+/**
+ * @dev Reverts when the token owner changed inside the {_beforeTokenTransfer} hook.
+ */
+error LSP8TokenOwnerChanged(
+    bytes32 tokenId,
+    address oldOwner,
+    address newOwner
+);
--- a/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
+++ b/contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
@@ -41,7 +41,8 @@ import {
     LSP8NotifyTokenReceiverIsEOA,
     LSP8TokenIdsDataLengthMismatch,
     LSP8TokenIdsDataEmptyArray,
-    LSP8BatchCallFailed
+    LSP8BatchCallFailed,
+    LSP8TokenOwnerChanged
 } from "./LSP8Errors.sol";
 
 // constants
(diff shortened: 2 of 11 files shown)


contracts/LSP8IdentifiableDigitalAsset/LSP8Errors.sol
```diff
@@ -103,3 +103,8 @@ error LSP8TokenIdsDataEmptyArray();
  * @notice Batch call failed.
  */
 error LSP8BatchCallFailed(uint256 callIndex);
+
+/**
+ * @dev Reverts when the token owner has changed after the _beforeTokenTransfer hook.
```
CJ42, line 108: Let's make that as an anchor link to the internal function for the docs:

```suggestion
 * @dev Reverts when the token owner changed inside the {_beforeTokenTransfer} hook.
```

contracts/Mocks/Tokens/LSP8TransferOwnerChange.sol
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    function _beforeTokenTransfer(
+        address,
+        address,
+        bytes32 tokenId,
+        bytes memory
+    ) internal override {
+        // if tokenID exist transfer token ownership to this contract
+        if (_tokenOwners[tokenId] != address(0)) {
```
CJ42, line 47: Maybe we can use the internal function here?

```suggestion
        if (_exists(tokenId)) {
```

image

contracts/LSP8IdentifiableDigitalAsset/LSP8Errors.sol
```diff
@@ -103,3 +103,8 @@ error LSP8TokenIdsDataEmptyArray();
...
+ */
+error LSP8TokenOwnerChanged(address oldOwner, address newOwner);
```
CJ42, line 110: I think we should also specify for which `tokenId` the owner has changed here. Wdyt?

```suggestion
error LSP8TokenIdOwnerChanged(bytes32 tokenId, address oldOwner, address newOwner);
```

contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
```diff
@@ -634,9 +635,11 @@ abstract contract LSP8IdentifiableDigitalAssetCore is
 
         _beforeTokenTransfer(from, to, tokenId, data);
 
-        // Re-fetch and update `tokenOwner` in case `tokenId`
-        // was transferred inside the `_beforeTokenTransfer` hook
-        tokenOwner = tokenOwnerOf(tokenId);
+        // Check that `tokenId`'s owner was not changed inside the `_beforeTokenTransfer` hook
+        address currentTokenOwner = tokenOwnerOf(tokenId);
+        if (tokenOwner != currentTokenOwner) {
+            revert LSP8TokenOwnerChanged(tokenOwner, currentTokenOwner);
```
CJ42, line 641: Suggest using named parameters for readability here:

```suggestion
            revert LSP8TokenOwnerChanged({
                oldOwner: tokenOwner, 
                newOwner: currentTokenOwner
            });
```

CJ42 (commented):
Added review comments

contracts/LSP8IdentifiableDigitalAsset/LSP8IdentifiableDigitalAssetCore.sol
skimaharvey: well, in most of our errors we don't use them. If I do for this one we will need to update all the errors out of consistency

CJ42: Ok let's keep it as it is for this one.

CJ42 (approved):

b00ste (approved):

PR #855, opened 2024-01-23
build!: move LSP4 in its own package

b00ste:

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`


packages/LSP4DigitalAssetMetadata/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+4. Setup the npm commands for linting, building, testing, etc... under the `"scripts"` in the `package.json`
+
+5. Test that all commands you setup run successfully
+
+By running the commands below, your LSP package should come up in the list of packages that Turbo is running this command for.
+
+```bash
+turbo build
```
richtera, line 44: You should not install turbo globally but rather as part of the package.json and then just create a script called turbo so you can call yarn turbo ... The reason is that if you rely on developers to install npm i -g turbo each developer will end up with a different version causing things to possibly work different or not.

richtera (approved):
Looks ok, but npx or npm i -g is not ideal. In many cases yarn turbo will work automatically if turbo is a dependency.

CJ42 (changes requested):
This PR is not complete.

- [x] LSP4 in the contracts folder needs to be removed
- [x] `"lsp4": "*" needs to be added at the root `package.json`
- [ ] all the Solidity import paths and other import paths needs to be updated

richtera (approved):

CJ42 (approved):

PR #857, opened 2024-01-23
build!: move LSP16 in its own package

YamenMerhi:
# What does this PR introduce?

Create package for LSP16 UniversalFactory standard.

PR #863, opened 2024-01-26
ci: cache Typescript `types/` of packages in CI

CJ42:
This is necessary for the typechain types of #857 to pass.


CJ42:
Please don't review this PR. Might probably be closed.

PR #884, opened 2024-02-09
feat: move constants in their respective packages

CJ42:

- Move all the constants respective to each LSP inside their own LSP package as a `constants.ts` file.
- Import these constants into the root `constants.ts` file of the `@lukso/lsp-smart-contracts`, referring to interface IDs, LSP1 Type IDs, and ERC725Y Data Keys for ease of use.
- add main file exported as index


richtera (approved):
LGTM

PR #925, opened 2024-03-11
feat: create `LSP4Utils.sol`

b00ste:

## 🚀 Feature

This PR creates a solidity lib for LSP4 that allows to easier manage an on-chain json for LSP4Metadata

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/packages/lsp4-contracts/.solhint.json
+++ b/packages/lsp4-contracts/.solhint.json
@@ -20,6 +20,7 @@
     "reason-string": ["warn", { "maxLength": 120 }],
     "avoid-low-level-calls": "off",
     "no-empty-blocks": ["error", { "ignoreConstructors": true }],
-    "custom-errors": "off"
+    "custom-errors": "off",
+    "quotes": "off"
   }
 }
--- a/packages/lsp4-contracts/contracts/LSP4Utils.sol
+++ b/packages/lsp4-contracts/contracts/LSP4Utils.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.17;
+pragma solidity ^0.8.24;
 
 import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 import {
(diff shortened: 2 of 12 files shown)


packages/lsp4-contracts/.solhint.json
```diff
@@ -20,6 +20,7 @@
     "reason-string": ["warn", { "maxLength": 120 }],
     "avoid-low-level-calls": "off",
     "no-empty-blocks": ["error", { "ignoreConstructors": true }],
-    "custom-errors": "off"
+    "custom-errors": "off",
+    "quotes": "off"
```
CJ42, line 24: why was this rule added?

packages/lsp4-contracts/contracts/LSP4Utils.sol
```diff
@@ -0,0 +1,288 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+// lib
```
CJ42, line 4: no need for a comment if it's just one import. but no biggie

packages/lsp4-contracts/contracts/LSP4Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+}
+
+struct Attribute {
+    string key;
+    string value;
+    string valueType;
+}
+
+struct Verification {
+    string method;
```
CJ42, line 19: Wouldn't it be better to use `bytes4` here, as we use verification methods identifiers in the end?

- `0x6f357c6a`: `keccak256('keccak256(utf8)')`
- etc...

packages/lsp4-contracts/contracts/LSP4Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+                ",",
+                '"url":',
+                string.concat('"', link.url, '"'),
+                "}"
+            );
+    }
+
+    function toJson(
+        Link[] memory links
+    ) internal pure returns (string memory object) {
```
CJ42, line 64: nice function 👌

packages/lsp4-contracts/contracts/LSP4Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+                object = string.concat(object, ",");
+            }
+        }
+
+        object = string.concat("[", object, "]");
+    }
+
+    function toJson(
+        Attribute memory attribute
+    ) internal pure returns (string memory) {
```
CJ42, line 78: I am not 100% sure, but I would believe the less elements to concat, the better and more efficient. So suggest following refactoring:

```solidity
return string.concat(
          "{'key':'", 
          attribute.key, 
          "','value':'", attribute.value, 
          "','type':'", attribute.valueType,
          "'}"
        );
```

packages/lsp4-contracts/contracts/LSP4Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+                object = string.concat(object, ",");
+            }
+        }
+
+        object = string.concat("[", object, "]");
+    }
+
+    function toVerifiableUri(
+        string memory data
+    ) internal pure returns (bytes memory verifiableUri) {
```
CJ42, line 262: Suggesting same renaming for function name.

```suggestion
    ) internal pure returns (bytes memory verifiableURI) {
```

packages/lsp4-contracts/contracts/LSP4Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+        return
+            bytes.concat(
+                bytes2(0),
+                bytes4(0x8019f9b1),
```
CJ42, line 266: suggest placing that in a `constant`.
Maybe this is a good candidate for a new `LSP2Constants.sol` file?

```solidity
bytes4 constant _KECCAK256_UTF8 = 0x6f357c6a;
bytes4 constant _KECCAK256_BYTES = 0x8019f9b1;
```

CJ42 (commented):
Very nice new library for LSP4Utils! 🚀 

I have added some review comments, and I also suggest with fuzz the functions with some Foundry tests. Good candidate here.

packages/lsp4-contracts/contracts/LSP4Utils.sol
b00ste: we could use bytes4 but then we have to have a if/else if/else in the function. and that will run on every image, not sure if its the best approach

packages/lsp4-contracts/.solhint.json
b00ste: because we have single quotes and linter exited with error

packages/lsp4-contracts/contracts/LSP4Utils.sol
b00ste: but it removes from readability, I propose testing the gas cost and if its significant, then refactor to a more compact concatenation

CJ42: OK agree. Let's test and see. You can post the difference in this comment, so we keep the context of the thread.

b00ste: image
image

The difference seems signifiant, I will optimise all the string concats

CJ42 (approved):

skimaharvey (approved):
lgtm

PR #926, opened 2024-03-13
build: replace `*` placeholder with actual dependency version

CJ42:


richtera (approved):
It's ok but npm pack should already be writing resolved versions into the output package.json

CJ42:
@richtera we might want to close this PR then?

PR #960, opened 2024-06-13
ci: use hardhat deploy plugin `tags` property in hardhat config instead of checks in workflow

CJ42:


CJ42:
Closing in favour of #963

PR #971, opened 2024-08-22
test: add extreme case tests

b00ste:

## 🧪 Tests

Add tests:
- Make sure follow/unfollow cannot be reverted by return bomb attack
- Make sure follow/unfollow cannot be reverted by an infinite loop
- Make sure follow/unfollow cannot be reverted when contract self destructs on interface check

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/packages/lsp26-contracts/tests/LSP26FollowerSystem.test.ts
+++ b/packages/lsp26-contracts/tests/LSP26FollowerSystem.test.ts
@@ -15,6 +15,12 @@ import {
   LSP0ERC725Account__factory,
   RevertOnFollow__factory,
   RevertOnFollow,
+  ReturnBomb__factory,
+  ReturnBomb,
+  SelfDestructOnInterfaceCheck__factory,
+  SelfDestructOnInterfaceCheck,
+  InfiniteLoopURD,
+  InfiniteLoopURD__factory,
 } from '../types';
 
 describe('testing `LSP26FollowerSystem`', () => {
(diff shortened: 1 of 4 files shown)


YamenMerhi (approved):

JeneaVranceanu (approved):
Whoami to approve? Looks good in the eyes of an amateur. Something new to learn 🙂

packages/lsp26-contracts/tests/LSP26FollowerSystem.test.ts
```diff
@@ -115,6 +121,96 @@ describe('testing `LSP26FollowerSystem`', () => {
...
+      expect(
+        await context.followerSystem.isFollowing(
+          context.owner.address,
+          await returnBomb.getAddress(),
+        ),
+      ).to.be.false;
+    });
+  });
+
+  describe('testing follow/unfollow a contract that has an infinite loop in urd', () => {
```
JeneaVranceanu, line 184: So.. a newbie question, how does it pass with this infinite loop running?
I see we get to 
```
ILSP1UniversalReceiver(addr).universalReceiver(
                    _TYPEID_LSP26_FOLLOW,
                    abi.encodePacked(msg.sender)
                )
```

In the LSP26 contract and how does it not fail meaning the loop doesn't prevent transaction from failing letting the user to follow/unfollow?

b00ste: The external call to the contract that has the `unviersalReceiver` function with an infinite loop reverts, but due to the fact that we call that function using `try/catch` and we don't do anything in the `catch`, it does not revert while following or unfollowing someone. Usually you would catch the revert in the catch block and do what you need to from there. In our case we want to make sure we notify yet never revert.

PR #972, opened 2024-08-28
feat: add ERCTokenCallbacks Extension

YamenMerhi:
Add an extension for all major token callbacks


skimaharvey (approved):
LGTM

CJ42 (changes requested):
LGTM. Ideally we should add some basic tests to ensure these work and don't revert on ERC token transfers with the callback hooks.

CJ42 (approved):
Let's merge and add the tests for this after. As we need to unlock to get the repo back in working state with only `main`

PR #973, opened 2024-08-30
fix: Match node versions across dev, github actions so it works with mise, asdf and so on

richtera:

## 🐛 Bug

Currently the project will not work on a machine with mise or asdf
due to the fact that the .tool-versions points to node 16.x which is no
longer compatible with most of the npm modules.

Along the way I also found that a lot of the github actions were still using
the old actions/setup-node@v2 and actions/setup-node@v3 as well
as specifying node 16.x

### PR Checklist

- [ ] Wrote Tests
- [ ] Ran `npm run build`
- [ ] Ran `npm run test`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -34,7 +34,7 @@ jobs:
           fetch-depth: 0
 
       - name: Use Node.js v20
-        uses: actions/setup-node@v3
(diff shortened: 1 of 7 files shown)


CJ42 (approved):
Nice that was indeed needed! 🙌🏻

PR #1032, opened 2025-10-07
refactor: use `struct` params in `constructor` of `CustomizableToken` to prevent stack-too-deep error 

CJ42:


frozeman:
Task linked: [DEV-15106 Remove compilation via IR for CustomizableToken](https://app.clickup.com/t/2645698/DEV-15106)

PR #1090, opened 2026-04-07
feat: create LSP8 version of new extension contracts, refactored with `AccessControlExtended` in inheritance

CJ42:

## 🚀 Feature

New version of LSP8 contracts, refactored to use `AccessControlExtended` in inheritance

### PR Checklist

- [x] Wrote Tests
- [x] Wrote & Generated Documentation (readme/natspec/dodoc)
- [x] Ran `npm run build`
- [x] Ran `npm run test`


LUKSOAgent (changes requested):
Thanks, I pulled this locally and reviewed the diff. I'm requesting changes for two parity issues and one breaking-change/documentation issue.

1. `packages/lsp8-contracts/contracts/extensions/AccessControlExtended/AccessControlExtendedAbstract.sol`
   The new LSP8 `AccessControlExtended` is already out of sync with the LSP7 version merged in #1089. `_transferOwnership()` here still only transfers `DEFAULT_ADMIN_ROLE`, while the LSP7 version now transfers **all** roles held by the old owner to the new owner. If both extensions ship under the same `AccessControlExtended` name/API family, downstream users will get different ownership-sync semantics depending on whether they use LSP7 or LSP8.

2. `packages/lsp8-contracts/contracts/extensions/AccessControlExtended/IAccessControlExtended.sol` + `AccessControlExtendedConstants.sol`
   This interface also still lacks `getRoleMembers(bytes32)`, and the ERC165 constant remains the pre-#1089 value (`0x4b701f90`). So the parity gap is not just behavioral, it also exists at the interface / ERC165 layer.

3. The PR deletes `LSP8Allowlist` and `presets/ILSP8Mintable`. If that is intentional, it needs to be called out explicitly as a breaking change in the PR description/changelog. Right now it reads like a refactor, but it removes public imports/contracts that downstream users may rely on.

Once the LSP8 extension is brought back in sync with the current LSP7 `AccessControlExtended` surface/semantics, or intentionally renamed/scoped, I'm happy to re-review.

LUKSOAgent (approved):
Re-review done.

I re-checked the points from my earlier changes-requested review against the current head:
- `AccessControlExtended` parity is now back in sync on the LSP8 side, including ownership transfer moving the full role set.
- `getRoleMembers(bytes32)` is present in the interface/implementation and the ERC165 constant has been updated to match the computed interface id.
- The new Foundry coverage around constructor/init ownership sync and full-role transfer is the right regression guard for this refactor.

I still think the removal of the legacy `LSP8Allowlist` / preset import paths should be called out clearly in changelog/release notes if this ships as part of a public package update, but I no longer see a code-level blocker in the PR itself.

PR #1094, opened 2026-04-13
refactor: add `isRevokable()` public function and param in `constructor` / `initalize(...)`

CJ42:


frozeman:
Task linked: [DEV-16165 Add feature to enable / disable revokable feature (or a public getter function)](https://app.clickup.com/t/2645698/DEV-16165)

